PR #191, opened 2015-04-03
refactor result table printing

whoburg:
This could still use work, but I took a shot at refactoring some of the table printing. In particular

* now sorting so that Vectors and non-vectors come in separate chunks
* can now optionally print in fixed-width columns.
* made many of the formatting strings explicit parameters
* removed code that prints constant-valued vectors as if they are scalars, as this seems possibly misleading

Here's an example of the current output for `print sol.table()`

```
Cost
 0.00474 

Variables
 CDA : 0.005                              [m**2]    projectile drag area
   R : 211                                [m]       range               
Vmax : 100                                [m/s]     max firing velocity 
   g : 9.8                                [m/s**2]  gravity             
   m : 1                                  [kg]      projectile mass     
 rho : 1.2                                [kg/m**3] air density         
   V : [ 1e+02    95       91      ... ]  [m/s]                         
  dt : [ 0.13     0.14     0.14    ... ]  [s]                           
  vx : [ 79       76       73      ... ]  [m/s]                         
  vy : [ 61       57       54      ... ]  [m/s]                         

Constant sensitivities
 CDA : 0.494   projectile drag area
Vmax : -1.01   max firing velocity 
   g : 0.506   gravity             
   m : -0.494  projectile mass     
 rho : 0.494   air density   
```

and here's `print sol.table(fixedcols=False)`
```

Cost
 0.00474 

Variables
 CDA : 0.005    [m**2] projectile drag area
   R : 211      [m] range
Vmax : 100      [m/s] max firing velocity
   g : 9.8      [m/s**2] gravity
   m : 1        [kg] projectile mass
 rho : 1.2      [kg/m**3] air density
   V : [ 1e+02    95       91      ... ]  [m/s] 
  dt : [ 0.13     0.14     0.14    ... ]  [s] 
  vx : [ 79       76       73      ... ]  [m/s] 
  vy : [ 61       57       54      ... ]  [m/s] 

Constant sensitivities
 CDA : 0.494   projectile drag area
Vmax : -1.01   max firing velocity
   g : 0.506   gravity
   m : -0.494  projectile mass
 rho : 0.494   air density
```

### DIFF
--- a/gpkit/geometric_program.py
+++ b/gpkit/geometric_program.py
@@ -89,24 +89,31 @@ class GPSolutionArray(DictOfLists):
             assert not subbed.exp
             return subbed.c
 
-    def table(self, tables=["cost", "variables", "sensitivities"]):
+    def table(self,
+              tables=["cost", "variables", "sensitivities"],
+              fixedcols=True):
         if isinstance(tables, Strings):
             tables = [tables]
         strs = []
         n = len(self)
         if "cost" in tables:
-            strs += ["              |"]
+            strs += [""]
             if len(self) > 1:
-                costs = ["%-7.2g" % self["cost"][i] for i in range(min(n, 3))]
-                strs += ["         Cost : [%s ... ]" % "  ".join(costs)]
+                costs = ["%-7.2g" % c for c in self["cost"][:3]]
+                strs += ["Cost\n [%s ... ]" % "  ".join(costs)]
             else:
-                strs += ["         Cost :  %-8.3g" % self["cost"]]
-            strs += ["              | "]
+                strs += ["Cost\n %-8.3g" % self["cost"]]
+            strs += [""]
         if "variables" in tables:
-            strs += [results_table(self["variables"], "Variables")]
+            strs += [results_table(self["variables"],
+                                   "Variables",
+                                   fixedcols=fixedcols)]
         if "sensitivities" in tables:
             strs += [results_table(self["sensitivities"]["variables"],
-                                   "Constant sensitivities", senss=True)]
+                                   "Constant sensitivities",
+                                   fixedcols=fixedcols,
+                                   minval=1e-2,
+                                   printunits=False)]
         return "\n".join(strs)
 
 
--- a/gpkit/small_scripts.py
+++ b/gpkit/small_scripts.py
@@ -178,7 +178,7 @@ def results_table(data, title, minval=0, printunits=True, fixedcols=True,
         skip values with all(abs(value)) < minval
     printunits: bool
     fixedcols: bool
-        print rhs (val, units, label) in fixed-width cols?
+        if True, print rhs (val, units, label) in fixed-width cols
     varfmt: string
         format for variable names
     valfmt: string
@@ -206,7 +206,7 @@ def results_table(data, title, minval=0, printunits=True, fixedcols=True,
     if not fixedcols:
         maxlens = [maxlens[0], 0, 0, 0]
     dirs = ['>', '<', '<', '<']
-    assert len(dirs) == len(maxlens)
+    assert len(dirs) == len(maxlens)  # always check lengths before using zip
     fmts = ['{0:%s%s}' % (direc, L) for direc, L in zip(dirs, maxlens)]
     lines = [[fmt.format(s) for fmt, s in zip(fmts, line)]
              for line in lines]
--- a/gpkit/small_scripts.py
+++ b/gpkit/small_scripts.py
@@ -64,16 +64,16 @@ def mag(c):
         return c
 
 
-def unitstr(units, into="%s", options="~"):
+def unitstr(units, into="%s", options="~", dimless='-'):
     if hasattr(units, "descr"):
         if isinstance(units.descr, dict):
-            units = units.descr.get("units", "-")
+            units = units.descr.get("units", dimless)
     if units and not isinstance(units, Strings):
         try:
             rawstr = ("{:%s}" % options).format(units)
         except:
             rawstr = "1.0 " + str(units.units)
-        units = "".join(rawstr.replace("dimensionless", "-").split()[1:])
+        units = "".join(rawstr.replace("dimensionless", dimless).split()[1:])
     if units:
         return into % units
     else:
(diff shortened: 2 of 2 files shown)


gpkit/geometric_program.py
```diff
@@ -89,24 +89,31 @@ def senssubinto(self, p):
...
             tables = [tables]
         strs = []
         n = len(self)
         if "cost" in tables:
-            strs += ["              |"]
+            strs += [""]
             if len(self) > 1:
-                costs = ["%-7.2g" % self["cost"][i] for i in range(min(n, 3))]
-                strs += ["         Cost : [%s ... ]" % "  ".join(costs)]
+                costs = ["%-7.2g" % c for c in self["cost"][:3]]
```
bqpd: much better way to do this indexing!

gpkit/small_scripts.py
```diff
@@ -163,41 +163,55 @@ def sort_and_simplify(exps, cs):
...
+    printunits: bool
+    fixedcols: bool
+        print rhs (val, units, label) in fixed-width cols?
+    varfmt: string
+        format for variable names
+    valfmt: string
+        format for scalar values
+    vecfmt: string
+        format for vector values
+    """
```
bqpd: whoo, docstring!

whoburg:
Should I merge it?

bqpd:
Ah, not yet; just going to read and comment some more, then I'll pull it!

bqpd:
(ask for a code review, get a code review is my thinking here...)

whoburg:
yes yes, good!

gpkit/geometric_program.py
```diff
@@ -89,24 +89,31 @@ def senssubinto(self, p):
...
+    def table(self,
+              tables=["cost", "variables", "sensitivities"],
+              fixedcols=True):
         if isinstance(tables, Strings):
             tables = [tables]
         strs = []
         n = len(self)
         if "cost" in tables:
-            strs += ["              |"]
+            strs += [""]
```
bqpd: I like the new formatting.

gpkit/small_scripts.py
```diff
@@ -64,16 +64,16 @@ def mag(c):
         return c
 
 
-def unitstr(units, into="%s", options="~"):
+def unitstr(units, into="%s", options="~", dimless='-'):
```
bqpd: Good to have this be settable; is a different string for `dimless` used anywhere?

gpkit/small_scripts.py
```diff
@@ -163,41 +163,55 @@ def sort_and_simplify(exps, cs):
...
+    Parameters
+    ----------
+    data: dict whose keys are VarKey's
+        data to represent in table
+    title: string
+    minval: float
+        skip values with all(abs(value)) < minval
+    printunits: bool
+    fixedcols: bool
+        print rhs (val, units, label) in fixed-width cols?
```
bqpd: Why the question mark?

gpkit/small_scripts.py
```diff
@@ -163,41 +163,55 @@ def sort_and_simplify(exps, cs):
...
+    varfmt: string
+        format for variable names
+    valfmt: string
+        format for scalar values
+    vecfmt: string
+        format for vector values
+    """
+    lines = []
+    decorated = [(bool(v.shape) if isinstance(v, Iterable) else False,
```
bqpd: Oh, zero-dimensional arrays.

gpkit/small_scripts.py
```diff
@@ -163,41 +163,55 @@ def sort_and_simplify(exps, cs):
...
+                 ('%s : ' % k),
+                 i, k, v) for i, (k, v) in enumerate(data.items())
+                 if np.max(abs(v)) >= minval]
+    decorated.sort()
```
bqpd: Ah, sorting here is a good plan.

gpkit/small_scripts.py
```diff
@@ -163,41 +163,55 @@ def sort_and_simplify(exps, cs):
...
+        format for variable names
+    valfmt: string
+        format for scalar values
+    vecfmt: string
+        format for vector values
+    """
+    lines = []
+    decorated = [(bool(v.shape) if isinstance(v, Iterable) else False,
+                 ('%s : ' % k),
+                 i, k, v) for i, (k, v) in enumerate(data.items())
```
bqpd: Why place `i` in the structure if it's not used in the for loop below?

gpkit/small_scripts.py
```diff
@@ -163,41 +163,55 @@ def sort_and_simplify(exps, cs):
...
-                         + "[ %s%s ]" % ("  ".join(vals), ellipsis)
-                         + ("  [%s] " % units if units else "  ")
-                         + "%s" % label]
-    strs += ["              |"]
-    return "\n".join(strs)
+            valstr = valfmt % val
+        lines.append([varstr, valstr, units, label])
+    maxlens = np.max([map(len, line) for line in lines], axis=0)
+    if not fixedcols:
+        maxlens = [maxlens[0], 0, 0, 0]
```
bqpd: Nice!

gpkit/small_scripts.py
```diff
@@ -163,41 +163,55 @@ def sort_and_simplify(exps, cs):
...
+    dirs = ['>', '<', '<', '<']
+    assert len(dirs) == len(maxlens)
```
bqpd: This could maybe raise a more useful error?

bqpd:
Made some comments; looks great overall! I think my only question was on the `assert`.

gpkit/small_scripts.py
whoburg: Yes -- `results_table` sets `dimless` to `""` -- nothing is printed for unitless quantities.

`unitstr` is being a little sloppy and probably needs documentation -- it's potentially confusing because it returns `into % units if units else ""`. This means that when `dimless` is `"-"` and `into` is `"[%s]"`, `unitstr` currently returns `"[-]"`, whereas when `dimless` is `""` and `into` is `"[%s]"`, `unitstr` currently returns `""`.

Documentation, suggestions, or improvements welcome...

gpkit/small_scripts.py
whoburg: Oh, because that's the question that the boolean parameter `fixedcols` answers. Ideas for better way to document welcome.

gpkit/small_scripts.py
whoburg: This is the old "decorate-sort-undecorate" strategy, and the `i` makes the sort "stable", meaning that if two items happen to have the same vector-ness and str(key), their order will be preserved in the sorted list.

I'm not staying that's likely to happen here; rather I'm following a sorting recipe that guarantees stability.

gpkit/small_scripts.py
whoburg: Yes, probably.

This is a habit I picked up a long time ago -- always asserting lengths are equal before using Python's `zip`. That's because `zip` is happy to have different-length arguments:

```python
>>> a = [1,2,3]
>>> b = [1,2,3,4,5]
>>> zip(a,b)
[(1, 1), (2, 2), (3, 3)]
```
... and I find it scary that if the lists I'm assuming are the same length were different lengths, it could go undetected.

Thoughts on the right error to `raise`?

Or I could use `map` instead of `zip`, with some increase in verbosity...

whoburg: maybe `map` is better...

gpkit/small_scripts.py
bqpd: Maybe `if True, print rhs (val, units, label) in fixed-width cols`

gpkit/small_scripts.py
bqpd: Ah, neat.

gpkit/small_scripts.py
bqpd: The assert seems fine; maybe add an inline comment to note that rationale?

gpkit/small_scripts.py
whoburg: nice

gpkit/small_scripts.py
whoburg: done

bqpd:
:+1:

whoburg:
should I merge?

bqpd:
yup!

PR #203, opened 2015-04-17
PosyArray improvements

whoburg:
* implement new convenience methods useful for finite-differencing `PosyArray`s

Example usage:
```python
from gpkit import VectorVariable
x = VectorVariable(10, 'x', 'm')
dx = VectorVariable(10, 'dx', 'm')
constraint = (x >= x.left + dx/2. + dx.left/2.)
```

### DIFF
--- a/gpkit/posyarray.py
+++ b/gpkit/posyarray.py
@@ -151,3 +151,15 @@ class PosyArray(np.ndarray):
     def sub(self, subs, val=None, allow_negative=False):
         "Substitutes into the array"
         return PosyArray([p.sub(subs, val, allow_negative) for p in self])
+
+    def right(self):
+        "Self, sampled one index up, with zeropad"
+        if self.ndim != 1:
+            raise NotImplementedError("not implemented for ndim=%s" % self.ndim)
+        return PosyArray(np.hstack((self[1:], 0)))
+
+    def left(self):
+        "Self, sampled one index down, with zeropad"
+        if self.ndim != 1:
+            raise NotImplementedError("not implemented for ndim=%s" % self.ndim)
+        return PosyArray(np.hstack((0, self[:-1])))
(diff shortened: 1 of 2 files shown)


bqpd:
The above also imposes the constraint `x[0] >= dx[0]/2`, is that intended?

bqpd:
(also, do you think this should be an @property so that you can do `x.left`?)

whoburg:
Yes, the idea on imposing `x[0] >= dx[0]/2` is that the `x`'s are the center of some computational bins, and the `dx`s are the bin widths. Therefore, the first x should be at least half a bin width from zero.

I went back and forth on @property, and I don't have a strong opinion. I'll change it to a property, sounds user-friendly for modelers.

whoburg:
The names might need work. On the plus side, they're easy and short when this is the behavior you want :/

bqpd:
Sounds good to me! The names `left` and `right` aren't super clear to me (though, no strong opinion).

Alternatives:
  - `leftshift`/`rightshift`
  - `lshift`
  - `bumpleft`

bqpd:
`left` and `right` seem fine, but for whatever reason my first guess was that `left` would leave a 0 on the left

whoburg:
`left` does indeed leave a 0 in the leftmost position.

My concern about `bumpleft` (and the related alternatives) is that they sound like we're bumping/shifting the entries in the PosyArray. In contrast, here we're bumping the indices. "shift your view over one and sample there". I claim this is the most intuitive behavior if you're thinking about finite differences, i.e.

`dx = x.right - x.left`

But there's definitely potential confusion here

bqpd:
So it does! I know I got it wrong at first, so guess I've already internalized it. :)

I'd say this looks good to merge! Might be good to get @cjk12's opinion on syntax, since he's written these constraints before?

PR #206, opened 2015-04-17
python3 compatibility

bqpd:
still fails on dict re-ordering in tests and Posy equality


bqpd:
Doesn't break python 2 compatibility, but would like to have working python 3 compatibility before merging...

bqpd:
Superseded, closing.

PR #213, opened 2015-04-18
pylint challenge! 

galbramc:
Modify gpkit/tests/pylint.cfg to only create warnings and errors for things you care about. You can test with the command:

pylint --rcfile=gpkit/test/pylint.cfg $(find .  -name "*.py" -print) gpkit/

Alternatively you can rename gpkit/test/pylint.cfg to gpkit/../pylint or gpkit/../.pylintrc and
remove the need for the --rcfile flag.

### DIFF
--- a/gpkit/tests/pylint.cfg
+++ b/gpkit/tests/pylint.cfg
@@ -1,0 +1,280 @@
+[MASTER]
+
+# Specify a configuration file.
+#rcfile=
+
+# Python code to execute, usually for sys.path manipulation such as
+# pygtk.require().
+#init-hook=
+
+# Profiled execution.
+profile=no
+
+# Add files or directories to the blacklist. They should be base names, not
+# paths.
+ignore=CVS
+
+# Pickle collected data for later comparisons.
+persistent=yes
+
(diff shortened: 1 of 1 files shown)


galbramc:
This was also a test to make sure jenkins did not build from a user that is not white listed. It worked as expected.

galbramc:
Can one of the admins verify this patch?

galbramc:
That last question was actually posted by jenkins (its using my user name a the moment and I need to fix that)

bqpd:
Heh! This is great. Should we set it up to ignore the ctypesgen-generated files (which are a pylint disaster) and so on before merging?

galbramc:
You should setup the config file so it tests what you think is appropriate to be tested. This most likely only means files that you are directly authoring. The file I committed is just the default pylint configuration file, I didn't make any changes to it.

bqpd:
Sounds good.

whoburg:
Guys, let's definitely talk together about what warnings/errors to keep vs. disable, before we make any changes.

whoburg:
Continuing this in #216, so that we have access to make changes before merging.

Issue #212, opened 2015-04-17
improve partitioning of variables (into free variables, constants, etc)


whoburg:
Results table used to separate free variables from constants, but I think I accidentally killed that behavior with https://github.com/convexopt/gpkit/pull/191. Should separate them again -- that was nice.

whoburg:
Good example of the difference is here: 363b8ec0b04ba74673e817fbbe753a91c0306938

bqpd:
Oops, no, I killed it accidentally during the Signomial programming
process; GPSolutionArray no longer distinguishes!

We were talking about giving GPSolutionArray some additional information
about the program, and variable-type could be a good candidate.

whoburg:
Should we think about putting this info in the `VarKey`? Or is that dumb?

bqpd:
Hmmmm. I'd store it in the solution array and keep the VarKeys as the user created them.

whoburg:
Just to make sure, what I meant is, if I call 

`x = Variable('x', 3, 'm')`

and then look at the resulting `VarKey`, there's no way I can see to tell that this `VarKey` corresponds to a constant. Should there be? I feel like maybe we've already visited this question though. Sorry if I'm derailing.

bqpd:
Sure you can tell; `x.value != x`, therefore it's a constant.

alternate test: `"value" in x.varkeys.values()[0].descr`

whoburg:
This broke `docs/source/examples/simple_box.py`:

```python
examples whoburg$ python simple_box.py
Traceback (most recent call last):
  File "simple_box.py", line 35, in <module>
    print sol.table()
  File ".../anaconda/envs/gpkit/lib/python2.7/site-packages/gpkit/geometric_program.py", line 125, in table
    strs += [results_table({k: v[0] for (k, v) in self["variables"].items()
  File ".../anaconda/envs/gpkit/lib/python2.7/site-packages/gpkit/geometric_program.py", line 126, in <dictcomp>
    if k in self.gp.substitutions},
IndexError: too many indices for array
```

bqpd:
Yeah, it wasn't the greatest way to do this. It's fixed in my gp-refactor branch, so I'll focus on getting that working...

whoburg:
It seems to me that perhaps the partitioning of variables (into free variables, constants, etc) should be either a `@property` or an attribute at the `GPSolutionArray` level -- right now it happens inside the `table()` method, but I can imagine other places we might want this partitioning...

[...]

PR #215, opened 2015-04-21
Issue212

whoburg:
Proposed fix for issue #212


bqpd:
This doesn't catch variables that are substituted but don't have values (such as sweeps). Should it?

whoburg:
Yes, it should. My fault. Is there an easy way to do that, or is my plan of getting this info from a `VarKey` a dead end?

bqpd:
Unfortunately it may be...whether something has recently been substituted or not is a more dynamic property than I feel comfortable putting into a VarKey.

whoburg:
Ok, bailing on this strategy, and issue #212 remains open.

PR #227, opened 2015-04-24
rename ctypesgen.py to modified_ctypesgen.py

whoburg:
Rename ctypesgen.py to modified_ctypesgen.py, so it’s more transparent what happened (we added command line options to the script).


whoburg:
@bqpd assigning you since this touches the build script. It seems to run fine for me still.

bqpd:
I...think Jenkins built it ok? Hard to tell.

whoburg:
Yes, as far as I can tell it seems to be building fine and passing all our unit tests; Jenkins seems to be marking as failure due to pylint... I think?

I also find it hard to read.

galbramc:
Yes this passes the tests on Linux and Mac, still have the failures on Windozs due to deleting the non-empty directory,

whoburg:
Ok, so given that the build failures on windows are not due to the changes under this pull request, I think we can merge. @bqpd, if you agree, please do so.

bqpd:
SGTM

whoburg:
Just spent 5 minutes trying to determine whether this means "sounds good to me" or "silently giggling to myself".

bqpd:
Hehehe. The former.

Issue #224, opened 2015-04-22
standardizing method names


whoburg:
proposal:

* rename function `enableUnits()` --> `enable_units()`
* rename function `disableUnits()` --> `disable_units()`
* create functions `enable_signomials()` and `disable_signomials()`, to replace current practice of modifying the module-level boolean variable `gpkit.enable_signomials`.

bqpd:
Sounds good to me.

whoburg:
Implemented in #228. Forgot to build the docs, resulting in #231, but that's now fixed.

PR #228, opened 2015-04-25
Issue224 -- standardizing method names

whoburg:
As discussed in #224,

* rename function `enableUnits()` --> `enable_units()`
* rename function `disableUnits()` --> `disable_units()`
* create functions `enable_signomials()` and `disable_signomials()`, to replace current practice of modifying the module-level boolean variable `gpkit.enable_signomials`.

### DIFF
--- a/gpkit/__init__.py
+++ b/gpkit/__init__.py
@@ -17,7 +17,7 @@
 """
 
 
-def disableUnits():
+def disable_units():
     """Disables units support in a particular instance of GPkit.
 
     Posynomials created after calling this are incompatible with those created
@@ -41,7 +41,7 @@ def disableUnits():
     DimensionalityError = ValueError
 
 
-def enableUnits():
+def enable_units():
     """Enables units support in a particular instance of GPkit.
 
     Posynomials created after calling this are incompatible with those created
@@ -56,9 +56,9 @@ def enableUnits():
     except ImportError:
         print("Optional Python units library (Pint) not installed;"
               "unit support disabled.")
-        disableUnits()
+        disable_units()
 
-enableUnits()
+enable_units()
 
 _SIGNOMIALS_ENABLED = False
 
--- a/gpkit/__init__.py
+++ b/gpkit/__init__.py
@@ -60,7 +60,20 @@ def enableUnits():
 
 enableUnits()
 
-enable_signomials = False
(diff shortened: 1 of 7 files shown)


gpkit/__init__.py
```diff
@@ -56,11 +56,24 @@ def enableUnits():
...
+def enable_signomials():
+    """Enables signomial support in a particular instance of GPkit."""
+    global _SIGNOMIALS_ENABLED
+    _SIGNOMIALS_ENABLED = True
+
+
+def disable_signomials():
+    """Disables signomial support in a particular instance of GPkit."""
+    global _SIGNOMIALS_ENABLED
+    _SIGNOMIALS_ENABLED = False
```
whoburg: @bqpd, are the above implementations fine, or do you suggest something else?

bqpd: Such _CAPS! Looks fine to me.

gpkit/tests/t_nomials.py
```diff
@@ -1,7 +1,7 @@
 import math
 import unittest
 from gpkit import Monomial, Posynomial, Signomial
-import gpkit
+from gpkit import enable_signomials, disable_signomials
```
bqpd: Hmm, I think that doing a top-level gpkit import might be necessary here.

whoburg: Is it straightforward to explain why? (Unit tests all pass with the changes above.)

gpkit/__init__.py
whoburg: From pep8:

Constants
---------
Constants are usually defined on a module level and written in all capital letters with underscores separating words. Examples include `MAX_OVERFLOW` and `TOTAL` .

Global Variable Names
Modules that are designed for use via from M import * should use the __all__ mechanism to prevent exporting globals, or use the older convention of prefixing such globals with an underscore (which you might want to do to indicate these globals are "module non-public").

bqpd: Oh, nice that there's a pep standard for this.

gpkit/tests/t_nomials.py
bqpd: Oops, in that case no worries! I meant to pull and test this yesterday.

bqpd:
LGTM

PR #239, opened 2015-04-29
implement arrayVars; closes #237

### LINKED ISSUE #237: N Dimensional Variables
So, I have been working on some Trefftz Plane code in matlab, and I'd really like to have N Dimensional variable capability in GPkit.  This would be useful for something like tensors.  I'm actually storing a really complicated set of vectors that results in a 3D matrix that is (N-1)x(N)x(2).  I'm thinking the declaration should be something like:

``` python
A = ArrayVariable([2,2,2], "A", "-", "Array Variable")
```

would create a 2x2x2 variable.

bqpd:


bqpd:
Made this a PR just to check in with Jenkins; seems to be doing alright, so I'll merge it.

Issue #245, opened 2015-04-29
PosyArray.sum and posyArray.prod don't act like numpy sum/prod


whoburg:
continuation from #241 

* [x] `PosyArray` inherits from `np.ndarray`, so one would expect the following to work (currently commented-out in tests/t_vars/TestArrayVariable/test_sum):

```python
    def test_sum(self):
        x = ArrayVariable((2, 3), 'x')
        rowsums = x.sum(axis=1)
        self.assertTrue(isinstance(rowsums, PosyArray))
        self.assertEqual(len(rowsums), 2)
        colsums = x.sum(axis=0)
        self.assertTrue(isinstance(colsums, PosyArray))
        self.assertEqual(len(colsums), 3)
```

* [ ] Most of the tests under TestArrayVariable should probably be moved to t_posyarray/TestPosyArray.

bqpd:
Ah, good point. I think I implemented a custom sum / prod so they wouldn't return 0-d arrays? But feel free to delete the relevant lines in posarray.py and see if that breaks anything...

bqpd:
Is moving the tests the only thing remaining?

whoburg:
Affirm.

bqpd:
10-4

PR #250, opened 2015-05-01
Ndarray subclassing

whoburg:
Proposed fix for #245

### DIFF
--- a/gpkit/posyarray.py
+++ b/gpkit/posyarray.py
@@ -58,6 +58,16 @@ class PosyArray(np.ndarray):
             return
         self.info = getattr(obj, 'info', None)
 
+    def __array_wrap__(self, out_arr, context=None):
+        """Called by numpy ufuncs.
+        Special case to avoid creation of 0-dimensional arrays
+        See http://docs.scipy.org/doc/numpy/user/basics.subclassing.html"""
+        if out_arr.ndim:
+            return np.ndarray.__array_wrap__(self, out_arr, context)
+        else:
+            return out_arr.item()
+
+
     def _latex(self, unused=None, matwrap=True):
         "Returns 1D latex list of contents."
         if len(self.shape) == 1:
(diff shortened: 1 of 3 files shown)


gpkit/posyarray.py
```diff
@@ -30,33 +28,33 @@ class PosyArray(np.ndarray):
...
-        self.info = getattr(obj, 'info', None)
+        pass
+
+    def __array_wrap__(self, out_arr, context=None):
+        """Called by numpy ufuncs.
+        Special case to avoid creation of 0-dimensional arrays
+        See http://docs.scipy.org/doc/numpy/user/basics.subclassing.html"""
+        if out_arr.ndim:
+            return np.ndarray.__array_wrap__(self, out_arr, context)
+        return float(out_arr)
```
bqpd: `float`? But what if `out_arr` is a Variable / Posynomial?

whoburg: Doh - obviously dumb. What's most confusing is *why this worked*.

It seems that numpy catches all Exceptions coming from `__array_wrap__`, which is really sad. In this case, `float(out_arr)` raises an `AttributeError` when `out_arr` is a Signomial. What happens next has me baffled -- numpy catches the error, and then for some odd reason successfully returns a Signomial instead of a 0-dimensional array containing a Signomial.

This happens to be the behavior we want, but is obviously a terrible implementation -- pushed a change, will comment on that diff.

gpkit/posyarray.py
```diff
@@ -30,33 +28,40 @@ class PosyArray(np.ndarray):
...
+        See http://docs.scipy.org/doc/numpy/user/basics.subclassing.html"""
+        if out_arr.ndim:
+            return np.ndarray.__array_wrap__(self, out_arr, context)
+        try:
+            val = out_arr.item()
+            return np.float(val) if isinstance(val, np.generic) else val
+        except:
+            print("Something went wrong. I'd like to raise a RuntimeWarning,"
+                  " but you wouldn't see it because numpy seems to catch all"
+                  " Exceptions coming from __array_wrap__.")
```
bqpd: lol.

whoburg:
Brought up some of the unexpected behaviors we encountered here over at numpy:  https://github.com/numpy/numpy/issues/5819

Issue #268, opened 2015-05-21
Document Signomial.sub


whoburg:
Noticed this wasn't documented when trying to look up whether `Signomial.sub` mutates the signomial in place, or returns a new `Signomial`.

Issue #267, opened 2015-05-21
Substituting for a single element of a PosyArray


whoburg:
What is the best-practice syntax/method for substitutions on a single element of an existing `PosyArray`? Looking for a simple example to add to the docs.

bqpd:
from gpkit import VectorVariable
v = VectorVariable(3, "x")
v[0] = v[0].sub(v[0], 3)

whoburg:
Any thoughts on overriding `Variable.sub` so that

```python
x = Variable('x')
subbed = x.sub(3)
```

works?

whoburg:
bump

bqpd:
Subbed would be x with "value=3" in the descr, or would it just be 3?

whoburg:
Well, the proposal would be for `Variable('x').sub(3)` to be equivalent to `Variable('x', value=3)`. Seems really clean from a user interface perspective, but maybe it's not worth it if it requires a major refactor?

bqpd:
Naw, that'd be easy!

bqpd:
Turns out this isn't easy, because we currently assume that anything that inherits from Signomial uses its method of substitution. We could add a 'value' method instead?

whoburg:
I mistakenly glossed over the distinction between `Variable('x', value=3)` and `Monomial(3)`. Note that `Variable('x').sub({'x': 3})` currently returns `Monomial(3)`, which I'm fine with. My intent was simply to enable `Variable('x').sub(3)` as a shorthand, which works only for Variable objects, since they have a single `VarKey`.

I've implemented and tested this on the sub_improvements branch in 5e59f2adc73b8443e3e0582e8bc8d159810a5567.

bqpd:
Oh! Looks good, then.

I was thinking of `x.sub(3)` as a shorthand for `x.varkeys.values()[0].descr["value"] = 3`, which could serve as an alternative to passing around dictionaries of substitutions...but on second thought I think dictionaries are better.

whoburg:
just looked up this ticket for a use case where I'd really like to fix an element of a VectorVariable to 0. I think it'd be convenient to implement @bqpd's last suggestion, so that one can do

`x.setto(4.3)`  # set's the value of x to 4.3; mutates x.

Is that straightforward or does the race condition problem resurface here?

bqpd:
For a vector variable, is it ever going to have a nonzero value? If not, I'd just replace that element of the array with a 0...

PR #270, opened 2015-05-26
Sub improvements

whoburg:
Started some attempts to clean up and document substitution. Others welcome to contribute as well.

Related:
 -  #267 
 -  #268 
 -  #279

### DIFF
--- a/docs/source/advancedcommands.rst
+++ b/docs/source/advancedcommands.rst
@@ -57,12 +57,12 @@ You can also substitute in sweep variables (see Sweeps_), strings, and monomials
     yvk = y.varkeys.values()[0]
     for x_ in ["x", xvk, x]:
         for y_ in ["y", yvk, y]:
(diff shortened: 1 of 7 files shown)


whoburg:
Would like to merge current changes into master, since they resolve #279 . @bqpd

bqpd:
I'm not seeing where this fixes #279. It seems to assume that every varkey has a `.units`, but that's not true, is it?

whoburg:
In fact, every `VarKey` does have a units! This is due to the line `self.units = self.descr.get("units", None)` in the VarKey `__init__`.

Note that I'm also proposing a simplified `VarKey` constructor, along with making units an `@property`, in #280.

By the way, the merge conflict on this is just in a docstring -- I will take the master version when merging (once @bqpd agrees this should be merged).

whoburg:
Can we merge this? I want to make sure it doesn't get lost or become unmergable due to the GP refactor...

bqpd:
Yes! Merge it on in.

PR #276, opened 2015-06-15
simple_wing example touch-up

whoburg:
* Corrected `Variable` keyword argument from `descr` to `label`
* Added units
* pep8
* added simple_wing to Makefile and remade output
* Added vars for wing weight model params


bqpd:
This looks nice. We should standardize on the `label=` tags; some variable declarations use them, but some don't. I'm not sure which is better.

whoburg:
Related: I was thinking of opening a ticket to eliminate the `**kwargs` from `Variable`'s `__init__`. Proposal would be to explicitly enumerate kwargs we allow, and disallow all others. Reasoning is that the `descr=` mistake this PR replaced went through without error; I think that's potentially confusing for users.

Anyone have thoughts? if so let's open that discussion in a new ticket.

bqpd:
I like the notion that users (and libraries) can use custom keys for their own purposes; we've used this to test things like an sp_init flag.

whoburg:
Ok, fair enough... in that case perhaps my point can be addressed by tidying up / improving the docstring.

Issue #275, opened 2015-06-05
latex printing has bugs with units


whoburg:
Minimum working example:
```python
from gpkit import Variable
rho = Variable('\\rho', units='kg/m^3')
S = Variable('S', units='m^2')
x = rho*S
print x
```

output:
```python
---------------------------------------------------------------------------
DimensionalityError                       Traceback (most recent call last)
<ipython-input-9-2d264e11d975> in <module>()
----> 1 print x

.../gpkit/nomials.pyc in __str__(self, mult_symbol)
    241         for c, exp in zip(self.cs, self.exps):
    242             varstrs = ['%s**%.2g' % (var, x) if x != 1 else "%s" % var
--> 243                        for (var, x) in sorted(exp.items(), key=vkSortBy) if x != 0]
    244             c = mag(c)
    245             cstr = "%.2g" % c

.../envs/gpkit/lib/python2.7/site-packages/pint/quantity.pyc in <lambda>(self, other)
    815                   other.to_base_units().magnitude)
    816
--> 817     __lt__ = lambda self, other: self.compare(other, op=operator.lt)
    818     __le__ = lambda self, other: self.compare(other, op=operator.le)
    819     __ge__ = lambda self, other: self.compare(other, op=operator.ge)

.../envs/gpkit/lib/python2.7/site-packages/pint/quantity.pyc in compare(self, other, op)
    811         if self.dimensionality != other.dimensionality:
    812             raise DimensionalityError(self.units, other.units,
--> 813                                       self.dimensionality, other.dimensionality)
    814         return op(self.to_base_units().magnitude,
    815                   other.to_base_units().magnitude)

DimensionalityError: Cannot convert from 'kilogram / meter ** 3' ([mass] / [length] ** 3) to 'meter ** 2' ([length] ** 2)
```

PR #278, opened 2015-06-19
Fix printing bugs

whoburg:
Proposed fix for #275 

* pep8
* removes `vkSortBy`
* un-protects `_latex()` methods

### DIFF
--- a/gpkit/interactive/printing.py
+++ b/gpkit/interactive/printing.py
@@ -102,7 +102,7 @@ o can be printed with LaTeX.
             return all(_can_print_latex(i) and _can_print_latex(o[i]) for i in o)
         elif isinstance(o, bool):
             return False
-        elif hasattr(o, '_latex'):
+        elif hasattr(o, 'latex'):
             return True
         elif isinstance(o, (float, integer_types)) and print_builtin:
             return True
--- a/gpkit/nomials.py
+++ b/gpkit/nomials.py
@@ -253,16 +253,16 @@ class Signomial(object):
     def __repr__(self):
         return "gpkit.%s(%s)" % (self.__class__.__name__, str(self))
 
-    def _latex(self, unused=None):
+    def latex(self, unused=None):
         "For pretty printing with Sympy"
         mstrs = []
         for c, exp in zip(self.cs, self.exps):
             pos_vars, neg_vars = [], []
             for var, x in exp.items():
                 if x > 0:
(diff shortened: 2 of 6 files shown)


gpkit/interactive/printing.py
```diff
@@ -102,7 +102,7 @@ def _can_print_latex(o):
             return all(_can_print_latex(i) and _can_print_latex(o[i]) for i in o)
         elif isinstance(o, bool):
             return False
-        elif hasattr(o, '_latex'):
+        elif hasattr(o, 'latex'):
```
bqpd: This code was copied from SymPy, by the way. I think you're totally right though, `_latex` is gross!

gpkit/nomials.py
```diff
@@ -240,7 +235,8 @@ def __str__(self, mult_symbol='*'):
         mstrs = []
         for c, exp in zip(self.cs, self.exps):
             varstrs = ['%s**%.2g' % (var, x) if x != 1 else "%s" % var
-                       for (var, x) in sorted(exp.items(), key=vkSortBy) if x != 0]
+                       for (var, x) in exp.items() if x != 0]
```
bqpd: Hmm, so does this no longer sort by the order of the variables?

bqpd:
Does this now sort within posynomials and monomials alphabetically?

gpkit/nomials.py
whoburg: The very next line sorts `varstrs` by string comparison. Is there some other type of sorting we want/need?

Previously we were running into issues because `vkSortBy` attempted to compare the `descr` of two `VarKey` objects, and when those keys have units, Pint tries to compare two `Quantity` objects with different units, resulting in a `DimensionalityError`.

whoburg:
I think so -- if I understand the question correctly. In both `latex()` and in `__str__()`, sorting happens twice -- once inside the `for` loop, to sort the monomial varkey strings, and once via a call to `sorted` in the return, to sort by posynomial terms.

Am I understanding correctly? The intent was not to change the overall behavior, other than sorting inside each monomial by the string representation of the VarKey, instead of sorting by the (now removed) `vkSortBy`.

gpkit/nomials.py
bqpd: Derp, that sounds right.

bqpd:
That makes sense! I was just confused, I think.

bqpd:
:+1: for merging

Issue #273, opened 2015-05-28
use NomialData more often


whoburg:
Some functionality could be better implemented as methods of `PosyTuple` (e.g. substitution), or at least we could make methods that take in `PosyTuple` classes instead of  `varlocs`, `exps`, etc.

Renaming idea: `NomialData`.

`small_scripts.locate_vars` could be refactored as part of this. Everywhere it is called we should probably instead be calling the constructor of a `NomialData`.

Related to #226.

bqpd:
Substitution should happen in NomialData?

whoburg:
also `simplify_and_mmap`

bqpd:
This seems pretty closeable to me with the refactor

PR #282, opened 2015-06-29
(in progress) GP refactor!

bqpd:
Well, here's the first stab at a GP refactor. Three hundred lines of code down!

 ----------

**KNOWN ISSUES**

- [x] Variable sensitivities are out-of-order / incorrect
- [x] Tests are a giant mess
- [x] `model.py` is a bit of a mess, and the others should be cleaned up a bit
- [x] monomial maps don't deal with combined monomials properly (linear separation)

----------

The major change is to have users create 'Models' as an abstraction level _above_ GPs and SPs.

Models keep everything as signomials and dictionaries, making updating them easy and pythonic. GPs and SPs, on the other hand, no longer accept substitutions: any GP object should always return the same result (and SPs as well, if you start at the same initial condition).


bqpd:
Well, I know this fails tests, soooo @acdl-jenkins seems a bit out of the loop. @galbramc

galbramc:
It did fail. https://acdl.mit.edu/csi/job/gpkit/187/

I suspect you got the success email from commit by @whoburg https://acdl.mit.edu/csi/job/gpkit/186/, which was a success. I'm trying to sort out the message "Build finished. No test results found." I don't know why that is there.

whoburg:
:+1: for reducing # lines of code! (though some of the reduction seems to be deleted docstrings -- less good).

@bqpd are you getting a malloc crash when you try to run tests?

Thanks for the nice overall description above. Let me know when it's a good time to kick off an initial code review.

whoburg:
What's the plan for `NomialData` as in #273 ? Should we do that work on a separate branch? And should it happen before, after, or in parallel with the GP refactor?

bqpd:
Okay, this now works for vanilla GPs. There was a bug deep in how we handled the way substitution can change the ordering and number of monomials (the latter of which applies to the non-refactored version as well).

 I'm going through now and getting the rest of the tests working...

whoburg:
I wonder if the deep substitution bug is related to #313 ?

whoburg:
@bqpd once you get tests working, would it make sense to rebase this onto master?

bqpd:
Yup, it was!

bqpd:
Why rebase instead of merge?

whoburg:
Or merge -- master into this, right?

bqpd:
Yeah, or maybe something a bit more by hand. It'll be fun.

bqpd:
One test remaining in both `t_sub` and `t_user_scripts`, and then Signomials need their own monomial-mapping setup.

bqpd:
Signomials work now, better than before thanks to using the nearest-feasible heuristic.

However, variable sensitivities are still a bit odd.

whoburg:
@bqpd, let me know when it's code review time.

I'm not following the terms "monomial-mapping" and "mmaps". Can you elaborate?

bqpd:
Ready for code review!

I added a long comment in `models.py` to explain mmaps, reproduced below:

```
Remap monomials after substitution and simplification.
 each "mmap" is a list whose elements are either None
 (indicating that the monomial was removed after subsitution)
 or a tuple with:
     the index of the monomial they became after subsitution,
     the percentage of that monomial that they formed
     (their c/that monomial's final c)
 The monomial sensitivities from the GP/SP are in terms of this
 smaller post-substitution list of monomials, so we need to map that
 back to the pre-substitution list.
```

bqpd:
Also added proper tempdirs to the `mosek_cli` solver in this refactor.

gpkit/tests/t_sub.py
```diff
@@ -120,119 +118,119 @@ def test_vector_init(self):
...
-        def sorted_solve_array(sol):
-            return np.array([x[1] for x in
-                             sorted(sol["variables"].items(),
-                                    key=lambda x: x[0].name)])
-        a = sorted_solve_array(sf.solve(printing=False))
-        sol = gpl.solve(printing=False)
-        del sol["variables"]["dum"], sol["variables"]["dum"]
-        b = sorted_solve_array(sol)
-        self.assertTrue(all(abs(a-b)/(a+b) < 1e-7))
+    # def test_simpleaircraft(self):
```
bqpd: This test is contingent on how we want Model composition to work.

gpkit/user_scripts.py
```diff
@@ -5,49 +5,6 @@
 from .nomials import Posynomial
 from .variables import Variable, VectorVariable
 from .posyarray import PosyArray
-from .geometric_program import GeometricProgram
-
-
-def make_feasibility_gp(gp, varname=None, flavour="max"):
```
bqpd: This moved to being a method of GP. I think we might want a similar method for Model at some point.

bqpd:
retest this please

galbramc:
Unfortunately you will not get any testing until you have resolved all the merge conflicts.

gpkit/tests/t_sub.py
whoburg: Is there a straightforward way to get it back to a working state for now? It would be nice to know we didn't lose any functionality with the refactor.

bqpd:
Superseded by #317

Issue #269, opened 2015-05-23
Nans appear in GPSolutionArray when a VectorVariable element is substituted for


pgkirsch:
It seems that the solution isn't printing is because of the `nan's`. Notice that these `nan's` appear where the vector element value has been substituted, and it doesn't matter what the substituted value is.

So this issue has two parts:
- the vector variables have `nan's` where they should have the value of `eps`.
- the solution doesn't print when some of its values contain `nan's`

Code:
```
from gpkit.shortcuts import *
from numpy import ones

N = 6
L = 5.
dx = L/(N-1)

EI = Var("EI",10)

P = 100
p = Vec(N, "p", descr="Distributed load")
p = p.sub(p, P*ones(N))

V  = Vec(N, "V", descr="Internal shear")
M  = Vec(N, "M", descr="Internal moment")
th = Vec(N, "th", descr="Slope")
w  = Vec(N, "w", descr="Displacement")

eps = 1E-6 #something small

V[-1] = V[-1].sub(V[-1], eps)
M[-1] = M[-1].sub(M[-1], eps)
th[0] = th[0].sub(th[0], eps)
w[0]  = w[0].sub(w[0], eps)

objective = w[-1]

constraints = [EI*V.left[1:N]     >= EI*V[1:N]    + 0.5*dx*p.left[1:N]     + 0.5*dx*p[1:N],
               EI*M.left[1:N]     >= EI*M[1:N]    + 0.5*dx*V.left[1:N]     + 0.5*dx*V[1:N],
               EI*th.right[0:N-1] >= EI*th[0:N-1] + 0.5*dx*M.right[0:N-1]  + 0.5*dx*M[0:N-1],
               EI*w.right[0:N-1]  >= EI*w[0:N-1]  + 0.5*dx*th.right[0:N-1] + 0.5*dx*th[0:N-1]]

gp = GP(objective, constraints)

sol = gp.solve()

print sol.table()
```

Output:
```
In [17]: %run beam.py
Using solver 'mosek'
Solving for 20 variables.
Solving took 0.0283 seconds.

Cost
 0.781   

Variables
---------
EI : 10      

Constant sensitivities
----------------------
EI : -4  
```

If I inspect `sol`, the "second" solution seems to be:
```
'variables': {EI: array(10),
  w: array([        nan,  0.05125112,  0.18500127,  0.36625144,  0.57000162,
          0.78125184]),
[...]

PR #298, opened 2015-07-23
Allow nans to print in solution

pgkirsch:
To allow printing of `nans` in solution.

### DIFF
--- a/gpkit/small_scripts.py
+++ b/gpkit/small_scripts.py
@@ -189,7 +189,7 @@ def results_table(data, title, minval=0, printunits=True, fixedcols=True,
     decorated = [(bool(v.shape) if isinstance(v, Iterable) else False,
                  (varfmt % k),
                  i, k, v) for i, (k, v) in enumerate(data.items())
-                 if (np.max(abs(v)) >= minval) | (np.any(np.isnan(v)))]
+                 if (np.max(abs(v)) >= minval) or (np.any(np.isnan(v)))]
                 #                               allows nans to be printed 
     decorated.sort()
     for isvector, varstr, _, var, val in decorated:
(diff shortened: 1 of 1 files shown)


pgkirsch:
Related to issue 269

gpkit/small_scripts.py
```diff
@@ -189,7 +189,8 @@ def results_table(data, title, minval=0, printunits=True, fixedcols=True,
     decorated = [(bool(v.shape) if isinstance(v, Iterable) else False,
                  (varfmt % k),
                  i, k, v) for i, (k, v) in enumerate(data.items())
-                 if np.max(abs(v)) >= minval]
+                 if (np.max(abs(v)) >= minval) | (np.any(np.isnan(v)))]
+                #                               allows nans to be printed
```
whoburg: Let's change the pipe (`|`) to `or` -- I think it's more readable.

whoburg: @pgkirsch, agreed this list comprehension is dense and ugly. Refactoring table printing is definitely on the todo list! Just lower priority.

whoburg:
+1 for merging as a hotfix for #269.

Longer term, @bqpd, should we discuss whether `nan`s should *ever* be in a `GPSolutionArray`? I'd be in favor of raising exceptions long before table printing if there are any nans, and therefore removing this fix. If you agree, let's still merge this but open a new issue for eliminating nans for a later milestone.

whoburg:
PS note inline commit comment to address before merging.

pgkirsch:
Made the or change as request.

And yeah, you're right there really shouldn't be `nans` there in the first place- I guess I just carried away with actually doing something.

whoburg:
Actually doing something is great!! Continuous improvement...

whoburg:
might want to close this and make the `or` change on #300 ?

bqpd:
@whoburg good point. I haven't reached that point in the refactor. We could potentially use a sparse vector instead? I put `nan` there here just so it'd be obvious if something weird happened (even several steps of algebra later), which seems to have worked...

PR #301, opened 2015-07-24
Documentation improvements

whoburg:
Address a number of documentation (and some testing) issues for 0.2.0 release milestone.

### DIFF
--- a/gpkit/nomials.py
+++ b/gpkit/nomials.py
@@ -388,6 +388,7 @@ class Signomial(object):
         return self * other
 
     def __div__(self, other):
+        """Support the / operator in Python 2.x"""
         if isinstance(other, Numbers):
             return Signomial(self.exps, self.cs/other)
         elif isinstance(other, Monomial):
@@ -399,6 +400,7 @@ class Signomial(object):
             return NotImplemented
 
     def __truediv__(self, other):
+        """Support the / operator in Python 3.x"""
         return self.__div__(other)
 
     def __pow__(self, x):
@@ -479,6 +481,7 @@ class Monomial(Posynomial):
             return NotImplemented
 
     def __rtruediv__(self, other):
+        """rdiv for python 3.x"""
         return self.__rdiv__(other)
 
     def __pow__(self, other):
--- a/gpkit/nomials.py
+++ b/gpkit/nomials.py
@@ -583,9 +583,10 @@ class Constraint(Posynomial):
 
 
 class MonoEQConstraint(Constraint):
-    '''
-    TODO: Add docstring
-    '''
+    """
+    A Constraint of the form Monomial == Monomial.
+    Stored internally as a Monomial Constraint, (1 == self).
+    """
     def _set_operator(self, p1, p2):
         self.oper_l = " = "
         self.oper_s = " == "
--- a/gpkit/nomials.py
+++ b/gpkit/nomials.py
@@ -489,9 +489,14 @@ class Monomial(Posynomial):
 
 
 class Constraint(Posynomial):
-    '''
-    TODO: Add docstring
-    '''
+    """A constraint of the general form monomial > posynomial
+    Stored internally (exps, cs) as a single Posynomial (1 >= self)
+    Usually initialized via operator overloading, e.g. cc = y**2 >= 1 + x
+    Additionally stores input format (lhs vs rhs) in self.right and self.left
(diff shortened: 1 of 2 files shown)


whoburg:
retest this please

bqpd:
Ah, this is great, and I think ready to merge.

The only very minor comment is that the Test seems to be covering several bases; I wonder if it should be a couple of tiny tests?

gpkit/nomials.py
```diff
@@ -147,14 +155,25 @@ def value(self):
...
+            Variable to take derivative with respect to
+
+        Returns
+        -------
+        Signomial (or Posynomial or Monomial)
+        """
+        if wrt in self.varkeys:
+            wrt = self.varkeys[wrt]
+        elif isinstance(wrt, Monomial):
+            vks = list(wrt.exp)
```
whoburg: By the way, do we actually support differentiation with respect to a `Monomial`? How exactly is that defined? (It's currently missing from the docs). It seems like the current implementation assumes that the `Monomial` has only one `VarKey`?

gpkit/nomials.py
```diff
@@ -430,14 +449,30 @@ def __float__(self):
...
+
+    Arguments
+    ---------
+    Same as Signomial.
+    Note: Monomial historically supported several different init formats
+          These will be deprecated in the future, replaced with a single
+          __init__ syntax, same as Signomial.
+    """
     def __rdiv__(self, other):
+        """Divide other by this Monomial"""
```
whoburg: Had a thought -- would it be possible to remove the `elif isinstance(other, Monomial)` case from `Signomial.__div__`, so that we always fall back on `Monomial`'s `rdiv`? That seems conceptually cleaner, but I failed to get Python to fall back on `__rdiv__` in the simple attempts I made.

gpkit/nomials.py
```diff
@@ -454,9 +489,14 @@ def __pow__(self, other):
...
 
 class Constraint(Posynomial):
-    '''
-    TODO: Add docstring
-    '''
+    """A constraint of the general form monomial > posynomial
+    Stored internally (exps, cs) as a single Posynomial (1 >= self)
+    Usually initialized via operator overloading, e.g. cc = y**2 >= 1 + x
+    Additionally stores input format (lhs vs rhs) in self.right and self.left
+    Form is self.left >= self.right.
```
whoburg: The last line, `Form is self.left >= self.right.` is technically not correct due to the current logic for choosing left and right, correct?

@bqpd can you think of a better way to describe this? Any interest in removing this logic, simplifying things so that `left` is always what the user put on the greater than side?

whoburg:
@bqpd great point - I split up the tests in 37fb6d78decb6392b9a756d1fb0b298218577e6f.

This is ready to merge I think; I made some inline comments that might result in new tickets but should not hold up merging this PR.

gpkit/nomials.py
bqpd: Yeah, it's really differentiation with respect to a Variable.

gpkit/nomials.py
bqpd: Yeah, this should be possible! Just having Signomial.__div__ always return "NotImplemented" doesn't work?

gpkit/nomials.py
bqpd: Sounds good to me, this has been bothering me.

gpkit/nomials.py
whoburg: Removed Monomial case in fc52265d5605aa26d100b15dd40b347d3c5479c3

gpkit/nomials.py
whoburg: Unfortunately it's slightly tricky, but got this working in 252197dcbafdfcca4750fdca88c1956fba83ad3d.

The trickiness is that Python only calls `__rdiv__` for two objects of different type. For two objects of the same type, `__rdiv__` is never called. This means that `__div__` needs to support Monomial-Monomial division, for example. The solution I chose was to simply call `__rdiv__` from `__div__` for these cases.

gpkit/nomials.py
whoburg: Implemented in d00ef9654edc515d04203e416dd01f6583f5add8. I went with the standard form `left <= right` for all operator-overloaded initializations, in keeping with the notation/syntax of the Boyd GP tutorial.

Issue #312, opened 2015-08-04
cvxopt-only case for CI testing 


whoburg:
Heard from a new windows user that when only cvxopt is installed, we get a unit test failure related to mosek.

 - update continuous integration to include a machine with only cvxopt installed (?) @galbramc 
 - find what's causing this issue and fix it (@whoburg knows how, but waiting so we can reproduce this failure in CI)

whoburg:
![gpkit_tests_error](https://cloud.githubusercontent.com/assets/6042395/9064142/292eab7e-3a99-11e5-9f0f-75a31b127aac.png)

galbramc:
Would you also want to have testing that only tests mosek without cvxopt? This is what is actually happening on the windows test machine because I could not get cvxopt compiled on windows.

How did the user get cvxopt installed?!

whoburg:
I believe they used Pythonxy (https://code.google.com/p/pythonxy/), which includes an option to install cvxopt as part of the pythonxy install.

Testing with cvxopt (and no mosek) is the most pressing special case because cvxopt is open source, making it the only free option for non-academic users.

galbramc:
I currently have anaconda installed on the Windows machine. Do you suggest
I uninstall that and install pythonxy instead? Pythonxy says it does not
play nice with other python installs.

whoburg:
No, I think the current CI setup is great and we should stick with it. If we did pythonxy tests I would think they should be in addition. Does a virtualenv setup make sense here, or does that defeat the purpose of on-hardware CI?

galbramc:
Yeah we could setup virtualenv. I could probably use some help though from
someone, I have never used virtualenv, particularly on a windows machine. I
presume we would need to install pythonxy in the virtualenv every time we
run the test?

whoburg:
This might go completely against the spirit of continuous integration, but for this add-on unit testing case, I think it might be fine to just run the unit tests (on the same installation every time), to make sure we haven't introduced code that assumes mosek exists.

Does that seem reasonable? And more importantly, would it make this a lot easier?

galbramc:
I added another axis so we can try to test cvxopt by it self, and
cvxopt+mosek. I don't have an explicit mosek only test on linux and OSX,
but since I could not get cvxopt installed on the windows machine it is
only testing with mosek.

I can't easily test things at the moment until someone creates a pull
request and triggers the builds. This should of course fail unless the
mosek dependency in the tests is fixed.

whoburg:
[...]

PR #314, opened 2015-08-10
reintroduce error from #312; CI testing only; not to be merged

whoburg:


galbramc:
As for the test again, I had it hard coded to test on the master to make sure I had things setup correctly

whoburg:
retest this please

whoburg:
Worked perfectly

galbramc:
It didn't. Things passed on the linux machine when it should have failed. I don't understand though how it passed...

galbramc:
So I only get a failure on linux if I remove ~/.gpkit/expopt.so. So this is actually yet another reason why I need to serialize the testing since I now need to delete this file inbetween tests.

galbramc:
ok try testing this one more time

whoburg:
retest this please

whoburg:
retest this please

bqpd:
If you run mosek without having expopt.so, a failure will occur.

whoburg:
That's the point -- we're trying to test with only cvxopt installed and fail for tests that request `solver="mosek"`

bqpd:
We might be able to move expopt.so into the gpkit dir and remove ~/.gpkit entirely. We did that originally, then there was some bug such that adding ~/.gpkit was the easiest way to fix it, but now that gpkit_tmp doesn't need that folder it's worth re-examining.

bqpd:
@whoburg ... is the desired behaviour failure, then?

whoburg:
that would be excellent! 

the desired behavior for this PR is for cvxopt-only testing to fail on both Reynolds and Macys.

whoburg:
retest this please

bqpd:
The easiest fix for this would be for a cvxopt-only install to refuse to even try to solve with MOSEK.

whoburg:
This is just a temporary PR to get the CI setup correctly. Once it's catching the error introduced by this PR we will delete this PR unmerged.

galbramc:
The testing properly catches this error. We can remove this PR now.

whoburg:
hallelujah

PR #317, opened 2015-08-10
Testing merge of newgp

bqpd:


bqpd:
retest this please

bqpd:
retest this please pretty please?

bqpd:
retest this please

bqpd:
Passes tests on Jenkins.

whoburg:
Ok, should we merge this into newgp? I'm trying to figure out where to do my code review.

whoburg:
I think I'll just do the code review here, and then we can merge this into master

gpkit/__init__.py
```diff
@@ -90,11 +90,8 @@ def disable_signomials():
 from .variables import Variable, VectorVariable, ArrayVariable
 from .varkey import VarKey
 from .posyarray import PosyArray
-from .geometric_program import GeometricProgram
-from .signomial_program import SignomialProgram
```
whoburg: Why remove `GeometricProgram` and `SignomialProgram` but keep `GP` and `SP`? Isn't that the point of `gpkit.shortcuts`? I think that if we provide `GP` we should also provide the full class name `GeometricProgram`.

gpkit/_cvxopt.py
```diff
@@ -1,4 +1,4 @@
-def cvxoptimize_fn(options):
+def cvxoptimize_fn(options={}):
```
whoburg: Empty dicts and lists are dangerous default arguments. I see that it probably doesn't matter here though.

gpkit/_mosek/cli_expopt.py
```diff
@@ -27,13 +28,14 @@ def errorRemoveReadonly(func, path, exc):
         pass
 
 
-def imize_fn(filename):
-    folder = os.path.expanduser("~") + os.sep + "gpkit_tmp"
-    filename = folder + os.sep + filename
+def imize_fn(path=None):
```
whoburg: This function could use a one line docstring.

gpkit/_mosek/cli_expopt.py
```diff
@@ -27,13 +28,14 @@ def errorRemoveReadonly(func, path, exc):
...
+    if not path:
+        path = tempfile.mkdtemp()
+    filename = path + os.sep + "gpkit_mosek"
     os.environ['PATH'] = (os.environ['PATH'] + ':%s' %
                           settings["mosek_bin_dir"][0])
 
-    def imize(c, A, p_idxs, k):
+    def imize(c, A, p_idxs, k, *args, **kwargs):
```
whoburg: `*args` is unused, is this intended?

gpkit/_mosek/expopt.py
```diff
@@ -66,14 +66,14 @@ def __getattr__(self, attribute):
...
+                            "PRIM_INFEAS_CER",
                             "NEAR_OPTIMAL",
                             "NEAR_DUAL_FEAS",  # switched with below
                             "NEAR_PRIM_FEAS",
                             "NEAR_PRIM_AND_DUAL_FEAS",
-                            "NEAR_PRIM_INFEAS_CER",
                             "NEAR_DUAL_INFEAS_CER",  # switched with below
+                            "NEAR_PRIM_INFEAS_CER",
                             "INTEGER_OPTIMAL",
                             "NEAR_INTEGER_OPTIMAL"]
```
whoburg: The "switched with below" comments are now confusing -- did these changes undo those switches? Does this list match the mosek documentation (or some source somewhere)? Or is it our own creation?

gpkit/_mosek/expopt.py
```diff
@@ -121,7 +121,7 @@ def printcb(void, msg):
     return 0
 
 
-def imize(c, A, p_idxs, k):
+def imize(c, A, p_idxs, k, *args, **kwargs):
```
whoburg: unused `*args` here too.

gpkit/geometric_program.py
```diff
@@ -1,380 +1,99 @@
...
-    CLIENT = Client(timeout=0.01)
-    assert len(CLIENT) > 0
-    POOL = CLIENT[:]
-    POOL.use_dill()
-    print("Using parallel execution of sweeps on %s clients" % len(CLIENT))
-except:
-    POOL = None
-
-
-class GPSolutionArray(DictOfLists):
```
whoburg: Much cleaner to have this moved out of this file - yay.

gpkit/geometric_program.py
```diff
@@ -1,380 +1,99 @@
...
-                                   fixedcols=fixedcols,
-                                   minval=1e-2,
-                                   printunits=False)]
-        return "\n".join(strs)
-
-
-class GeometricProgram(Model):
-    """Holds a model and cost function for passing to solvers.
+class GeometricProgram(object):
+    """Prepares a collection of posynomials for a GP solver.
```
whoburg: I think this docstring could be improved -- is it fair to say that `gpkit.GeometricProgram` roughly represents the standard mathematical representation of a GP (with an A matrix, coefficients, etc)?

whoburg: And the docstring could even mention that at this level we do not separate out monomial equality constraints; those are implemented externally vie two inequality constraints. (if my understanding of that is correct).

gpkit/geometric_program.py
```diff
@@ -1,380 +1,99 @@
...
+        self.posynomials = [cost] + list(constraints)
+        self.cs = np.hstack((mag(p.cs) for p in self.posynomials))
+        if not all(self.cs > 0):
+            raise ValueError("GeometricPrograms cannot contain coefficients"
+                             " less than or equal to zero.")
+        self.exps = functools_reduce(add, (x.exps for x in self.posynomials))
+        self.varlocs, self.varkeys = locate_vars(self.exps)
+        # k [j]: number of monomials (columns of F) present in each constraint
+        self.k = [len(p.cs) for p in self.posynomials]
+        # p_idxs [i]: posynomial index of each monomial
```
whoburg: nice helpful inline comment

gpkit/geometric_program.py
```diff
@@ -1,380 +1,99 @@
...
+        for p_len in self.k:
+            p_idxs += [p_idx]*p_len
+            p_idx += 1
+        self.p_idxs = np.array(p_idxs)
+
+        self.A, missingbounds = genA(self.exps, self.varlocs)
+
+        if verbosity > 0:
+            for var, bound in sorted(missingbounds.items()):
+                print("%s has no %s bound" % (var, bound))
```
whoburg: Why not make verbosity boolean if it's used in this manner? Just trying to maintain the same kwarg names for user-friendliness? (that does seem valid...)

gpkit/geometric_program.py
```diff
@@ -1,380 +1,99 @@
...
-            If True (default), then prints out solver used and time to solve.
+        solver : str or function (optional)
+            By default uses one of the solvers found during installation.
+            If set to "mosek", "mosek_cli", or "cvxopt", uses that solver.
+            If set to a function, passes that function cs, A, p_idxs, and k.
+        verbosity : int (optional)
+            If greater than 0, prints solver name and solve time.
+        *args, **kwargs :
+            Passed to solver constructor and solver function.
+
```
whoburg: very nice docstring

gpkit/geometric_program.py
```diff
@@ -1,380 +1,99 @@
...
         -------
-        solution : dict
-            A dictionary containing the optimal values for each free variable.
+        result : dict
+            A dictionary containing the translated solver result.
         """
         if solver is None:
-            solver = self.solver
+            from . import settings
+            solver = settings['installed_solvers'][0]
```
whoburg: `__init__.py` sets `settings = {"installed_solvers": [""]}` in case of an `IOError` reading settings. Would it be cleaner to set `"installed_solvers": []`, and then move the conditional here to `if settings['installed_solvers']`?

(I'm just a little vexed about the `[0]` access relying on that empty string having been written).

Very minor, not urgent.

gpkit/geometric_program.py
```diff
@@ -1,380 +1,99 @@
...
+        *args, **kwargs :
+            Passed to solver constructor and solver function.
+
 
         Returns
         -------
-        solution : dict
-            A dictionary containing the optimal values for each free variable.
+        result : dict
+            A dictionary containing the translated solver result.
```
whoburg: I think this docstring should list and define the keys of the result dict.

I know this request is borderline feature creep...

gpkit/geometric_program.py
```diff
@@ -382,235 +101,182 @@ def _solve(self, solver=None, printing=True, skipfailures=False,
...
+        if verbosity > 0:
+            print("Solving took %.3g seconds." % (time() - self.starttime))
 
-        if self.sweep:
-            solution = self._solve_sweep(printing,
-                                         skipfailures,
-                                         allownonoptimal)
+        result = {}
+        result["variables"] = dict(zip(self.varlocs,
+                                   np.exp(solver_out['primal']).ravel()))
```
whoburg: I'm confused about something very basic. `self.varlocs` is a dict. I assume `solver_out['primal']` is a list or tuple. How on earth do we know that the order of the dict keys matches the order in `solver_out['primal']`?

gpkit/geometric_program.py
```diff
@@ -382,235 +101,182 @@ def _solve(self, solver=None, printing=True, skipfailures=False,
...
-        -------
-        solution : dict
-            A dictionary containing the array of optimal values
-            for each free variable.
-        """
-        solution = GPSolutionArray()
-        solution.gp = self
+            "max" (default) : Apply the same slack to all constraints and
+                              minimize that slack. Described in Eqn. 10
+                              of [Boyd2007].
```
whoburg: How does this handle monomial equality constraints? As far as I can tell they are not known to `GeometricProgram`, so this must apply slack to each each side of the MEC as well?

That's a little odd -- we might eventually need a way to run `flavour="max"`, but only on a subset of the Constraints.

gpkit/geometric_program.py
```diff
@@ -382,235 +101,182 @@ def _solve(self, solver=None, printing=True, skipfailures=False,
...
+        elif flavour == "product":
+            slackvars = VectorVariable(len(self.constraints), varname)
+            gp = GeometricProgram(slackvars.prod(),
+                                  (1/slackvars).tolist() +  # slackvars > 1
+                                  [constraint/slackvars[i]  # constraint <= sv
+                                   for i, constraint in enumerate(self.constraints)],
+                                  *args, **kwargs)
         else:
-            mapfn = map
+            raise ValueError("'%s' is an unknown flavour of feasibility." % flavour)
```
whoburg: haha nice

gpkit/geometric_program.py
```diff
@@ -382,235 +101,182 @@ def _solve(self, solver=None, printing=True, skipfailures=False,
...
+
+    Returns
+    -------
+        A : sparse Cootmatrix
+            Exponents of the various free variables for each monomial: rows
+            of A are monomials, columns of A are variables.
+        missingbounds : dict
+            Keys: variables that lack bounds. Values: which bounds are missed.
+    """
+
```
whoburg: Yes!! Dcostring!

gpkit/geometric_program.py
```diff
@@ -382,235 +101,182 @@ def _solve(self, solver=None, printing=True, skipfailures=False,
...
+                varsign = "both"
+
+        if varsign != "both":
+            if varsign == 1:
+                bound = "lower"
+            elif varsign == -1:
+                bound = "upper"
+            missingbounds[var] = bound
+
+    # add subbed-out monomials at the end
```
whoburg: would love a slightly more descriptive comment here, feature creep

gpkit/model.py
```diff
@@ -1,250 +1,424 @@
...
+from .variables import Variable
+
+try:
+    from IPython.parallel import Client
+    CLIENT = Client(timeout=0.01)
+    assert len(CLIENT) > 0
+    POOL = CLIENT[:]
+    POOL.use_dill()
+    print("Using parallel execution of sweeps on %s clients" % len(CLIENT))
+except:
```
whoburg: exception type?

gpkit/model.py
```diff
@@ -1,250 +1,424 @@
...
+    POOL = None
 
 
 class Model(object):
-    "Abstract class with substituion, loading / saving, and p_idx/A generation"
```
whoburg: Absolutely need to add a good docstring for Model, since it's the public-facing API now.

gpkit/model.py
```diff
@@ -1,250 +1,424 @@
...
+    model_nums = defaultdict(int)
+
+    def __init__(self, cost=None, constraints=[],
```
whoburg: empty list and empty dict defaults scare me.

gpkit/model.py
```diff
@@ -1,250 +1,424 @@
...
+                 substitutions={}, *args, **kwargs):
+        if cost is None:
+            if not hasattr(self, "setup"):
+                raise TypeError("Models can only be created without a cost"
+                                " if they have a 'setup' method.")
```
whoburg: need docstring on what setup is supposed to do

gpkit/model.py
```diff
@@ -1,250 +1,424 @@
...
+            try:
+                if "name" in kwargs:
+                    name = kwargs.pop("name")
+                else:
+                    name = self.__class__.__name__
+                setup = self.setup(*args, **kwargs)
+            except:
+                # TODO: pass error through!
+                print("The 'setup' method of this model had an error.")
+                raise TypeError
```
whoburg: would simply calling `raise` with no exception type here address the TODO?

gpkit/model.py
```diff
@@ -1,250 +1,424 @@
...
+        self.cost = cost
+        self.constraints = list(constraints)
+        self.substitutions = dict(substitutions)
+
+        if hasattr(self, "setup"):
+            # TODO: use super instead of Model?
+            k = Model.model_nums[name]
+            Model.model_nums[name] = k+1
+            name += str(k) if k else ""
+            for p in posynomials:
```
whoburg: Have models with `setup` been tested? I doubt it, because `posynomials` is undefined...

gpkit/model.py
```diff
@@ -1,250 +1,424 @@
...
+                for exp in p.exps:
+                    for k in exp:
+                        if "model" not in k.descr:
+                            newk = VarKey(k, model=name)
+                            exp[newk] = exp.pop(k)
+
+    @property
+    def variables(self):
+        variables = {}
+        for p in self.posynomials:
```
whoburg: has this been tested? I don't see a `posynomials` attribute in `Model`.

gpkit/model.py
```diff
@@ -1,250 +1,424 @@
...
-        if isinstance(constraints, Constraint):
-            constraints = [constraints]
-        for p in constraints:
-            self.constraints.remove(p)
-        self._gen_unsubbed_vars()
-
-    def _gen_unsubbed_vars(self, exps=None, cs=None):
+    # TODO: add get_item
+
+    def solve(self, solver=None, verbosity=2, skipfailures=True, *args, **kwargs):
```
whoburg: Probably need a docstring here -- how does this relate to `GeometricProgram.solve` from a user's point of view?

gpkit/model.py
```diff
@@ -1,250 +1,424 @@
...
+        #if not all([isinstance(c, Posynomial) for c in constraints]):
+        #    raise ValueError("'solve()' can only be called on models"
+        #                     " that do not contain Signomials.")
+        return self._solve("gp", solver, verbosity, skipfailures, *args, **kwargs)
+
+    def localsolve(self, solver=None, verbosity=2, skipfailures=True, *args, **kwargs):
+        #if all([isinstance(c, Posynomial) for c in constraints]):
+        #    raise ValueError("'localsolve()' can only be called on models"
+        #                     " that contain Signomials.")
```
whoburg: keeping commented out code for a reason?

gpkit/model.py
```diff
@@ -1,250 +1,424 @@
...
-        Returns
-        -------
-        No return value: the model is modified in place.
-        """
-        # look for sweep variables
-        found_sweep = []
+        self.unsubbed_cs = np.hstack((mag(p.cs) for p in posynomials))
+        self.unsubbed_exps = functools_reduce(add, (p.exps for p in posynomials))
+        (self.unsubbed_varlocs,
+         self.unsubbed_varkeys) = locate_vars(self.unsubbed_exps)
```
whoburg: side effects ahoy -- are side effects necessary here?

gpkit/model.py
```diff
@@ -1,250 +1,424 @@
...
+        #                     " that do not contain Signomials.")
+        return self._solve("gp", solver, verbosity, skipfailures, *args, **kwargs)
+
+    def localsolve(self, solver=None, verbosity=2, skipfailures=True, *args, **kwargs):
+        #if all([isinstance(c, Posynomial) for c in constraints]):
+        #    raise ValueError("'localsolve()' can only be called on models"
+        #                     " that contain Signomials.")
+        return self._solve("sp", solver, verbosity, skipfailures, *args, **kwargs)
+
+    def _solve(self, programType, solver, verbosity, skipfailures, *args, **kwargs):
```
whoburg: Add short doctoring to describe what this method does beyond calling `GP.solve`

gpkit/model.py
```diff
@@ -1,250 +1,424 @@
...
+                    return program, sol
+                except (RuntimeWarning, ValueError):
+                    return program, program.result
+
+            if POOL:
+                mapfn = POOL.map_sync
+            else:
+                mapfn = map
+
+            self.program = []
```
whoburg: so self.program is either a type `list` or a type `GeometricProgram`, depending on whether there are sweeps involved?

gpkit/model.py
```diff
@@ -1,250 +1,424 @@
...
+            self.program, subs, mmaps = self.formProgram(programType,
+                                                         posynomials,
+                                                         constants)
+            result = self.program.solve(*args, **kwargs)
+            sol = self.parse_result(result, subs, mmaps)
+            solution.append(sol)
+
+        solution.program = self.program
+        solution.toarray()
+        self.solution = solution
```
whoburg: side effect festival, not great...

gpkit/model.py
```diff
@@ -1,250 +1,424 @@
...
-                varlocs, exps, cs, subs = substitution(varlocs,
-                                                       self.varkeys,
-                                                       exps,
-                                                       cs,
-                                                       substitutions)
-            except KeyError:
-                # our new sub replaced the only previous sub
-                pass
+        self.substitutions.update(substitutions)
+        # edits in place...should it return a new model instead?
```
whoburg: At the very least, there should be a docstring describing exactly what it does.

gpkit/model.py
```diff
@@ -1,250 +1,424 @@
...
+
+    def formProgram(self, programType, signomials, subs):
```
whoburg: need docstring

gpkit/model.py
```diff
@@ -1,250 +1,424 @@
...
+                     for (var, locs) in self.unsubbed_varlocs.items()}
+        sensitivities["variables"] = sens_vars
+
+        # free-variable sensitivities must be < arbitrary epsilon 1e-4
+        for var, S in sensitivities["variables"].items():
+            if var in freevariables and abs(S) > 1e-4:
+                print("free variable too sensitive:"
+                      " S_{%s} = %0.2e" % (var, S))
+
+        localexp = {var: S for (var, S) in sens_vars.items() if abs(S) >= 0.1}
```
whoburg: magic `0.1`, what's going on here?

gpkit/model.py
```diff
@@ -1,250 +1,424 @@
...
+                            vardict[veckey] = np.full(var.descr["shape"], np.nan)
+                            vardict[veckey][idx] = vardict[var]
+
+                        del vardict[var]
+
+        # for veckey in veckeys:
+        #     # TODO: print index that error occured at
+        #     if any(np.isnan(variables[veckey])):
+        #         print variables
+        #         raise RuntimeWarning("did not fully fill vector variables.")
```
whoburg: is this commented code staying?

gpkit/model.py
```diff
@@ -1,250 +1,424 @@
...
+        return dict(cost=cost,
+                    constants=constants,
+                    sweepvariables=sweepvariables,
+                    freevariables=freevariables,
+                    variables=variables,
+                    sensitivities=sensitivities,
+                    localmodel=localmodel)
+
+
+def separate_subs(substitutions, varkeys, varlocs):
```
whoburg: add one-liner docstring for now

gpkit/shortcuts.py
```diff
@@ -1,12 +1,15 @@
...
 Arr = ArrayVariable
+Model = Model
 
-GP = GeometricProgram
-SP = SignomialProgram
+def GP(*args, **kwargs):
+    raise Exception("'Model' has replaced 'GP'. Please update your code!")
+
+def SP(*args, **kwargs):
+    raise Exception("'Model' has replaced 'GP'. Please update your code!")
```
whoburg: I think this is too harsh. Is the API backward compatible in any way? e.g. can I call `Model` with the same args as `GP`? if so why not set `GP=Model` and print a warning instead of raising?

gpkit/signomial_program.py
```diff
@@ -1,179 +1,184 @@
...
+
+class SignomialProgram(object):
+    """Prepares a collection of signomials for a SP solve.
+
+    Arguments
+    ---------
+    cost : Constraint
+        Signomial to minimize when solving
+    constraints : list of Posynomials
+        Constraints to maintain when solving (implicitly Signomials <= 1)
```
whoburg: Why are the constraints Posynomials? Shouldn't they be Signomial constraints?

gpkit/signomial_program.py
```diff
@@ -1,179 +1,184 @@
...
+            posy = Posynomial(p_exps, p_cs) if p_cs != [] else None
+            negy = Posynomial(n_exps, n_cs) if n_cs != [1] else None
+            self.posynomials.append(posy)
+            self.negynomials.append(negy)
+
+        if not self.negvarkeys:
+            raise ValueError("SignomialPrograms must contain coefficients"
+                             " less than zero.")
+
+    def solve(self, solver=None, verbosity=1, x0={}, reltol=1e-4,
```
whoburg: rename to localsolve as before?

gpkit/signomial_program.py
```diff
@@ -1,179 +1,184 @@
...
+
+        # parse the result and return nu's of original monomials from
+        #  variable sensitivities
+        nu = result["sensitivities"]["monomials"]
+        sens_vars = {var: sum([gp.exps[i][var]*nu[i] for i in locs])
+                     for (var, locs) in gp.varlocs.items()}
+        nu_ = []
+        for signomial in self.signomials:
+            for c, exp in zip(signomial.cs, signomial.exps):
+                var_ss = [sens_vars[var]*val for var, val in sens_vars.items()]
```
whoburg: why create this list every time? Could move outside for loop?

gpkit/signomial_program.py
```diff
@@ -1,179 +1,184 @@
...
+                  + " and %.3g seconds." % (time() - self.starttime))
+
+        # parse the result and return nu's of original monomials from
+        #  variable sensitivities
+        nu = result["sensitivities"]["monomials"]
+        sens_vars = {var: sum([gp.exps[i][var]*nu[i] for i in locs])
+                     for (var, locs) in gp.varlocs.items()}
+        nu_ = []
+        for signomial in self.signomials:
+            for c, exp in zip(signomial.cs, signomial.exps):
```
whoburg: exp is unused

gpkit/small_scripts.py
```diff
@@ -140,12 +141,16 @@ def locate_vars(exps):
     return dict(varlocs), dict(varkeys)
 
 
-def sort_and_simplify(exps, cs):
+def sort_and_simplify(exps, cs, return_map=False):
     "Reduces the number of monomials, and casts them to a sorted form."
```
whoburg: need better docstring including return types

gpkit/solution_array.py
```diff
@@ -0,0 +1,99 @@
...
+from .posyarray import PosyArray
+from .small_classes import Strings
+from .small_classes import DictOfLists
+from .small_scripts import results_table
+from .small_scripts import unitstr
+from .small_scripts import mag
+
+
+class SolutionArray(DictOfLists):
+    "DictofLists extended with posynomial substitution."
```
whoburg: I think this docstring should be expanded to include the keys we typically find in a SolutionArray (since many other methods and classes rely on that API).

gpkit/solution_array.py
```diff
@@ -0,0 +1,99 @@
...
+        if len(self) > 1:
+            subbeds = [p.sub(self.atindex(i)["sensitivities"]["variables"],
+                             require_positive=False) for i in range(len(self))]
+            assert not any([subbed.exp for subbed in subbeds])
+            return np.array([mag(subbed.c) for subbed in subbeds],
+                            np.dtype('float'))
+        else:
+            subbed = p.sub(self["sensitivities"]["variables"],
+                           require_positive=False)
+            assert isinstance(subbed, Monomial)
```
whoburg: has this been tested? I doubt it, because `Monomial` is not defined.

gpkit/tests/t_geometric_program.py
```diff
@@ -31,10 +32,9 @@ def test_trivial_gp(self):
         """
```
whoburg: should this file be renamed to test_model? And `TestGP` renamed to `TestModel`?

whoburg: @bqpd need to think carefully about what this file is really testing. Is it testing `Model` or `GP`? There is no `t_model` currently. Should probably start splitting up tests to match the class boundaries.

gpkit/tests/t_sub.py
```diff
@@ -120,119 +118,119 @@ def test_vector_init(self):
...
+    #
+    #     def sorted_solve_array(sol):
+    #         return np.array([x[1] for x in
+    #                          sorted(sol["variables"].items(),
+    #                                 key=lambda x: x[0].name)])
+    #     a = sorted_solve_array(sf.solve(verbosity=0))
+    #     sol = gpl.solve(verbosity=0)
+    #     del sol["variables"]["dum"], sol["variables"]["dum"]
+    #     b = sorted_solve_array(sol)
+    #     self.assertTrue(all(abs(a-b)/(a+b) < 1e-7))
```
whoburg: why commented out?

whoburg:
Looking good! Take a look at #320, which can be merged into this, and address inline comments.

gpkit/geometric_program.py
bqpd: Yup, that's correct.

gpkit/geometric_program.py
bqpd: Trying to maintain consistency! SPs and Models pass the verbosity arg - 1 to programs they create

gpkit/geometric_program.py
bqpd: Ah, good point. I forgot to check where that default was set!

gpkit/geometric_program.py
bqpd: No, I think that's a good point!

gpkit/geometric_program.py
bqpd: Because we entered them in according to varlocs order in genA

gpkit/geometric_program.py
bqpd: Yup. I think that should be done at a Model level.

gpkit/model.py
bqpd: There's several possible exceptions in that code, which is probably not the greatest thing.

gpkit/model.py
bqpd: I copy them immediately in the initializer. I just used them because it makes figuring out what type an argument is easier?

gpkit/model.py
bqpd: Yup. Part of the composability thing.

gpkit/model.py
bqpd: Yup!

gpkit/model.py
bqpd: Ha, yup, it's code from the old GP composability stuff, waiting for its own revamp.

gpkit/model.py
bqpd: Yeah, I'm wondering where we should have that error message. Right now it's in GP/SP.

gpkit/model.py
bqpd: Nope, it's bad old code. They're used in parse_result, but that should be cleaner.

gpkit/model.py
bqpd: Yup!

gpkit/model.py
bqpd: I just want to make sure we don't lose the program and solution if there's an error.

gpkit/model.py
bqpd: Well, yes, but any opinions on whether it should return a new model?

gpkit/model.py
bqpd: For the localmodel, determining what sensitivities we care about.

gpkit/model.py
bqpd: It's there in case we change our minds about nans in the solution.

gpkit/shortcuts.py
bqpd: I'm just worried about seeing code in four months that's still calling GP! But the warning sounds fine.

bqpd: Added them as a warning, but keeping these as functions instead of classes; don't want any "isinstance(, GP)"

gpkit/signomial_program.py
bqpd: Erp yes.

gpkit/signomial_program.py
bqpd: This is for consistency inside of Model, but we could do.

gpkit/signomial_program.py
bqpd: It is outside the for loops?

gpkit/signomial_program.py
bqpd: My mistake, should have been iterated over in the first line.

gpkit/__init__.py
bqpd: Do you still think (after reading the shortcuts file) that we should reimport these?

gpkit/_mosek/cli_expopt.py
bqpd: Yup, so that someone could call solve with arguments that a particular solver doesn't use and not have that solver die.

gpkit/_mosek/expopt.py
bqpd: This list is mosek code (mosek.h), except switched where indicated.

bqpd: Edited comments for clarity.

gpkit/_mosek/expopt.py
bqpd: Unused no more!

gpkit/geometric_program.py
bqpd: New default is "", because that's what's read if no solvers are found.

gpkit/geometric_program.py
whoburg: cool, sounds good.

gpkit/geometric_program.py
whoburg: Ok yeah - sounds like a good plan.

gpkit/model.py
whoburg: I see where you're coming from, but I think `None` is a better default for these -- I think it's pretty standard to look at the docstring to determine expected types if a kwarg has a None default.

gpkit/model.py
whoburg: yeah conceptually I would think these errors would come from Models. I would expect the GP `__init__` not to ever allow creation of a GP with signomials, and, slightly less obvious, the SP `__init__` could require at least one card carrying signomial.

gpkit/model.py
whoburg: In my opinion it's a tradeoff between:

Maintaining consistency with nomials (which are immutable, right?)

and

Speed, for certain applications (which we would need to profile)

Lacking the profile data I'm unsure what's best...

whoburg: How about renaming to `sub_update` ?

That's in line with set operations like `difference_update` (https://docs.python.org/2/library/sets.html).

gpkit/model.py
whoburg: Not critical now, but eventually I'd vote for moving the `0.1` and the `1e-4` to kwarg defaults.

gpkit/model.py
whoburg: fair -- add issue number to the comment if easy?

gpkit/shortcuts.py
whoburg: Great, thats fine. I'm ok with turning into a full on Exception for 0.4.0.

gpkit/signomial_program.py
whoburg: I really liked that the difference required a user to know the mathematical properties of their model! I'd vote for keeping the "specify solve/localsolve" correctly requirement, and implementing at the `Model` level.

But willing to back down if that doesn't match your vision.

gpkit/__init__.py
whoburg: I appreciate wanting to force designers to create Models.

But I wonder if we might get "mathematician" users who want to create and solve GPs directly, in their mathematical form. That would be the case for exposing `GeometricProgram`.

gpkit/model.py
bqpd: Sounds good!

gpkit/model.py
bqpd: Sounds good!

gpkit/model.py
bqpd: Done!

gpkit/model.py
bqpd: Actually, what about just requiring to user to use `model.substitutions.update`, which is identical?

gpkit/signomial_program.py
bqpd: Changed to localsolve

gpkit/_mosek/cli_expopt.py
bqpd: Now it has one. :)

gpkit/model.py
bqpd: Added one!

gpkit/tests/t_geometric_program.py
bqpd: Hmm, good point.

bqpd:
Ready for re-review!

gpkit/_mosek/cli_expopt.py
whoburg: +1

gpkit/solution_array.py
bqpd: It's now been tested with the example in the shiny new docstring.

gpkit/tests/t_geometric_program.py
bqpd: Renamed `t_geometric_program`, but kept TestGP, since it makes sense to distinguish the tests between GPs and SPs

bqpd:
How on earth do you find which test caused a Jenkins failure? Other than that I think we're good here.

bqpd:
I think the issues is it's not getting any test data, not that it's not running the tests. @galbramc ?

whoburg:
retest this please

whoburg:
I think it's occurring only on the cvxopt-only test machines (thank goodness we created those). I have a guess about the problem.

Here's the traceback
```Traceback (most recent call last):
File "test.py", line 1, in <module>
import gpkit.tests; gpkit.tests.run(xmloutput=True)
File "/Users/jenkins/workspace/gpkit/buildnode/macys/optimizer/cvxopt/gpkit/tests/__init__.py", line 1, in <module>
from .run_tests import run
File "/Users/jenkins/workspace/gpkit/buildnode/macys/optimizer/cvxopt/gpkit/tests/run_tests.py", line 23, in <module>
from gpkit.tests import t_geometric_program
File "/Users/jenkins/workspace/gpkit/buildnode/macys/optimizer/cvxopt/gpkit/tests/t_geometric_program.py", line 319, in <module>
setattr(test, "ndig", NDIGS[solver])
KeyError: 'c'
Build step 'Conditional step (single)' marked build as failure
[WARNINGS] Skipping publisher since build result is FAILURE
```

I think that, somewhere, the KeyError is coming from something along the lines of 

```
In [2]: [x for x in "cvxopt"]
Out[2]: ['c', 'v', 'x', 'o', 'p', 't']
```

whoburg:
Like from here:

```python
for testcase in TEST_CASES:
    for solver in settings["installed_solvers"]:
        if solver:
            test = type(testcase.__name__+"_"+solver,
                        (testcase,), {})
            setattr(test, "solver", solver)
            setattr(test, "ndig", NDIGS[solver])
            TESTS.append(test)
```

bqpd:
Parsing strings is hard.

bqpd:
retest this please

bqpd:
retest this please, please

bqpd:
It doesn't seem to be testing this...

galbramc:
Try asking one more time. I'm not on the white list that allows me to ask

galbramc:
The testing is setup to only check roughly every 5 min if you ask for a test. There is some randomness built on that 5 min interval as well

bqpd:
It tested! And passed. @whoburg, can we merge this?

whoburg:
CONGRATS!!

PR #320, opened 2015-08-11
pylint, imports, clean up

whoburg:


gpkit/model.py
```diff
@@ -166,7 +161,7 @@ def _solve(self, programType, solver, verbosity, skipfailures, *args, **kwargs):
 
         if sweep:
             if len(sweep) == 1:
-                sweep_grids = np.array(list(sweep.values()))
+                sweep_grids = np.array(sweep.values())
```
bqpd: This breaks in Python 3 I think

bqpd:
Looks good!

PR #321, opened 2015-08-11
allow zero substitutions into the denominator

### LINKED ISSUE #290: Setting variables to zero
As a user, I sometimes want to set variables in a problem to zero. I'm working on one right now -- I have a 3-segment mission defined by three parameters: R_outbound, T_loiter, and R_return. I want to try setting T_loiter to zero (including inside a widget). GPkit is willing to substitute 0 for T_loiter. But other variables in the problem, such as the loiter fuel weight, no longer have a lower bound. The solution where those variables become 0, however, makes physical sense.

Here is a minimum working example of substituting zero for a variable. GPkit substitutes 0 for `y`, the problem remains well-posed, and the optimal solution (with cost 4) is found.

```python
from gpkit import Variable, GP
x = Variable('x', value=4)
y = Variable('y', value=0)
z = Variable('z')

prob = GP(z, [z >= x + y])
sol = prob.solve()
print sol.table()
```

But here's an equivalent problem, with `y` masked by a dummy variable `t`:

```python
from gpkit import Variable, GP
x = Variable('x', value=4)
y = Variable('y', value=0)
z = Variable('z')
t = Variable('t')

prob = GP(z, [z >= x + t,
              t >= y])
sol = prob.solve()
print sol.table()
```
This results in a `t has no lower bound` warning, and the problem is not solved (solution status: unknown). Mathematically, the argument could be made that this is not a well-posed GP. But from a user's point of view, the solution (x=4, y=0, z=4, t=0) seems perfectly valid.

It's not always practical to require the modeler to move all the zero variables to posynomials, because they might appear in a long chain of monomial constraints, inside other models, etc. So it would be nice to have methods for automatically dealing with this.

We should be careful proceeding down this path, to make sure we retain our rigorous mathematical definition of what constitutes a valid GP. For now, I'm suggesting we implement a method, 

`gp_out = replace_unbounded_below_variables_with_zero(gp_in)`,

(better names desperately needed) that takes in a GP, substitutes zero for all unbounded below variables, recursively, until there are no unbounded below variables, (and removes constraints of the form 0 <= 0 if that's not already happening), and returns another GP. This could of course also be made a method of `GeometricProgram` that mutates the object in place, and could be default behavior inside `solve`, if we decide so. But for now, let's just get the capability in place so users can call it right before `solve()` and then sub zeros at will.

bqpd:


bqpd:
Currently not working on jenkins linux because of an older version of numpy...

bqpd:
Working on Jenkins, now.

gpkit/tools.py
```diff
@@ -7,6 +7,12 @@
 from .posyarray import PosyArray
 
 
+def zero_lower_unbounded(model):
```
whoburg: docstring

whoburg:
Ok, this is looking great, but I think `zero_lower_unbounded` should be recursive. Here's an example that won't solve with a single call to `zero_lower_unbounded `, but solves with two calls:

```python
from gpkit.tools import zero_lower_unbounded
x = Variable('x', value=4)
y = Variable('y', value=0)
z = Variable('z')
t1 = Variable('t1')
t2 = Variable('t2')

prob = Model(z, [z >= x + t1,
                 t1 >= t2,
                 t2 >= y])
zero_lower_unbounded(prob)
# zero_lower_unbounded(prob)
sol = prob.solve()
print sol.table()
```

galbramc:
The test failures here are a result the serialized testing. If we remove .gpkit the testing can be done in parallel and each commit should pass the tests.

gpkit/model.py
```diff
@@ -430,7 +430,8 @@ def formProgram(self, programType, signomials, subs, verbosity=2):
         for s in signomials:
             _, exps, cs, _ = substitution(s.varlocs, s.varkeys,
                                           s.exps, s.cs, subs)
-            if any((mag(c) != 0 for c in cs)):
+            notnan = ~np.isnan(cs)
+            if np.any(notnan) and np.any(cs[notnan] != 0):
```
whoburg: is this the implementation of "drop posynomials with only nan coefficients"?

I think that deserves an inline comment.

bqpd: Yup, will do.

bqpd: Done

gpkit/tools.py
bqpd: Done.

whoburg:
fix merge conflicts, otherwise looks great

PR #326, opened 2015-08-12
Use 'with' environment to enable signomials

### LINKED ISSUE #260: handle environments using with statements
as suggested by @bqpd, use of signomials would happen as follows:

```python
from gpkit import Variable
x = Variable('x')
y = Variable('y')
with gpkit.signomial_environment:
    sig = (x + y >= 1)
```

bqpd:
Closes #260

### DIFF
--- a/gpkit/__init__.py
+++ b/gpkit/__init__.py
@@ -76,9 +76,19 @@ SIGNOMIALS_ENABLED = False
 
 
 class enable_signomials(object):
-    "Class to put up and tear down signomial support in an instance of GPkit."
+    """Class to put up and tear down signomial support in an instance of GPkit.
 
-    def __enter__(self, enable=True):
+    Example
+    -------
+    >>> import gpkit
+    >>> x = gpkit.Variable("x")
+    >>> y = gpkit.Variable("y", 0.1)
+    >>> with enable_signomials():
+    >>>     constraints = [x >= 1-y]
+    >>> gpkit.Model(x, constraints).localsolve()
+    """
+
+    def __enter__(self):
         global SIGNOMIALS_ENABLED
         if enable:
             SIGNOMIALS_ENABLED = True
--- a/gpkit/__init__.py
+++ b/gpkit/__init__.py
@@ -75,16 +75,17 @@ enable_units()
 SIGNOMIALS_ENABLED = False
 
 
-def enable_signomials():
-    """Enables signomial support in a particular instance of GPkit."""
-    global SIGNOMIALS_ENABLED
-    SIGNOMIALS_ENABLED = True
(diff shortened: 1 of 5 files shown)


bqpd:
restest this please

gpkit/__init__.py
```diff
@@ -75,16 +75,26 @@ def enable_units():
...
+
+    Example
+    -------
+    >>> import gpkit
+    >>> x = gpkit.Variable("x")
+    >>> y = gpkit.Variable("y", 0.1)
+    >>> with enable_signomials():
+    >>>     constraints = [x >= 1-y]
+    >>> gpkit.Model(x, constraints).localsolve()
+    """
```
whoburg: nice!!

whoburg: is the idea of keeping the name the same to create backwards compatibility?

whoburg:
This is so much cleaner!!

Only questions are
1. Should the class be called `EnableSignomials`?
2. Related to above, does the old method of calling `enable_signomials()` as a method work after we merge this, or will this break that?

 - [ ] Note merge conflict to fix...

whoburg:
make sure this gets advertised in the 0.3.0 change log

bqpd:
This will break backward compatibility, but it doesn't have to; it's just a different way of working. Should I add backwards compatibility in?

whoburg:
If it's easy, I like the following:

1. change the class name of the `with` version to `EnableSignomials` (this follows standard class naming, right? or are there different conventions for environments?)
2. maintain the `enable_signomials` method (and `disable_signomials`) for compatibility, but add a deprecation warning explaining to use `with EnableSignomials` instead.

bqpd:
retest this please

bqpd:
Changed the name.

bqpd:
retest this please

bqpd:
@galbramc, this failed because of `FATAL: Could not checkout 57aa5bcd4259ca1915a9087dc547190d6097c31c`

bqpd:
retest this please

bqpd:
Yeah, it wasn't failing any tests, just failing to run. :-/

bqpd:
I think this is ready to merge.

galbramc:
So the error you got has something to do with a new commit added to the repo before all the different tests have checked out the current commit. Effectively, there is a newer command than the one that is supposed to be tested. I'm not sure why that results in this error though... I'm going to use pull request #329 to test a setting change in jenkins.

galbramc:
So I don't know if the change fixed the problem, but it didn't break anything. I would need you to make a commit while the testing is running to see if the problem is fixed.

PR #328, opened 2015-08-12
Update documentation

### LINKED ISSUE #327: Update documentation to match new syntax

bqpd:
Closes #327


whoburg:
Love it, looks great

whoburg:
Not merging because the commit comment "begin updating docs" makes it seem like there is more coming.

bqpd:
Yupyup.

bqpd:
Alright, this is the last step!

docs/source/advancedcommands.rst
```diff
@@ -1,6 +1,35 @@
 Advanced Commands
 *****************
 
+Feasibility Analysis
+====================
+
+If your Model doesn't solve, you can automatically find the nearest feasible version of it with the ``Model.feasibility()`` command, as shown below. The feasible version can either involve dividing all constraints by the smallest number possible, dividing each constraint by its own number and minimizing the product of those numbers, or changing each constant by the smallest total percentage possible.
+
```
whoburg: instead of "dividing constraints by", maybe "relaxing constraints by"?

bqpd: I like that, but I think we should also mention what that means; not all users will know that the whole constraint ends up on the left-hand side. ("dividing by" also fails to mention this, whoops. so maybe "relaxing" is better?)

whoburg: sure, "relaxing" is fine by me!

whoburg:
this looks great! @bqpd please go ahead and merge when you are finished pushing any changes.

docs/source/advancedcommands.rst
bqpd: updated!

Issue #274, opened 2015-06-05
SP heuristic finds incorrect optimum for simple SP


pgkirsch:
This example should solve to an optimum at x = 1, y = 1.

```
import gpkit
from gpkit.shortcuts import *

gpkit.enable_signomials() 

x = Var('x')
y = Var('y')

J = 0.01*((x - 1)**2 + (y - 1)**2) + (x*y - 1)**2

sp = SP(J)
print sp
sol = sp.localsolve(x0={x: 1.1, y: 1.1})

print sol.table()

gpkit.disable_signomials()
```

Output: 
```
In [1]: %run Mike_SP2.py
Optional Python units library (Pint) not installed;unit support disabled.
gpkit.GeometricProgram( # minimize
          -0.02*x + -0.02*y + -2*x*y + 0.01*x**2 + 0.01*y**2 + 1 + x**2*y**2,
          [   # subject to
          ],
          substitutions={  },
          solver="mosek")
Beginning signomial solve.
Using solver 'mosek'
Solving for 2 variables.
Solving took 0.159 seconds.
Solving took 5 GP solves.

Cost
 0.624   

Variables
---------
x : 0.789   
y : 0.789   

Constant sensitivities
----------------------
```
Possibly related issue: when we `print sp` it outputs `-0.02*x + -0.02*y + -2*x*y + 0.01*x**2 + 0.01*y**2 + 1 + x**2*y**2`. Notice that the constant term should be `1.02` instead of `1`.

whoburg:
Todo:

* [ ] fix this
* [ ] add a test
* [ ] also test `sol = sp.localsolve()` (no initial guess)

pgkirsch:
Update: The output appears to be different from when this ticket was first opened.

```
[philippekirschen@kirschens-MBP tests]$ ipython issue_274.py 
Optional Python units library (Pint) not installed; unit support disabled.
gpkit.GeometricProgram( # minimize
          -0.02*x + -0.02*y + -2*x*y + 0.01*x**2 + 0.01*y**2 + 1 + x**2*y**2,
          [   # subject to
          ],
          substitutions={  },
          solver="mosek")
Beginning signomial solve.
Using solver 'mosek'
Solving for 2 variables.
Solving took 0.191 seconds.
Solving took 5 GP solves.

Cost
----
 0.1433

Free variables
--------------
x : 0.789 
y : 0.789 

Swept variables
---------------

Constants
---------

Constant and swept variable sensitivities
[...]

PR #329, opened 2015-08-12
test #274

bqpd:
test #274

### DIFF
--- a/gpkit/signomial_program.py
+++ b/gpkit/signomial_program.py
@@ -116,7 +116,8 @@ class SignomialProgram(object):
 
         while (iterations < iteration_limit
                and (not (cost and prevcost)
-                    or abs(prevcost-cost)/(prevcost + cost) > reltol)):
+                    or abs(prevcost-cost)/(prevcost + cost) > reltol
+                    and cost < 1e-5):
             gp = self.step(x0, verbosity)
             self.gps.append(gp)  # NOTE: SIDE EFFECTS
 
--- a/gpkit/signomial_program.py
+++ b/gpkit/signomial_program.py
@@ -117,7 +117,7 @@ class SignomialProgram(object):
         while (iterations < iteration_limit
                and (not (cost and prevcost)
                     or abs(prevcost-cost)/(prevcost + cost) > reltol
-                    and cost < 1e-5):
+                    and cost > 1e-5)):
             gp = self.step(x0, verbosity)
             self.gps.append(gp)  # NOTE: SIDE EFFECTS
 
--- a/gpkit/tests/t_model.py
+++ b/gpkit/tests/t_model.py
@@ -233,17 +233,13 @@ class TestSP(unittest.TestCase):
         x = Variable('x')
         y = Variable('y')
         z = Variable('z')
-
-        gpkit.enable_signomials()
-        J = 1*((x - 1)**2 + (y - 1)**2) + (x*y - 1)**2
-
-        sp = Model(z, [z >= J])
+        with EnableSignomials():
+            J = 1*((x - 1)**2 + (y - 1)**2) + (x*y - 1)**2
+            sp = Model(z, [z >= J])
         sol = sp.localsolve("cvxopt", verbosity=0)
         self.assertAlmostEqual(sol["cost"], 0.0, 2)
         self.assertAlmostEqual(sol(x), 1.0, 2)
         self.assertAlmostEqual(sol(y), 1.0, 2)
-
-        gpkit.disable_signomials()
 
     def test_issue180(self):
         L = Variable("L")
--- a/gpkit/tests/t_model.py
+++ b/gpkit/tests/t_model.py
@@ -243,6 +243,22 @@ class TestSP(unittest.TestCase):
         m = Model(obj, constraints)
         m.localsolve(verbosity=0)
 
+    def test_issue274(self):
+        x = Variable('x')
+        y = Variable('y')
+        z = Variable('z')
+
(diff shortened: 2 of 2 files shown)


bqpd:
It works on the mac system, but none of the others? And it works on @pgkirsch's and my computers. Weird...

galbramc:
test this please

whoburg:
Those `ValueError`s on jenkins match the errors I'm seeing on my system.

bqpd:
Yerp. This is the first non-OS-specific bug I haven't been able to reproduce on my machine...which makes me suspect it might be a package version bug?

whoburg:
```python
In [1]: import numpy

In [2]: numpy.__version__
Out[2]: '1.9.2'
```

whoburg:
which reminds me -- 

```python
In [4]: import gpkit

In [5]: gpkit.__version__
AttributeError                            Traceback (most recent call last)
<ipython-input-5-4300fb7eb2bd> in <module>()
----> 1 gpkit.__version__

AttributeError: 'module' object has no attribute '__version__'
```

bqpd:
I've got the same numpy version.

Ack, we have to add another place to change the version number?

whoburg:
Not  big deal, that's feature creep as far as this issue is concerned.

bqpd:
Yup, put it in #332

galbramc:
On the linux machine

>>> numpy.__version__
'1.8.2'

on OS X

>>> numpy.__version__
'1.9.2'

whoburg:
@bqpd does this work for you on all operating systems you have access to?

whoburg:
bingo, it just occurred to me what might be going on. The optimal objective value for this is 0.0. I wonder if we're dealing with floating point precision making that objective slightly negative.

Indeed, if I add 0.1 to `J`, I get the correct `x = y = 1`, and the correct cost of 0.1.

I think the monomial approximation of a signomial about a point where the signomial takes on a negative value is not defined... ? But I need to think that through for more than the 15 seconds I just allotted to it...

whoburg:
That still leaves the issue of why this:

```python
import gpkit
from gpkit import Model, Variable

x = Variable('x')
y = Variable('y')
z = Variable('z')

with gpkit.EnableSignomials():
    J = 1*((x - 1)**2 + (y - 1)**2) + (x*y - 1)**2 + 0.1
    sp = Model(J, [])
    sol = sp.localsolve()

print sol.table()

# output

Beginning signomial solve.
Solving took 3 GP solves and 0.12 seconds.

Cost
----
 0.8643

Free Variables
--------------
x : 0.9059
y : 0.9059

Sensitivities
-------------


Cost
----
 0.8643

Free Variables
--------------
x : 0.9059
y : 0.9059

Sensitivities
-------------
```

gives a different answer than the almost equivalent

```python
import gpkit
from gpkit import Model, Variable

x = Variable('x')
y = Variable('y')
z = Variable('z')

with gpkit.EnableSignomials():
    J = 1*((x - 1)**2 + (y - 1)**2) + (x*y - 1)**2 + 0.1
    sp = Model(z, [z >= J])
    sol = sp.localsolve()

print sol.table()

# output

Beginning signomial solve.
Solving took 7 GP solves and 0.145 seconds.

Cost
----
 0.1

Free Variables
--------------
x : 1
y : 1
z : 0.1

Sensitivities
-------------


Cost
----
 0.1

Free Variables
--------------
x : 1
y : 1
z : 0.1

Sensitivities
-------------
```

whoburg:
and this discussion should really be moved back to #274

bqpd:
retest this please

galbramc:
@bqpd did you push two commits at once?

bqpd:
Yup.

bqpd:
retest this please

bqpd:
retest this please

galbramc:
It appears that jenkins tried to test each one of them, rather than just the most recent one... I didn't expect that... This is also the reason why the test are failing. You should be able to push multuiple commits and it should only test the most recent one. I'll look to see if there is a setting related to this, but we might need the webhooks to get it to only test the most recent commit.

bqpd:
@whoburg, this is passing now.

galbramc:
test this please (I made some changes to jenkins)

bqpd:
Looks like it didn't install cvxopt

galbramc:
I tried setting up a virtualenv with cvxopt installed, and then load it from the script. Not sure why it's not working. I can import cvxopt in the virtualenv from the command line

galbramc:
test this please (I'm giving up on the virtualenv for now and working on it later)

whoburg:
Sorry guys, I'm not convinced on the `1e-5`. As I understand it, this considers any SP with a cost less than 1e-5 converged?

First, I would have expected a cutoff of `> 0` to work. I still think the fundamental issue has to do with making monomial approximations of signomials with value ` <= 0`. I see no reason why a nonzero cutoff should be needed.

Changing the cutoff to 0 doesn't pass the unit test for this issue, though. I think that's because the cost cutoff only addresses the objective, but we need to address the constraints too. We'll need to think carefully about how to do that -- possibly in a separate issue.

For now, I've proposed a patch (will push it in a moment) that simply adds a small constant to `J` for this example.

I think we should leave handling signomials that evaluate to `<= 0` for either 0.5 or for 0.3.1.

whoburg:
this PR was originally just for testing. Is anyone's intent that it ever gets merged? If not let's close it.

bqpd:
re: not handling Signomials <= 0, sounds good

Issue #336, opened 2015-08-13
localsolve while loop not exiting when numerator and denominator both approach zero


pgkirsch:
Pulling this issue over from the discussion on #274 

```
import gpkit
from gpkit import Model, Variable
with gpkit.EnableSignomials():
    x = Variable('x')
    z = Variable('z')
    J = 0.01*(x - 1)**2
    sp = Model(z, [z >= J])
    sol = sp.localsolve(iteration_limit=50)


#output

Beginning signomial solve.
{x: 0.099503719020998152, z: 0.0094544313448217156}
{x: 0.30858339675657204, z: 0.0070641292715984388}
{x: 0.55138385186126571, z: 0.004137253654887871}
{x: 0.7558555212372341, z: 0.0020329156059758929}
{x: 0.88772160560639424, z: 0.0008742012957197896}
{x: 0.95412334675043098, z: 0.00034540249534560776}
{x: 0.98237655634014498, z: 0.00013081534457576836}
{x: 0.99340766018349569, z: 4.86639535899045e-05}
{x: 0.99755961776297863, z: 1.7977255827085814e-05}
{x: 0.99910015431588806, z: 6.6236742434287643e-06}
{x: 0.99966868273464993, z: 2.4381000785678302e-06}
{x: 0.99987807689494357, z: 8.9711476347932211e-07}
{x: 0.99995514181077461, z: 3.3005551447630321e-07}
{x: 0.9999834968924961, z: 1.2142408125806915e-07}
{x: 0.99999392875103243, z: 4.4669889151738637e-08}
{x: 0.99999776649946515, z: 1.6433196923939157e-08}
{x: 0.99999917833933127, z: 6.0454438375448823e-09}
{x: 0.9999996977276967, z: 2.2239956551815462e-09}
{x: 0.99999988880020207, z: 8.1816243629095626e-10}
{x: 0.99999995909187633, z: 3.0098516157414271e-10}
{x: 0.9999999849507416, z: 1.1072625650178e-10}
{x: 0.99999999446368726, z: 4.0733912983436912e-11}
{x: 1.0, z: 1.0, \fbox{0}: 1.0}
{x: 0.099503719020998152, z: 0.0094544313448217156}
{x: 0.30858339675657204, z: 0.0070641292715984388}
{x: 0.55138385186126571, z: 0.004137253654887871}
{x: 0.7558555212372341, z: 0.0020329156059758929}
{x: 0.88772160560639424, z: 0.0008742012957197896}
[...]

PR #337, opened 2015-08-13
Possible solution to address #336

pgkirsch:


gpkit/tests/t_model.py
```diff
@@ -311,6 +311,16 @@ def test_sp_initial_guess_sub(self):
                       "and variables failed")
         self.assertAlmostEqual(sol["cost"], 1, self.ndig)
 
+    def test_while_loop_exit_condition(self):
+        with EnableSignomials():
+            x = Variable('x')
+            z = Variable('z')
+            J = 0.01*(x - 1)**2
+            sp = Model(z, [z >= J])
+            sol = sp.localsolve(verbosity=0, iteration_limit=50)
```
bqpd: You could also check the length of `sp.program.gps` to make sure it's less than 50

pgkirsch: good idea

pgkirsch: do you mean just `sp.gps`? I get an AttributeError with `program` and `len(sp.gps)` gives me what I want (`22`).

pgkirsch: Oh sorry ignore that I know what you mean now

pgkirsch:
Added the extra assertion and a slightly improved (although perhaps still not acceptable) conditional

gpkit/tests/t_model.py
bqpd: Best practices might be to use either:

```python
m = Model(z, [z >= J])
sp = m.sp(Model(z, [z >= J]))
```

or 

```
sp = SignomialProgram(z, [z >= J])
```

pgkirsch: I thought `SignomialProgram` was deprecated...

bqpd: Naw, just SP. SignomialProgram no longer accepts substitutions or equality constraints, is all.

bqpd:
:+1:

bqpd:
@pgkirsch, now you can use all the ValueErrors you want!

gpkit/signomial_program.py
```diff
@@ -116,7 +116,9 @@ def localsolve(self, solver=None, verbosity=1, x0=None, reltol=1e-4,
 
         while (iterations < iteration_limit
                and (not (cost and prevcost)
-                    or abs(prevcost-cost)/(prevcost + cost) > reltol)):
+                    or abs(prevcost-cost)/(prevcost + cost) > reltol)
+               and (cost > 1e-10 or cost == None)):
```
bqpd: `None` is funky. and might always return False to ==; the preferred check is `cost is None`

Alternately, 

```and (not (cost and prevcost)
            or abs(prevcost-cost)/(prevcost + cost) > reltol
            and cost > 1e-10)
```
This doesn't need the None check because it comes after an `and`; if the first bit is False, Python won't even try the second bit.

Unrelatedly, to work on my Windows machine this needs an eps of `1e-6`

galbramc:
test this please

bqpd:
The proper solution is to define an error type for "program did/did not contain signomials"

PR #339, opened 2015-08-13
fix #338

### LINKED ISSUE #338: units not converted properly
Major major bug. I'm surprised our unit tests didn't catch this.

First I'll show the correct behavior. Here's a simple GP, run on an old version of GPkit:

```python
from gpkit import Variable, GP
x = Variable('x', value=3, units="ft")
y = Variable('y', units="m")
prob = GP(y, [y == x])
sol = prob.solve()
print sol.table()
```
`x` is in feet; `y` is in meters. The solution gets it right:
```

Using solver 'cvxopt'
Solving for 1 variables.
Solving took 0.00434 seconds.

Cost
----
 0.9144 [m] 

Free variables
--------------
y : 0.9144  [m] 

Swept variables
---------------

Constants
---------
x : 3  [ft] 

Constant and swept variable sensitivities
-----------------------------------------
x : 1 
```

And now for the same thing, run on the current master:
```python
x = Variable('x', value=3, units="ft")
y = Variable('y', units="m")
prob = Model(y, [y == x])
sol = prob.solve()
```
```

Using solver 'cvxopt'
Solving for 1 variables.
Solving took 0.0053 seconds.

Cost
----
 3 [m] 

Free Variables
--------------
y : 3  [m] 

Constants
---------
x : 3  [ft] 

Sensitivities
-------------
x : 1 
```
(hmm, 3 meters does not equal 3 feet).

@bqpd, aside from fixing this, what's the best targeted unit test for this?

bqpd:
fix #338

### DIFF
--- a/gpkit/nomials.py
+++ b/gpkit/nomials.py
@@ -85,6 +85,7 @@ class Signomial(object):
                     units = cs[0]/cs[0].magnitude
                     if units.dimensionless:
                         cs = [c * ureg.dimensionless for c in cs]
+                        units = ureg.dimensionless
                     cs = [c.to(units).magnitude for c in cs] * units
                     if not all([c.dimensionality == units.dimensionality
                                 for c in cs]):
@@ -590,9 +591,8 @@ class MonoEQConstraint(Constraint):
         # self.leq = self
         # self.geq = 1 / self
         # instead, we'll do this -- TODO improve
-        m = Monomial(self.exps, self.cs)
-        self.leq = m
-        self.geq = 1/m
+        self.leq = m1/m2
+        self.geq = m2/m1
 
     def __nonzero__(self):
         # a constraint not guaranteed to be satisfied
--- a/gpkit/nomials.py
+++ b/gpkit/nomials.py
@@ -590,8 +590,9 @@ class MonoEQConstraint(Constraint):
         # self.leq = self
         # self.geq = 1 / self
         # instead, we'll do this -- TODO improve
-        self.leq = m1/m2
-        self.geq = m2/m1
+        m = Monomial(self.exps, self.cs)
+        self.leq = m
+        self.geq = 1/m
 
     def __nonzero__(self):
         # a constraint not guaranteed to be satisfied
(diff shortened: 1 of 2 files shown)


whoburg:
I think this needs some tests. At least the example given in #338, and maybe additional ones that test directly on `Constraint` and `MonoEQConstraint` objects?

bqpd:
I really think it just needs the one test, that a Signomial isn't created with units of `ft/m`

gpkit/nomials.py
```diff
@@ -85,6 +85,7 @@ def __init__(self, exps=None, cs=1, varlocsandkeys=None,
                     units = cs[0]/cs[0].magnitude
                     if units.dimensionless:
                         cs = [c * ureg.dimensionless for c in cs]
+                        units = ureg.dimensionless
```
whoburg: why does this fix it? I'm a little lost reading this section of code.

bqpd: The unit [ft/m] has a `dimensionless` property of `True`, but traps the conversion multiplier by setting the units to be the specific [dimensionless] unit, the line below coerces all the cs to that, freeing the conversion multiplier. Pints, man.

whoburg: Should we use pint's `to` method instead of growing our own?

```python
In [3]: ureg = pint.UnitRegistry()

In [4]: x = 3 * ureg.feet

In [5]: x
Out[5]: <Quantity(3, 'foot')>

In [6]: y = 1 * ureg.meter

In [7]: x/y
Out[7]: <Quantity(3.0, 'foot / meter')>

In [8]: (x/y).to(ureg.dimensionless)
Out[8]: <Quantity(0.9144, 'dimensionless')>
```

(I've barely played with pint; feel free to tell me this is a bad idea.)

whoburg: aaaand, that's exactly what you do in the next line. Sorry. catching up.

gpkit/nomials.py
```diff
@@ -85,6 +85,7 @@ def __init__(self, exps=None, cs=1, varlocsandkeys=None,
...
                     cs = [c.to(units).magnitude for c in cs] * units
```
whoburg: this multiplies a list by units?

bqpd: Which converts it into an array.

PR #341, opened 2015-08-14
remove cost cutoff and update test

whoburg:
removes cost cutoff as described in #329 conversation. Note commented test we still need to get working as identified in #274 .


whoburg:
rebased to squash these commits for clarity, superseded by #343

PR #331, opened 2015-08-12
Add NomialData class, reorganization

bqpd:
Closes #335.

PR #351, opened 2015-08-14
pep8 and assorted improvements during code review

whoburg:


gpkit/geometric_program.py
```diff
@@ -43,7 +43,8 @@ def __init__(self, cost, constraints, verbosity=1):
         self.cost = cost
         self.constraints = constraints
         self.posynomials = [cost] + list(constraints)
-        NomialData.__init__(self, signomials=self.posynomials, simplify=False)
+        super(GeometricProgram, self).__init__(signomials=self.posynomials,
```
bqpd: This seems harder to read and less correct. We want to create a NomialData of the current object, not generic `super`

gpkit/nomials.py
```diff
@@ -94,7 +94,7 @@ def __init__(self, exps=None, cs=1, require_positive=True, **descr):
             except AssertionError:
                 raise TypeError("cs and exps must have the same length.")
 
-        NomialData.__init__(self, exps, cs, simplify=simplify)
+        super(Signomial, self).__init__(exps, cs, simplify=simplify)
```
bqpd: As above, I prefer calling `NomialData` explicitly.

whoburg:
@bqpd do you want to work on the merge conflicts, or should I?

bqpd:
These look mostly great! I can get the merge conflicts once we come to an agreement about `super` vs `NomialData`

gpkit/__init__.py
```diff
@@ -92,7 +92,7 @@ def __enter__(self):
         global SIGNOMIALS_ENABLED
         SIGNOMIALS_ENABLED = True
 
-    def __exit__(self, type, value, traceback):
+    def __exit__(self, type_, val, traceback):
```
whoburg: (this was to get pylint to stop complaining about overwriting `value` from the outer scope)

bqpd: ah, nice.

whoburg:
gotcha. I'm no expert, but I think calling `super` is certainly the preferred / correct way to implement inheritance. Is there a strong reason to deviate from that?

also, https://rhettinger.wordpress.com/2011/05/26/super-considered-super/

whoburg:
said another way, should these perhaps not inherit from `NomialData`, but instead store a `NomialData` in a `data` attribute?

bqpd:
It's great when: multi-inheritance, or changing inheritance.

But the idea here isn't quite regular inheritance it's just "superimpose this data structure on me".

bqpd:
They could, but that's a bit inelegant and takes `exps` and `cs` another `.` away from the users.

bqpd:
Merging this for now without super, we can move the discussion to #331

whoburg:
Don't you mean the idea here isn't quite regular inheritance it's just "`super`impose this data structure on me" ? (couldn't resist)

bqpd:
Pretty much :p

gpkit/geometric_program.py
whoburg: also, note this does create a NomialData of the current object -- that's why `self` is passed.

bqpd: Yeah, they're identical. I guess I don't really care that much. I just feel like every time I see this I'm going to wonder "what the heck is it doing?"

whoburg: yeah, Python's super calls are super clunky-looking.

bqpd: using super()

gpkit/nomials.py
bqpd: in the end, using super()

PR #355, opened 2015-08-15
automatic solution checking

whoburg:


gpkit/geometric_program.py
```diff
@@ -177,6 +183,55 @@ def solve(self, solver=None, verbosity=1, *args, **kwargs):
...
+        RuntimeWarning, if any problems are found
+        """
+        def _almost_equal(a, b):
+            "local almost equal test"
+            return abs(np.log(a/b)) < tol
+        cost = sol["cost"]
+        primal_sol = np.log(sol["variables"].values())
+        nu = sol["sensitivities"]["monomials"]
+        la = sol["sensitivities"]["posynomials"]
+        A = self.A.tocsr()
```
bqpd: Note: will fail without scipy

gpkit/geometric_program.py
```diff
@@ -177,6 +183,55 @@ def solve(self, solver=None, verbosity=1, *args, **kwargs):
...
+        for mi in self.m_idxs[1:]:
+            if primal_exp_vals[mi].sum() > 1 + tol:
+                raise RuntimeWarning("Primal solution violates constraint:"
+                                     " %s is greater than 1." %
+                                     primal_exp_vals[mi].sum())
+        # check dual sol
+        nu0 = nu[self.m_idxs[0]]
+        if not _almost_equal(nu0.sum(), 1.):
+            raise RuntimeWarning("Dual variables associated with objective"
+                                 "sum to %s, not 1" % nu0.sum())
```
bqpd: Add a space before "sum"

gpkit/geometric_program.py
whoburg: is there a better way to get A?

gpkit/geometric_program.py
whoburg: done!

gpkit/geometric_program.py
bqpd: Not right now, nope. Just wanted to note.

whoburg: So, to clarify: 

1. are we dependent on scipy?
2. Does this commit change the answer to #1?

bqpd: I think we already are with cvxopt, because converting lambdas to nus uses .todense() (which should be converted to .csr(), probably).

whoburg:
@bqpd  this is ready for review.

gpkit/geometric_program.py
```diff
@@ -174,8 +185,63 @@ def solve(self, solver=None, verbosity=1, *args, **kwargs):
                                  " in the 'result' attribute\n"
                                  "(model.program.result)"
                                  " and its raw output in 'solver_out'.")
-        else:
-            return result
+        
+        self.check_solution(result)
+        if verbosity > 1:
+            print ("solution checking took %.2g%% of solve time" %
```
bqpd: Nice! What are the ratios looking like so far?

gpkit/geometric_program.py
```diff
@@ -174,8 +185,63 @@ def solve(self, solver=None, verbosity=1, *args, **kwargs):
...
+        sol: dict
+            solution dict, same format as return type of solve()
+
+        Raises
+        ------
+        RuntimeWarning, if any problems are found
+        """
+        def _almost_equal(num1, num2):
+            "local almost equal test"
+            return num1 == num2 or abs((num1 - num2) / (num1 + num2)) < tol
```
bqpd: Does it make sens to have this be the same `tol` as is used for distance from anything other than 1? (I think it's only used for 1 below, just checking)

bqpd:
Looks good! A couple comments, and want to change the .todense in`geometric_program` to .tocsr

gpkit/geometric_program.py
whoburg: for some very simple problems I saw result packing taking around 30%, and solution checking around 2%. I doubt that's representative of real-scale problems though.

gpkit/geometric_program.py
whoburg: sorry, I'm not following... which `tol` are you talking about?  (there's only one used throughout `check_solution, as far as I know`)

whoburg:
retest this please

gpkit/geometric_program.py
bqpd: Yeah, and it's used both in _almost_equal and in one other inequality `< 1 + tol`. Let's talk about this in person.

bqpd:
I think this is ready to merge; any parts you particularly want me to examine before doing so?

whoburg:
nope, agreed it's ready -- fire away.

PR #356, opened 2015-08-15
capture stdout to get log from solvers

### LINKED ISSUE #277: getting raw solver output
Moving e-mail conversation to an issue. Want a way to view/obtain raw output from solvers.

@pgkirsch:
...Do we want to add being able to see the number of solver iterations as a feature,...? 

For my own future reference (and for anyone else who doesn’t know/has forgotten how to do this) I’m including how I found the number of mosek iterations. Please let me know if there is a neater way that I missed. The only method I could figure out how to see the number of iterations was to add the following line before the return statement in `gpkit._mosek.expopt.printcb()`

`print msg[:-1]`

which printed the full solver log to the command line (big thank you to whoever wrote the docstring for `printcb`). Maybe the potential feature I mentioned above could be as simple as an optional argument to `solve()` that switches the statement above on?

@whoburg:
Yes, agreed we want this feature. @bqpd may have suggestions on implementation, but what you outlined sounds like a great start. Note we have to make sure it works cleanly independent of which solver is being used.

@bqpd:
Uncommenting that line seems like the right way to go! Ideally this would be controlled by the "settings" dictionary that is passed to the solver-function-constructor.

In particular, the MOSEK stream messager could have filters or flags of what to print, from everything (what you did) to nothing (the commented-out default).

bqpd:
Closes #277

### DIFF
--- a/gpkit/_mosek/expopt.py
+++ b/gpkit/_mosek/expopt.py
@@ -115,7 +115,7 @@ def printcb(void, msg):
     result : int
       0 indicates success
     """
-
+    print msg[:-1]
     return 0
 
 
--- a/gpkit/small_classes.py
+++ b/gpkit/small_classes.py
@@ -71,8 +71,7 @@ class SolverLog(list):
 
     def write(self, writ):
         if writ != "\n":
-            if writ and writ[-1] == "\n":
-                writ = writ[:-1]
+            writ = writ.rstrip("\n")
             self.append(writ)
         if self.verbosity > 0:
             self.original_stdout.write(writ)
--- a/gpkit/small_classes.py
+++ b/gpkit/small_classes.py
@@ -51,7 +51,7 @@ class CootMatrix(CootMatrixTuple):
         return self.tocoo().todia()
 
 
-class count(object):
+class Counter(object):
 
     def __init__(self):
         self.start = -1
@@ -59,6 +59,16 @@ class count(object):
     def __call__(self):
         self.start += 1
         return self.start
+
+
+class SolverLog(list):
+    "Adds a `write` method to list so it's file-like and can replace stdout."
+
+    def write(self, writ):
+        if writ != "\n":
+            if writ and writ[-1] == "\n":
+                writ = writ[:-1]
+            self.append(writ)
 
 
 class DictOfLists(dict):
--- a/gpkit/small_classes.py
+++ b/gpkit/small_classes.py
@@ -66,15 +66,18 @@ class Counter(object):
 
 class SolverLog(list):
     "Adds a `write` method to list so it's file-like and can replace stdout."
-    verbosity = 0
-    original_stdout = 0
+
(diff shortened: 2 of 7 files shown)


bqpd:
My thinking is that by capturing stdout, maybe we don't need those patches to the MOSEK solvers???

bqpd:
retest this please

bqpd:
retest this please

whoburg:
any idea what's up with the testing? CI seems to be reporting a merge conflict

bqpd:
test this please

bqpd:
It's probably still behind by 1

bqpd:
Yup.

gpkit/_mosek/expopt.py
```diff
@@ -115,7 +115,7 @@ def printcb(void, msg):
     result : int
       0 indicates success
     """
-
+    print msg[:-1]
```
whoburg: always? what about verbosity?

bqpd:
Argh, looks like the C code doesn't use / inherit the same stdout, so we still need those patches. :(

gpkit/_mosek/expopt.py
bqpd: ;D

bqpd: Print statements during a solver call are captured and saved to the log.

whoburg:
does this really close #277? How should a user get access to the solver output?

whoburg:
oh I see, just look at `gp.solver_log`?

bqpd:
Yup! Or now, just set verbosity higher.

gpkit/small_classes.py
```diff
@@ -64,6 +64,16 @@ def __call__(self):
...
 
+class SolverLog(list):
+    "Adds a `write` method to list so it's file-like and can replace stdout."
+
+    def write(self, writ):
+        if writ != "\n":
+            if writ and writ[-1] == "\n":
+                writ = writ[:-1]
+            self.append(writ)
+
```
whoburg: haven't tested, but I think the following simpler implementation will work:

```python
def write(self, writ):
    writ = writ.strip("\n")
    if writ:
        self.append(writ)
```

bqpd: I don't want to append strings that _started_ empty, is the trick.

whoburg: Did you mean to say something else?

Mine won't append `""`.

Yours actually will append `""`...

bqpd:
Ready for a code review!

gpkit/small_classes.py
bqpd: I edited my comment after I realized that, sorry about the collision.

bqpd:
That'd be, like 4 lines of code to __init__ list, too...

bqpd:
Yup!

whoburg:
retest this please

whoburg:
this is a great fix!

PR #360, opened 2015-08-21
(in progress) Signomial constraints

whoburg:
This is a refactor of how we handle signomial constraints. The major change is that instead of treating signomial constraints as `s(x) <= 1`, where `s(x)` is the signomial, we will treat signomial constraints as `s(x) <= 0`, where `s(x) = p(x) - n(x)` is the difference of two posynomials. Then, the signomial constraints take the form `p(x) <= n(x)`, and the signomial programming heuristic is to repeatedly lower bound `n(x)` with a monomial. This solves several issues involving units and convergence; it is the clean way to implement SP.

Changes include:
 - assorted fixes and improvements
 - effort to move methods into classes where it makes sense
 - Separate classes for `Constraint` (posynomial <= 1) and `SignomialConstraint` (signomial <= 0).

This currently fails the unit tests on two tests where we `solve` a `Model` that is an SP before substitution, but becomes a GP after substitution.

@bqpd, this is ready for an initial review, and I'd like your help with fixing the remaining unit tests. I've spend awhile trying various routes to getting this working, and all seem to involve some major refactoring. I'd like your input on the best route.

We have a lot of methods that parse lists called `signomials`, but those now contain `Constraint` and `SignomialConstraint` objects; I think we need to clean up some of the logic.

Regarding the failing unit tests, my current vision is that if a `Model` containing signomials becomes GP-compatible after substitution, then it will solve with `solve`.


gpkit/model.py
```diff
@@ -139,7 +138,7 @@ def allsubs(self):
     def signomials_et_al(self):
         "Get signomials, unsubbed, allsubs in one pass."
         signomials = self.signomials
-        unsubbed = NomialData(nomials=signomials)
+        unsubbed = self.unsubbed
```
bqpd: Nix! This is an extra line of code (and property) for slightly greater efficiency.

gpkit/nomials.py
```diff
@@ -425,7 +442,28 @@ class Posynomial(Signomial):
           These will be deprecated in the future, replaced with a single
           __init__ syntax, same as Signomial.
     """
-    pass
+    def __le__(self, other):
+        if isinstance(other, (Signomial, PosyArray)):
+            # fall back on other's __ge__
```
bqpd: Might be good to note why we let the other handle it? (v. minor)

gpkit/nomials.py
```diff
@@ -455,6 +493,21 @@ def __pow__(self, other):
         else:
             return NotImplemented
 
+    # Monomial.__le__ falls back on Posynomial.__le__
+
+    def __ge__(self, other):
+        if isinstance(other, Signomial) and not isinstance(other, Posynomial):
```
bqpd: use `other.any_positive_cs` instead of `not isinstance(other, Posynomial)`

gpkit/nomials.py
```diff
@@ -569,4 +601,80 @@ def __bool__(self):
         return self.__nonzero__()
 
 
+class SignomialConstraint(Signomial):
+    """A constraint of the general form posynomial >= posynomial
+    Stored internally (exps, cs) as a single Signomial (0 >= self)
```
bqpd: `self < =0`

gpkit/nomials.py
```diff
@@ -569,4 +601,80 @@ def __bool__(self):
...
+    Usually initialized via operator overloading, e.g. cc = (y**2 >= 1 + x - y)
+    Additionally retains input format (lhs vs rhs) in self.left and self.right
+    Form is self.left >= self.right.
+    """
+    def __str__(self):
+        return str(self.left) + self.oper_s + str(self.right)
```
whoburg: we've agreed to make pretty constraint printing a fragile property, so after substituting a SignomialConstraint, it will become either a posynomial (implicitly <= 1) or a signomial (implicitly <= 0). Need to implement this.

gpkit/signomial_program.py
```diff
@@ -66,8 +69,16 @@ def __init__(self, cost, constraints, verbosity=2):
...
+            if isinstance(sig, Posynomial):
+                posy = Posynomial(p_exps, p_cs) if p_cs != [] else None
+                negy = Posynomial(n_exps, n_cs) if n_cs != [1] else None
+                assert posy
+                assert not negy
+            else:
+                posy = Posynomial(p_exps, p_cs) if p_cs != [] else None
+                negy = Posynomial(n_exps, n_cs) if n_cs != [] else None
+                assert posy
+                assert negy
```
bqpd: Everything is going to be a constraint, and this separation should happen in those constraints `__init__`

gpkit/model.py
whoburg: done

gpkit/nomials.py
whoburg: replaced with `if isinstance(other, Numbers + (Monomial,))`.

gpkit/nomials.py
whoburg: done

gpkit/nomials.py
whoburg: done

whoburg:
@bqpd, I'm still working on cleaning things up here, but I'm running into one problem. This currently fails unit tests because `simplify_and_mmap` tries to create a `Signomial` with negative cs. What I'd really like to do there is call `s.sub`, which for the `SignomialConstraint` case will do something nice with the current unit test failures (they get cast to Constraint post-substitution).

However, my attempts at calling `s.sub` inside `simplify_and_mmap` all lead to completely breaking all the unit tests; I think it's difficult because the sub needed here needs to not simplify? Not sure.

Everything I've tried starts taking me down the road of a complete substitution refactor. Creative solutions short of that welcome.

gpkit/nomial_data.py
```diff
@@ -43,15 +44,65 @@ def __init__(self, exps=None, cs=None, nomials=None, simplify=True):
...
-            self._hash = hash(tuple(zip(self.exps, tuple(self.cs))))
-        return self._hash
+            self._hashvalue = hash(tuple(zip(self.exps, self.cs)))
+        return self._hashvalue
 
     def __repr__(self):
-        return "gpkit.%s(%s)" % (self.__class__.__name__, str(self._hash))
+        return "gpkit.%s(%s)" % (self.__class__.__name__, hash(self))
+
+    def _get_varkey(self, var):
```
bqpd: Is this used anywhere?

gpkit/nomials.py
```diff
@@ -145,27 +144,45 @@ def diff(self, wrt):
...
         Signomial (or Posynomial or Monomial)
         """
-        if wrt in self.varstrs:
-            wrt = self.varstrs[wrt]
-        elif not isinstance(wrt, VarKey):
-            wrt = wrt.varkey
-        exps, cs = diff(self, wrt)
-        return Signomial(exps, cs, require_positive=False)
+        deriv = super(Signomial, self).diff(wrt)
+        return Signomial(exps=deriv.exps, cs=deriv.cs)
```
bqpd: Not sure that the exp=, cs= are necessary, but don't mind 'em...

gpkit/nomials.py
```diff
@@ -558,4 +610,54 @@ def __bool__(self):
...
+        pgt, plt = (left, right) if oper_ge else (right, left)
+
+        from . import SIGNOMIALS_ENABLED
+        if not SIGNOMIALS_ENABLED:
+            raise TypeError("Cannot initialize SignomialConstraint "
+                            "without SignomialsEnabled.")
+
+        p = plt - pgt
+
+        super(SignomialConstraint, self).__init__(p)
```
bqpd: Lots of duplicated code below...

gpkit/nomials.py
```diff
@@ -558,4 +610,54 @@ def __bool__(self):
...
+            If true, form is left >= right; otherwise, left <= right.
+
+        Note: Constraints initialized via operator overloading always take
+              the form left >= right, e.g. (x <= y) becomes (y >= x).
+
+        Note: Unlike Constraints, SignomialConstraints have units.
+        """
+        left = Signomial(left)
+        right = Signomial(right)
+        pgt, plt = (left, right) if oper_ge else (right, left)
```
bqpd: Duplicated code.

gpkit/tests/t_nomials.py
```diff
@@ -216,6 +216,20 @@ def test_init(self):
...
+        x = Variable('x')
+        y = Variable('y')
+        self.assertTrue((1 + x) == (1 + x))
+        self.assertFalse((1 + x) == 2*(1 + x))
+        self.assertFalse((1 + x) == 0.5*(1 + x))
+        self.assertFalse((1 + x) == (1 + y))
+        x = Variable('x', value=3)
+        y = Variable('y', value=2)
+        # TODO should the next line work?
+        # self.assertEqual(1 + x**2, 4 + y + y**2)
```
bqpd: This should definitely not work, unless you call `.value`!

bqpd:
retest this please

bqpd:
retest this please

bqpd:
Looks good to me.

gpkit/nomial_data.py
whoburg: Yeah, it's used in `NomialData.diff`, and it's intended to be used more (and perhaps become public) as we refactor coercion to VarKey.

whoburg: E.g. we might use it to fix #371 ?

gpkit/model.py
```diff
@@ -126,30 +126,30 @@ def signomials(self):
...
-        return signomials, unsubbed, allsubs
+        return signomials, beforesubs, allsubs
 
     @property
     def constants(self):
         "All constants (non-sweep substitutions) currently in the Model."
-        _, unsubbed, allsubs = self.signomials_et_al
-        return get_constants(unsubbed, allsubs)
+        _, beforesubs, allsubs = self.signomials_et_al
+        return get_constants(beforesubs, allsubs)
```
whoburg: `beforesubs` seems much more descriptive, I like it.

gpkit/model.py
```diff
@@ -284,8 +286,9 @@ def _solve(self, programType, solver, verbosity, skipfailures,
         ValueError if programType and model constraints don't match.
         RuntimeWarning if an error occurs in solving or parsing the solution.
         """
-        signomials, unsubbed, allsubs = self.signomials_et_al
-        sweep, linkedsweep, constants = separate_subs(unsubbed, allsubs)
+        signomials, beforesubs, allsubs = self.signomials_et_al
+        beforesubs.signomials = signomials
```
whoburg: remind me what's going on here with setting the signomials attribute?

gpkit/model.py
```diff
@@ -544,3 +546,53 @@ def form_program(programType, signomials, verbosity=2):
...
+        if np.any(notnan) and np.any(cs[notnan] != 0):
+            exps, cs, smap = simplify_exps_and_cs(exps, cs, return_map=True)
+            if s.any_nonpositive_cs:
+                negative_c_count = (cs <= 0).sum()
+                if negative_c_count == 0:
+                    raise RuntimeWarning("""Infeasible SignomialConstraint.
+
+    %s became infeasible  when all negative terms were substituted out.""" % s)
+                elif negative_c_count == 1:
+                    # turn it into a Posynomial constraint
```
whoburg: I'm confused here because our implementation seems to assume that `signomials` is really a list of constraints. Is this function also intended to work with raw `Signomial` objects (non-constraints), as the docstring suggests? If so, I'm confused about how those are handled.

gpkit/model.py
```diff
@@ -544,3 +546,53 @@ def form_program(programType, signomials, verbosity=2):
...
+                    div_exp = exps.pop(idx)
+                    div_mmap = smap.pop(idx)
+                    cs /= -cs[idx]
+                    cs = np.hstack((cs[:idx], cs[idx+1:]))
+                    exps = tuple(exp-div_exp for exp in exps)
+                    smap = [mmap-div_mmap for mmap in smap]
+            signomials_.append(Signomial(exps, cs, simplify=False))
+            smaps.append(smap)
+        else:
+            smaps.append([])
```
whoburg: inline comment explaining the empty list append would be useful

gpkit/nomial_data.py
```diff
@@ -66,7 +117,7 @@ def __eq__(self, other):
         return True
 
 
-def sort_and_simplify(exps, cs, return_map=False):
+def simplify_exps_and_cs(exps, cs, return_map=False):
```
whoburg: Is the docstring still correct? It says `mmap` is a list of tuples...

gpkit/nomials.py
whoburg: once we refactor (i.e. remove) all the ways to create a `Signomial`, I'm hoping that `Signomial(deriv)` will work here. (deriv is a NomialData).

gpkit/nomials.py
whoburg: True, and agreed it would be nice to eliminate.

Unfortunately we can't have `SignomialConstraint` and `PosynomialConstraint` inherit from a generic `Constraint`, because one needs to inherit from `Signomial` and the other from `Posynomial`, so we'd end up in Python multiple inheritance hell.

We could consider replacing `left` and `right` with `sgt` and `slt`, which would constrain the user's options in terms of pretty printing. I suspect we might still end up with some duplication though.

Another ugly option would be to check the `__class__` in `Signomial.__str__`, so that all the constraint printing code goes there.

gpkit/small_scripts.py
```diff
@@ -1,45 +1,8 @@
...
-    m0 = 1
-    from . import units as ureg
-    for vk in p.varlocs:
-        units = vk.descr.get("units", 1) if ureg else 1
-        e = mag(x0[vk]*units * p.diff(vk).sub(x0, require_positive=False).c/p0)
-        exp[vk] = e
-        m0 *= (x0[vk]*units)**e
-    return p0/m0, exp
-
-
```
whoburg: loving all the red here

gpkit/tests/t_nomials.py
whoburg: great, removed the todo and changed to comparing value.

whoburg:
@bqpd, my code review complete. Check out a few inline comments; I'm otherwise ready to merge.

gpkit/model.py
bqpd: We use it to generate a p_idxs later...

gpkit/model.py
bqpd: Yup! they're handled just the same; nothing in here cares about anything the SignomialConstraint overrides.

gpkit/nomial_data.py
bqpd: Fixed!

gpkit/model.py
bqpd: Done!

gpkit/model.py
bqpd: It's used & referenced in solution_array.parse_result

bqpd:
Check that last commit, then merge!

bqpd:
retest this please

gpkit/model.py
whoburg: Ok, I still find it strange that the NomialData object doesn't already have all the info needed downstream, but I think that's outside the scope of this PR, and certainly lower priority than a lot of other issues.

gpkit/model.py
whoburg: So the "`# turn it into a Posynomial constraint`" is also valid for Signomial objects?

And what about "`# This constraint is being removed...`"?

bqpd: If it's all nans or 0s (see the "if" above), the constraint is removed.

whoburg: My point is just that the inline comments assume `signomials` contains `Constraint` or `SignomialConstraint` objects, but the docstring says they can also be `Signomial`. That seems confusing. And I still don't know what it means to turn a Signomial into a Posynomial constraint, which the inline comment says we're doing...

bqpd: Where do the inline comments assume that? It does assume that they are meant to be constraints, but not that they are Constraint objects.

The "turn into a Posynomial constraint" is the whole add-single-negative-c to both sides and then divide by it rigamarole done below.

whoburg: ok yes, what I mean is that they assume they are meant to be constraints.

I think my complaint can be resolved by tweaking the method docstring to clarify that the signomials must be intended to represent constraints. I'll take a stab at that.

bqpd: This is partly clarified by the fact that it's in models.py, but good point.

whoburg:
Yes, hallelujah for working SignomialConstraints!

gpkit/nomials.py
pgkirsch: related to #387

PR #376, opened 2015-08-27
fix nomial equality

### LINKED ISSUE #374: make nomial __ne__ and __eq__ more symmetric
Let's implement nomials `__eq__(self, other)` as 

1. If `self` and `other` can make a constraint, return that constraint
2. Otherwise, return `not self.__ne__(other)`

whoburg:
philosophy:
* every nomial_data.units is either UnitContainer, None, or list thereof. 
* (this means that every Signomial.units is either UnitsContainer or None)
* VarKey.units is a Quantity, string or KeyError
* cs are always numpy arrays or Quantities.

fixes #374

### DIFF
--- a/gpkit/nomials.py
+++ b/gpkit/nomials.py
@@ -77,7 +77,7 @@ class Signomial(NomialData):
                         raise ValueError("cannot add dimensioned and"
                                          " dimensionless monomials together.")
                 else:
-                    units = cs[0]/cs[0].magnitude
+                    units = Quantity(1, cs[0].units)
                     if units.dimensionless:
                         cs = [c * ureg.dimensionless for c in cs]
                         units = ureg.dimensionless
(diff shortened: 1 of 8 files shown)


gpkit/nomials.py
```diff
@@ -49,6 +50,7 @@ def __init__(self, exps=None, cs=1, require_positive=True, simplify=True,
                 exp = ({VarKey(**descr): 1} if exps is None else
                        {VarKey(exps, **descr): 1})
                 descr = list(exp)[0].descr
+                units = list(exp)[0].units
```
whoburg: for 2 jolly rangers why not implement the previous three lines as 

```python
vk = VarKey(**descr) if exps is None else VarKey(exps, **descr)
descr = vk.descr
units = vk.units
exp = {vk: 1}
```
?

gpkit/nomials.py
```diff
@@ -75,13 +77,13 @@ def __init__(self, exps=None, cs=1, require_positive=True, simplify=True,
                         raise ValueError("cannot add dimensioned and"
                                          " dimensionless monomials together.")
                 else:
-                    units = cs[0]/cs[0].magnitude
+                    units = Quantity(1, cs[0].units)
```
whoburg: I like this MUCH better.

whoburg:
This is fantastic!

whoburg:
retest this please

whoburg:
retest this please

Issue #369, opened 2015-08-26
Examples failing


pgkirsch:
The following examples fail with latest code and need to be fixed:
- [x] `beam_bending.py`

`NameError: name 'length' is not defined`

- [x] `simpleflight.py`

`AttributeError: 'Model' object has no attribute 'variables'`

- [x] `water_tank.py` (related to issue #361)

`RuntimeWarning: Primal solution violates constraint: 21.5443469003 is greater than 1.`

pgkirsch:
Add to make examples

whoburg:
@pgkirsch, are you on this?

pgkirsch:
It's on my radar but I'm not currently working on it. Should be easy so hopefully can get it done in the next day or so. Is that OK?

The beam bending example is a mess so I'm intending to redo it.

pgkirsch:
`simpleflight` exists in both `docs/source/examples` and `gpkit/tests`. Can I go ahead and remove the one from `tests` in light of issue #232?

whoburg:
yes, but note those two files differ, so please be sure to check out the diff and make sure everything relevant ends up in the final version.

whoburg:
Fixed by #378

Issue #232, opened 2015-04-27
Read The Docs Examples should be Unit Tests


cjk12:
At some point, everything we have online should be included in the unit tests.  Stems from issue #231

whoburg:
@bqpd, I know we discussed this early on. Is it true that read the docs currently runs all embedded examples? I agree further coordination of tests and docs would be good, especially after dealing with things like b1e0a4324d361b23193d16284998efe2b9509887.

@cjk12, note that in this case linked tests/docs would surprisingly not have solved the problem, as the docs had been updated in the gpkit repository. They just hadn't been re-built by read the docs. Good reminder that I should have pressed the rebuild button after changing them.

cjk12:
Ah, that makes sense.

I'll let you two continue discussing coordination and you guys can close the issue when satisfied.

bqpd:
Building the docs doesn't run embedded code at present, unfortunately.

whoburg:
Would the easiest approach be to create a unit test module that simply runs every example and makes sure nothing is raised?

bqpd:
I've been writing the advanced command examples in the docs as I write unit
tests for them, but one has to remember to keep them in sync.

Alternately, we could move the docs folder inside of the gpkit folder
(letting readthedocs know) and I think directly quote segments of the code
from the docs?

-Ned

bqpd:
I'm liking ^ that idea more and more.

whoburg:
That sounds cool!

bqpd:
So, is it alright with everyone if we move the docs folder inside the gpkit folder?

whoburg:
This is fine by me.

whoburg:
also address the simple flight example

whoburg:
what's the plan for this re: 0.3.2?

bqpd:
Still planned! I was busy moving bio lab equipment all day.

PR #378, opened 2015-08-30
update examples -- addresses issue #369

pgkirsch:
- Replaced the existing not-even-close-to-working beam_bending example with a new beam example that just solves the euler bernoulli beam bending problem for a cantilever beam with constant distributed load.

- Cleaned up and updated `simple flight`

- Changed `model.py`'s sweep message to only mention the number of passes not the number of variables because `geometric_program.py` outputs the number of variables already for every solve, and this was causing an error previously because `model` doesn't have a `variable` attribute.

- Removed duplicate `simpleflight` file from the `tests` directory in anticipation of all docs examples becoming unit tests anyway.


pgkirsch:
Ready for code review @bqpd @whoburg

pgkirsch:
Just kidding about removing the `simpleflight` unit test.....
Put it back in.

pgkirsch:
retest this please

docs/source/examples/beam.py
```diff
@@ -0,0 +1,46 @@
...
+import matplotlib.pyplot as plt
+
+N = 100
+L = 5.
+dx = L/(N-1)
+ei = 1E4
+EI = Var("EI", ei)
+
+P = 100.
+p = Vec(N, "p", descr="Distributed load")
```
whoburg: I think `descr` has changed to `label`.

docs/source/examples/beam.py
```diff
@@ -0,0 +1,46 @@
...
+p = p.sub(p, P*np.ones(N))
```
whoburg: This smells like a units issue -- what are the units of `P`? Does `P` change if I change `N`?

docs/source/examples/beam.py
```diff
@@ -0,0 +1,46 @@
...
+
+V  = Vec(N, "V", label="Internal shear")
+M  = Vec(N, "M", label="Internal moment")
+th = Vec(N, "th", label="Slope")
+w  = Vec(N, "w", label="Displacement")
+
+eps = 1E-16 #something small
+
+substitutions = {var: eps for var in [V[-1], M[-1], th[0], w[0]]}
```
whoburg: It feels odd that this is needed but I see why -- keeping all the vector variables the same length. Let's let this slide for now but keep in mind that we may want to revisit this as we build up more examples that involve discretization.

docs/source/examples/beam.py
```diff
@@ -0,0 +1,46 @@
...
+
+objective = w[-1]
+
+constraints = [V.left[1:N]     >= V[1:N]    + 0.5*dx*(p.left[1:N]    + p[1:N]),
+               M.left[1:N]     >= M[1:N]    + 0.5*dx*(V.left[1:N]    + V[1:N]),
+               th.right[0:N-1] >= th[0:N-1] + 0.5*dx*(M.right[0:N-1] + M[0:N-1])/EI,
+               w.right[0:N-1]  >= w[0:N-1]  + 0.5*dx*(th.right[0:N-1]+ th[0:N-1])
+              ]
```
whoburg: Note this violates python pep8 convention (multiple spaces), but I'm ok with it if @bqpd is too.

docs/source/examples/beam.py
```diff
@@ -0,0 +1,46 @@
...
+
+m = Model(objective, constraints, substitutions)
+
+sol = m.solve(verbosity=1)
+
+x = np.linspace(0, L, N)
```
whoburg: add inline comment stating what `x` is?

docs/source/examples/beam.py
```diff
@@ -0,0 +1,46 @@
...
+w_gp = sol("w")
+w_exact =  P/(24.*ei)* x**2 * (x**2  - 4*L*x + 6*L**2)
+
+assert max(abs(w_gp - w_exact)) <= 1e-4
+
+plt.plot(x, w_gp, 'k', x, w_exact, 'b')
```
whoburg: I'd throw all this in a conditional `if PLOT:` and set `PLOT=True` at the top of the file -- in preparation for turning examples into unit tests.

docs/source/examples/simpleflight.py
```diff
@@ -1,44 +1,43 @@
-import numpy as np
-
+from gpkit.shortcuts import *
 import gpkit
+import numpy as np
```
whoburg: by convention imports should be ordered by standard library stuff first, then site packages. Technically this means it's ok for numpy to come after gpkit, but I think most python developers would expect numpy to come first since it's so "standard".

docs/source/examples/simpleflight.py
```diff
@@ -46,24 +45,24 @@
...
+                Re <= (rho/mu)*V*(S/A)**0.5,
+                C_f >= 0.074/Re**0.2,
+                W <= 0.5*rho*S*C_L*V**2,
+                W <= 0.5*rho*S*C_Lmax*V_min**2,
+                W >= W_0 + W_w]
 
 
 def m():
-    return gpkit.Model(D, equations)
+    return gpkit.Model(D, constraints)
```
whoburg: might as well call this `model` for readability in code that calls it.

gpkit/model.py
```diff
@@ -300,8 +300,7 @@ def _solve(self, programType, solver, verbosity, skipfailures,
                            for (var, grid) in zip(sweep, sweep_grids)}
 
             if verbosity > 0:
-                print("Solving for %i variables over %i passes." %
-                      (len(self.variables), N_passes))
+                print("Solving over %i passes." % N_passes)
```
whoburg: @bqpd please confirm you agree with the message change -- @pgkirsch mentioned in the pull request comments that the # variables message gets repeated by the GP solve.

whoburg:
code review complete. see inline comments.

What ended up happening with trying to have a single simpleflight?

whoburg:
@pgkirsch, to make it easier to debug #379 and also to enable examples-as-unit-tests, could you please turn the beam example into a method that returns the model, as a function of `N`?

docs/source/examples/simpleflight.py
pgkirsch: You are referring to the function name right? Making the change, but I just realised it doesn't get called anywhere. Do we still want to leave it in?

docs/source/examples/beam.py
pgkirsch: Units of `P` are `N/m`. `P` shouldn't change if `N` changes, because `L` is fixed.

docs/source/examples/beam.py
bqpd: Oh, good call.

bqpd: Oh, good call.

bqpd: Oh, good call.

docs/source/examples/beam.py
bqpd: :+1:

gpkit/model.py
bqpd: But only if verbosity is 1, right? I think there's some utility in saying "x variables over y passes" if you're not planning on printing it out for each pass.

Maybe `if verbosity == 1:  print("Solving for %i variables" % len(self.variables))` ?

pgkirsch: But `self.variables` doesn't exist here

docs/source/examples/simpleflight.py
whoburg: I imagine we'll need it for making unit testing call this...

gpkit/model.py
bqpd: erps hrm

docs/source/examples/beam.py
whoburg: Ok. I'd advocate for using units here.

pgkirsch:
@whoburg added the changes including making the beam into a function of `N`. I also made it a function `EI`, `L` and `P`. Is that ok?

whoburg:
Yes I'm fine with them being arguments but let's give them defaults.

pgkirsch:
I decided it might make sense to leave the `tests/simpleflight` unit test in place until issue #232 is addressed, at which point the `docs/....../simpleflight` can become the simpleflight unit test.

pgkirsch:
That's not good! I added the units (and made `dx` into a `gpkit.Variable`) and suddenly my assertion is much weaker. Originally `assert max(abs(w_gp - w_exact)) <= 1e-4` now it can only get `1e-2`. Opening a ticket now...

@whoburg still want me to push with units?

whoburg:
Yes to pushing with units.

Might want to compare relative error instead of absolute.

pgkirsch:
The error is not caused by units, but by making `dx` into a `gpkit.Variable`. Changed the issue accordingly.

pgkirsch:
Last couple code review issues addressed.

Relative error doesn't seem great here only because of the first element of the w vector that is `~1e-16`.

whoburg:
ok gotcha, that makes sense.

I still think maybe `w[0]` shouldn't be an optimization variable, but let's not worry about that for now, this is looking good.

I think I'll leave this open unil we resolve #382, but I intend to merge.

pgkirsch:
Great! Thanks for the review. Interested in discussing ways of not having `w[0]` be an optimization variable.

bqpd:
@pgkirsch, you want to pull the trigger on this one?

pgkirsch:
I would love to

whoburg:
@pgkirsch, want to add the beam example to `docs/source/examples.rst` and to examples in `docs/Makefile`?

pgkirsch:
Done and added to a new pull request (#384) for ease

PR #391, opened 2015-09-04
make discretizations easier with nan constraints

bqpd:


bqpd:
@whoburg, @pgkirsch, would love your eyes on this...

gpkit/nomial_data.py
```diff
@@ -107,7 +107,7 @@ def diff(self, var):
     def __eq__(self, other):
         """Equality test"""
         if not all(hasattr(other, a) for a in ("exps", "cs", "units")):
-            return NotImplemented
+            return False
```
bqpd: This was a nice idea @whoburg but doesn't work: turns out that 0 is both `not equal` and `not not equal` to a Signomial, so passing it control is a bad idea. :-/

pgkirsch:
This looks great. Thanks!

gpkit/nomial_data.py
whoburg: Ahh I see, you mean like this: 

```python
In [8]: with SignomialsEnabled():
   ...:     s0 = Signomial(0)
   ...:

In [9]: s0
Out[9]: gpkit.Signomial(0)

In [10]: 0 == s0
Out[10]: False

In [11]: 0 != s0
Out[11]: False
```
? 

Do you believe they should both be False? I think `0 == Signomial(0)` should be True...

bqpd: Hmm, maybe. But the bigger issue is

```python
In [11]: 0 != Signomial(1)
Out[11]: False
```

whoburg: I don't love this fix, because it prevents me from writing a class that doesn't have exps/cs/units, but knows how to compare itself to a nomial data.

I'm about to propose something else in another PR.

whoburg: (this discussion continued and resolved at #392)

bqpd:
Ack, @whoburg, I'm back to thinking we should (maybe only with .left and .right) allow + nuke `0 >=` constants. The best I can do without that is:

```python
shear_eq = (V >= (V+dx*p).right)
# moment increases from right to left quadratically
moment_eq = (M >= (M + 0.5*dx*(V+V.left)).right)
# theta and displacement decrease from tip to base
theta_eq = (th >= (th + 0.5*dx*(M+M.right)/EI).left)
displ_eq = (w >= w.left + 0.5*dx*(th + th.left))[1:]
```

as opposed to the original

```python
# shear and moment increase from tip to base (left > right)
shear_eq = [V.left >= V + 0.5*dx*(p + p.left)]
moment_eq = [M.left >= M + 0.5*dx*(V + V.left)]
# theta and displacement decrease from tip to base (right > left)
theta_eq = [th.right >= th + 0.5*dx*(M + M.right)/EI]
displ_eq = [w.right >= w + 0.5*dx*(th + th.right)]
```

I think we're going to see this kind of fenceposting a lot in discretizations, and `.left` and `.right` become even more useful if they gain this feature.

(I'm still on board with not allowing `0 >= x` to be created.)

bqpd:
@whoburg: here's a new messy way to do this...

bqpd:
retest this please

bqpd:
This is failing from #392

bqpd:
retest this please

bqpd:
@whoburg this is ready for review

bqpd:
retest this please

bqpd:
retest this please

bqpd:
retest this please

bqpd:
retest this please

bqpd:
Looks like it's failing on MOSEK? but not on my computer...

bqpd:
Ah, got it. Mosek needs the lower unbounded variables to be zeroed...

bqpd:
retest this please

bqpd:
retest this please

bqpd:
retest this please

bqpd:
@whoburg: I do feel that 0 should be a Monomial, though. (because it can be used in Posynomial constraints)

whoburg:
How can 0 be used in Posynomial constraints?

In my opinion, and that of most texts I've seen on GP, 0 is neither a monomial nor a Posynomial, because cs are strictly greater than 0.

Indeed, it is always possible to take the log of a Posynomial/Monomial, but it's not possible to take the log of 0.

bqpd:
We currently allow constraints of the form `x >= 0`

whoburg:
Yeah but don't those get dropped? They get dropped precisely because `x` is guaranteed to be greater than 0 if it's a Posynomial. And we (rightfully) don't allow `Posynomial(0)`.

bqpd:
Right, but if we _do_ have a `Signomial(0)`, should it become a `Monomial`?

whoburg:
I don't think so. I think that's a Signomial.

whoburg:
Otherwise our inheritance structure is broken -- a Posynomial is a Signomial with strictly positive cs; a Monomial is a Posynomial with only one term.

According to that, 0 is neither Posynomial or Monomial.

bqpd:
Hmm, ok.

bqpd:
This could use a review, @whoburg. I think we need some tools to make the creation of Models with variable numbers of stations (discretized models?) easier.

whoburg:
sorry -- I didn't mean to let this drop. reviewing nowish -- assuming this is still a candidate for merge?

Any high-level info / philosophy of this PR that I should understand before reviewing? (there's no description at the top, but I'll refresh myself on the comments thus far)

whoburg:
Rereading our dialogue above, the unresolved issues appear to have been:

1. Should `0 >= posynomial` constraints ever get ignored?
2. Should 0 be a Monomial?

At the risk of initiating another long back-and-forth, what's the current status / thinking on these?

bqpd:
Heh, we could potentially go over this in person, too! It's a series of commits that try to make Models with discretizations easier.

whoburg:
That sounds great -- will PM re meeting

bqpd:
So, there's a bit of a weirdness, but no hack here:

Signomial(0) >= posynomial never gets ignored, but if you cast that by hand to Monomial(0) >= posynomial it'll throw an error unless the Monomial(0) has `"pgt"` in the `fencepost` attribute.

gpkit/nomials.py
```diff
@@ -539,12 +540,18 @@ class Constraint(Posynomial):
     Form is self.left >= self.right.
     """
     def __str__(self):
+        if self.void:
+            return "VOID"
```
whoburg: remove VOIDs?

bqpd: done!

bqpd:
@whoburg: I made a new Beam, just doing trapezoidal integration
(as @pgkirsch's original)
I've convinced myself that the sensitivity "issue" is a reasonable modeling side-effect, not a bug

bqpd:
retest this please

bqpd:
retest this please

bqpd:
(this works on cvxopt on my computer...but maybe not on Jenkins?)

whoburg:
What's the status here?

bqpd:
The new Beam example works on my computer but not on Jenkins, which I need to debug...but it's ready for code review!

whoburg:
Will review once this is changed to implement zeros as discussed in #290 -- let me know when.

bqpd:
This is superseded by #430

Issue #405, opened 2015-09-14
gpkit.interactive broken after v0.3


bqpd:
Widgets and latex printing don't work.

whoburg:
+1, high priority. @bqpd, are you able to work on this soon?

bqpd:
Yes! Do widgets work after your other fix?

whoburg:
They work on a separate example I have, but I haven't tested on any other widget examples.

bqpd:
Fixed by PR #439

whoburg:
:clap:

whoburg:
@bqpd, can you please check out the `LLhale` branch on gpkit-models and try to get the following working? I'm getting a silly SP error for a model that's a GP. Any chance of a quick fix?

```python
from solar_uav import solar_uav
m = solar_uav()
m.interact()
```

Output
```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
/Users/whoburg/anaconda/envs/gpkit/lib/python2.7/site-packages/gpkit/model.pyc in display(**subs)
    701                 self.solve(**solvekwargs)
    702             else:
--> 703                 self.localsolve(**solvekwargs)
    704             fn_of_sol(self.solution)
    705 

/Users/whoburg/anaconda/envs/gpkit/lib/python2.7/site-packages/gpkit/model.pyc in localsolve(self, solver, verbosity, skipsweepfailures, *args, **kwargs)
    358     'Model.localsolve()' can only be called on models with Signomials,
    359     since such models have only local solutions. Models without Signomials have
--> 360     global solutions, and can be solved with 'Model.solve()'.""")
    361             raise
    362 

ValueError: No Signomials remained after substitution.

    'Model.localsolve()' can only be called on models with Signomials,
    since such models have only local solutions. Models without Signomials have
    global solutions, and can be solved with 'Model.solve()'.


```

bqpd:
Fixed in fb8748f

whoburg:
Thanks!

bqpd:
Sorry for the late fix :/

PR #409, opened 2015-09-23
patches to get widgets marginally working

whoburg:
related to #405.

longer-term solution to #405 desperately needed

### DIFF
--- a/gpkit/interactive/widget.py
+++ b/gpkit/interactive/widget.py
@@ -13,9 +13,9 @@ def widget(gp, outputfn=None, ranges=None, **solveargs):
     # and this should not use copy anyway
     # keeping for now -- was intended to avoid widgets mutating the GP
     # that should not be possible and needs to be refactored
-    from copy import deepcopy
     original_cost_units = gp.cost.units
-    gp = deepcopy(gp)
+    # while we're hacking, hack some more -- avoid calling copy
+    gp = gp.__class__(gp.cost, gp.constraints)
     gp.cost.units = original_cost_units
     # end HACK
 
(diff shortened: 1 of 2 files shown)


whoburg:
assuming CI passes and if there are no objections in the next ~5 minutes, I'm going to merge this. Sorry for the unilateral decision w/o code review, but this passes unit tests and I need it to run an old example in a 2pm meeting

gpkit/interactive/widget.py
```diff
@@ -13,9 +13,9 @@ def widget(gp, outputfn=None, ranges=None, **solveargs):
     # and this should not use copy anyway
     # keeping for now -- was intended to avoid widgets mutating the GP
     # that should not be possible and needs to be refactored
-    from copy import deepcopy
     original_cost_units = gp.cost.units
-    gp = deepcopy(gp)
+    # while we're hacking, hack some more -- avoid calling copy
+    gp = gp.__class__(gp.cost, gp.constraints)
```
bqpd: ...excellent...

gpkit/interactive/widget.py
```diff
@@ -30,17 +30,17 @@ def outputfn(gp):
...
-        gp.sub(subs, replace=True)
-        if hasattr(gp, "localsolve"):
-            gp.localsolve(**solveargs)
-        else:
-            gp.solve(**solveargs)
+        gp.substitutions.update(subs)
+        # if hasattr(gp, "localsolve"):
+        #     gp.localsolve(**solveargs)
+        # else:
+        gp.solve(**solveargs)
```
bqpd: Assuming it's a GP?

whoburg: No actually, I think `gp` is a `Model` here (I wanted to minimize changes to the code to get things working, so I retained the variable name `gp`, but this is clearly bad naming). 

`Model` does not have a `sub` method, so I had to do the hack above instead.

bqpd: Oh,  I meant with the removal of the check for `localsolve`, this will not work for SPs. (Not that it would have worked anyway)

whoburg: Ah yes -- you are absolutely correct.

(This is a total hack/patch meant only to hold us over until #405 is resolved).

PR #415, opened 2015-10-01
Xu algorithm

pgkirsch:
Lots of things still need to be done before this can be considered a "serious" pull request, but wanted to put this out there as a way of getting feedback on general approach:

- [ ] Uncomment the warnings in `signomial_program.__init__()`
- [ ] Uncomment the `ValueError` in `solution_array.parse_result()`
- [ ] Fix unit tests so they pass (currently 2 failures and 1 error due to the expected `assertEqual` vs. operator overloading problem)
- [ ] Make sure I can replicate the results from the Xu paper using the examples in ` tests/Xu_examples`
- [ ] Make sure I can pass the 2 simple tests in `t_xu.py`
- [ ] Improve initial guess method (will likely address many of the issues with Xu_examples)
- [ ] Need better choice of (/method for choosing) `M`
- [ ] Decide what to do about the `w` update scheme (if anything)
- [ ] Figure out what is causing the sensitivity errors 
- [ ] Make some kind of sheltered dummy variable to replace the `t` dummy variable and the `s_i` slack variables (could use something like @bqpd's \fbox{} default)

### DIFF
--- a/gpkit/model.py
+++ b/gpkit/model.py
@@ -203,7 +203,7 @@ class Model(object):
         RuntimeWarning if an error occurs in solving or parsing the solution.
         """
         try:
-            return self._solve("gp", solver, verbosity, skipfailures, algorithm=None,
+            return self._solve("gp", solver, verbosity, skipfailures,
                                *args, **kwargs)
         except ValueError as err:
             if err.message == ("GeometricPrograms cannot contain Signomials"):
@@ -256,7 +256,7 @@ class Model(object):
     global solutions, and can be solved with 'Model.solve()'.""")
             raise
 
-    def _solve(self, programType, solver, verbosity, skipfailures, algorithm,
+    def _solve(self, programType, solver, verbosity, skipfailures, algorithm=None,
                *args, **kwargs):
         """Generates a program and solves it, sweeping as appropriate.
 
--- a/gpkit/model.py
+++ b/gpkit/model.py
@@ -173,7 +173,7 @@ class Model(object):
 
     # TODO: add get_item
 
-    def solve(self, solver=None, verbosity=2, skipfailures=True,
+    def solve(self, solver=None, verbosity=2, skipfailures=True, algorithm=None,
               *args, **kwargs):
         """Forms a GeometricProgram and attempts to solve it.
 
@@ -202,7 +202,7 @@ class Model(object):
         RuntimeWarning if an error occurs in solving or parsing the solution.
         """
         try:
-            return self._solve("gp", solver, verbosity, skipfailures,
+            return self._solve("gp", solver, verbosity, skipfailures, algorithm=None,
                                *args, **kwargs)
         except ValueError as err:
             if err.message == ("GeometricPrograms cannot contain Signomials"):
--- a/gpkit/model.py
+++ b/gpkit/model.py
@@ -213,7 +213,7 @@ class Model(object):
     have only local solutions, and are solved with 'Model.localsolve()'.""")
             raise
 
-    def localsolve(self, solver=None, verbosity=2, skipfailures=True,
+    def localsolve(self, solver=None, verbosity=2, skipfailures=True, algorithm=None,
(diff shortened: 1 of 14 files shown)


pgkirsch:
As requested by @whoburg, a motivating minimum working example for why I believe signomial equality constraints are necessary:

```
# minimize
x + 1/y
# subject to
x + y = 1
```

whoburg:
I'm not convinced by that example. I believe the optimum of that SP is at `x = 0` (or `log(x) = -\infty` in log space).

Here's one way to see that: the equality constraint can be everywhere eliminated by replacing `y` with `1 - x`:

```
# minimize
x + 1/(1-x)
```
And this is equivalent to:
```
# minimize
x + 1/t
# subject to
x + t <= 1
```
This is a GP, which is dual infeasible because `x` has no lower bound.

Indeed, plotting f(x) = x + 1/(1-x) shows that the optimal value of the cost is equal to 1, but this value is never attained for any x > 0.

pgkirsch:
Missed an important constraint:

```
# minimize
x + 1/y
# subject to
x + y = 1
x >= 0.1
```

whoburg:
In that case I'm still not fully convinced (by that example) -- I claim that a simple modeling reformulation converts the problem to a GP with the same optimal cost:

```python
x = Variable('x')
t = Variable('t')
​
m = Model(x + 1/t, [x + t <= 1, x >= 0.1])
m.solve()
Using solver 'cvxopt'
Solving for 2 variables.
Solving took 0.00769 seconds.

Cost
----
 1.211

Free Variables
--------------
t : 0.9 
x : 0.1 
```

The optimal value of `y` is equal to `t`.

pgkirsch:
Yeah just realized that was yet another bad example. I'm trying to abstract away from the Hyperloop model and not doing a very good job of it. One sec...

pgkirsch:
As discussed in person, finding a MWE that absolutely needed an equality constraint proved more difficult than I had anticipated ("one sec" turned into 2 days). I believe I may have finally found a case that requires an equality constraint:

```
from gpkit.shortcuts import *
from gpkit import SignomialsEnabled

b = Var('b') #  bypass cross sectional area 
a = Var('a') #pod cross sectional area 
x = Var('x') # cruise velocity
y = Var('y')

with SignomialsEnabled():
    m = Model(x**2*a + x + 1/x,
             [
              x >= 1,
              y * b >= 1.,
              y >= 1 + x**2, 
              y <= 1 + x**2,
              1 >= a + b, 
              1 <= a + b
              ]
             )

    sol = m.localsolve()
```

This model has two equality constraints, but in practice, I believe only the `1 = a + b` constraint must be an equality to give a feasible solution, in which case the `y = 1 + x**2` constraint can be relaxed to `y <= 1 + x**2`.

Given how difficult it has been to identify cases when a strict equality is absolutely necessary, I think this PR becomes much lower priority, and might even get closed without being merged.

bqpd:
This is super interesting. I wonder if we should do a speed comparison between the Xu method and using a first-solve to determine which side of an equality is needed?

pgkirsch:
Another case where equality constraints are necessary:

In the landing gear model (https://github.mit.edu/convex/GPjet/commit/cea39625cdee37ed398c4d6870b58cb46e51fdb5), I need `xcd_n + x_n == x_CG_f` because it is a geometric relationship and must be satisfied strictly.

Both `xcd_n + x_n <= x_CG_f` and `xcd_n + x_n >= x_CG_f` work, but neither gives a solution that is satisfied the constraint with equality.

I am sure there is some objective that would result in the "correct" solution but I don't think that is a satisfactory solution, because (a) if I only want to minimize weight, I shouldn't need to add another variable to the objective, and (b) finding the "right" objective sounds like a completely intractable problem.

bqpd:
Nifty! It's great having these examples.

bqpd:
This branch is superseded by newer Xu branches.

PR #419, opened 2015-10-13
simple_wing becomes simpleflight, resolves #410

### LINKED ISSUE #410: Simple wing code missing from docs
As seen here: http://gpkit.readthedocs.org/en/latest/examples.html#simple-wing, the simple wing example shows output but not the code itself

whoburg:


whoburg:
Only failing ElCap, so I'm merging.

PR #424, opened 2015-10-15
mkvar

pgkirsch:
I got really bored of writing 

```python
var1 = Variable("var1")
var2 = Variable("var2")
etc.
```

so I threw together a quick function `mkvar.py` that automatically generate the variable initialization from a list. Not sure if this belongs in GPkit at all, but I thought I'd put it here in case anyone else wants it.

### DIFF
--- a/mkvar.py
+++ b/mkvar.py
@@ -1,0 +1,37 @@
+"""
+Takes a .txt file of variable name strings and rewrites
+them as a .py file of GPkit variable initializations
+
+$ ls
+example.txt
+$ python mkvar.py example.txt
+$ ls
+example.py example.txt
+
+- - - - - - - example.txt - - - - - - -
+foo
+bar
+- - - - - - - - - - - - - - - - - - - - 
+
+                | |
+                | |
+               \   /
+                \ /
+                 v
+
+- - - - - - - example.py - - - - - - -
+foo = Variable('foo')
+bar = Variable('bar')
+- - - - - - - - - - - - - - - - - - - - 
+"""
+import sys
+
+def initialise_variable_list(fname):
+    with open(fname) as infile:
+        with open(fname.replace('.txt','.py'), 'w') as outfile:
+            for line in infile:
+                var = line.rstrip('\n')
+                outfile.write(var + " = Variable(\'" + var + "\')\n")
+
+if __name__ == "__main__":
+    initialise_variable_list(sys.argv[1])


mkvar.py
```diff
@@ -0,0 +1,37 @@
...
+- - - - - - - example.txt - - - - - - -
+foo
+bar
+- - - - - - - - - - - - - - - - - - - - 
+
+                | |
+                | |
+               \   /
+                \ /
+                 v
```
whoburg: SICK ascii art

bqpd:
Interesting. I'm going to need to make a text-format for GPs for the web server. Alternately, this could be a macro for vim / sublime / etc.

Alternately alternately, there's the old "modifying the globals() dict" technique for doing this live...

whoburg:
What if the standard format were a csv, with the columns being the same as the current argument list to `Variable`?

This would let people keep track of their variables in Excel (and even google spreadsheets...), which some might find handy.

bqpd:
Inteeeresting. Works fine for me; @pgkirsch?

whoburg:
Would need to figure out VectorVariables... a separate csv might be simplest, though I do a lot of changing of variables from vector to non-vector...

bqpd:
Hmm, there's a couple ways I would do that; first option would be to have the columns be:

`Variable name, Index (if vector), Lower bound, Upper bound`

then for the constraints have the var name be CONSTRAINT, the index be either GP, SP, or OBJECTIVE,  the third column be the cs list and the fourth column be the exps list

mkvar.py
pgkirsch: My ASCII-art-to-actual-code ratio is 0.56, i.e. far too low. I knew I should have made that arrow bigger....

whoburg:
Wait you would be trying to put Variable definitions and the text representation of a nomial into the same file / format? It seems to me those might want to be separate

pgkirsch:
As long as I can still initialize my variable list in a regular text editor that's fine with me. As far as I can tell it would just add a comma to the end of my current input file format right?

Whatever the solution, I still think that a modeler ultimately needs to be able to see the initialization of the variable in the same file as where the constraints are written, to be able to see inline comments/the `label`.

bqpd:
You wouldn't even need the comma :D

bqpd:
Proposal for CSV format: (`#` comments would not be present in the actual file

```
Variables  # optional. format is: name, shape, unitstr, label
foo  # semicolons are banned? or need to be escaped with a backslash?
bar, , m, length of bar
vector1, 3  # vector named vector1 of length 3
matrix, "3,3"  # 2d array named matrix of shape (3,3)

# optional arbitrary whitespace
Minimize  # required for a non-'1' objective
foo

Constraints  # optional; if blank, can be used to create a model with only variables, so as to also have mkvar's functionality.
foo >= 1 + bar  # spaces required
vector1;1 + vector1;2 + vector1;3 <= 1

Substitutions  # optional. format is variable name
bar, 3
```

whoburg:
Do we really want users to be writing their constraints, objective, and substitutions in csv files? (I thought the intent was to just support the `Variables` section).

And, even if we support the other sections, should they maybe go in separate files, so that standard csv parsing utilities like `pandas. read_csv()` can just work automatically, instead of needing to split into sections?

bqpd:
Hmm, good point.

whoburg:
Is the plan to leave this open, not merge it, but keep it as an example for the eventual csv reading functionality?

bqpd:
If that's ok with @pgkirsch

pgkirsch:
This PR can be closed because there is a more elegant vim solution. The command `:%s/\(\w\+\)/& Variable(‘\1’)/g` takes a list of variables

```
foo
bar
giraffe
```

and converts it into

```
foo = Variable(‘foo’)
bar = Variable(‘bar’)
giraffe = Variable(‘giraffe’)
```

Issue #438, opened 2015-11-01
Solution arrays can't print 2d variables


bqpd:
Including swept vectors.

``` python

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-20-6fa46dc91949> in <module>()
----> 1 sol.table()

/home/eburn/projects/src/gpkit/gpkit/solution_array.pyc in table(self, tables, fixedcols)
    118             strs += [results_table(self["freevariables"],
    119                                    "Free Variables",
--> 120                                    fixedcols=fixedcols)]
    121         if "constants" in tables and self["constants"]:
    122             strs += [results_table(self["constants"],

/home/eburn/projects/src/gpkit/gpkit/solution_array.pyc in results_table(data, title, minval, printunits, fixedcols, varfmt, valfmt, vecfmt)
    171         units = unitstr(var, into=" [%s] ", dimless="") if printunits else ""
    172         if isvector:
--> 173             vals = [vecfmt % v for v in val[:4]]
    174             ellipsis = " ..." if len(val) > 4 else ""
    175             valstr = "[ %s%s ] " % ("  ".join(vals), ellipsis)

TypeError: float argument required, not numpy.ndarray
```

bqpd:
(@whoburg ping)

whoburg:
hmm, what should they print? One fix is to replace `val[:4]` with something like `val.flatten()[:4]` (haven't tested this, but I suspect it would work)

bqpd:
How about we push that as a stopgap, but also think about the "right" way to do this?

whoburg:
sounds great. 

Re: the "right" way,  numpy and pandas have already thought about multidimensional printing -- might want to just use an OTS solution.

bqpd:
Hmm, yeah.

PR #439, opened 2015-11-01
Fixes and changes to interactive

### LINKED ISSUE #422: Fuel example is out of date
Didn't get updated after Model/GP switch

bqpd:
references #438, closes #422 and #405


bqpd:
retest this please

bqpd:
@whoburg this is ready for review (and for use, I think)

gpkit/interactive/ractor.py
```diff
@@ -10,23 +10,27 @@
...
             print sol.table(showtables)
-    return widget(gp, ractivefn, ranges)
+    return m.interact(ractivefn, ranges)
+
 
 new_jswidget_id = itertools.count().next
 
-def ractorjs(title, gp, update_py, ranges, constraint_js=""):
+
+def ractorjs(title, m, update_py, ranges, constraint_js=""):
```
whoburg: the signature of this method is so similar to `ractorpy` -- could we make `title` a keyword arg so that they actually do have the same args?

gpkit/interactive/ractor.py
```diff
@@ -73,10 +77,8 @@ def ractorjs(title, gp, update_py, ranges, constraint_js=""):
 
     evalarray = [""]*np.prod(lengths)
 
-    gp.sweep = {}
-    gp.prewidget = gp.last
```
whoburg: yes!! down goes `prewidget`!

gpkit/interactive/ractor.py
bqpd: Iiiinteresting. I'll think on that.

gpkit/model.py
```diff
@@ -176,7 +176,7 @@ def zero_lower_unbounded_variables(self):
                      if bound == "lower"}
             self.substitutions.update(zeros)
 
-    def solve(self, solver=None, verbosity=1, skipfailures=True,
+    def solve(self, solver=None, verbosity=1, skipfailures=False,
```
whoburg: I think this is minor, and I think I agree with this change, but can we just discuss what happened here with the `skipfailures` change so I understand? (any change to `solve` gets my attention)

bqpd: Ha. It shouldn't have been True by default, since it has a nice error message explaining what it does that shows up when it's false and a failure occurs.

whoburg: Excellent, that makes sense.

The documentation for the `skipfailures` argument is currently 
```python
        skipfailures : bool (optional)
            If True, when a solve errors during a sweep, skip it.
```
Should this be updated to include interactives instead of just sweeps?

gpkit/model.py
```diff
@@ -343,7 +343,7 @@ def solve_pass(i):
             self.program = []
             for program, result in mapfn(solve_pass, range(N_passes)):
                 self.program.append(program)  # NOTE: SIDE EFFECTS
-                if not hasattr(result, "status"):  # solve succeeded
+                if result:  # solve succeeded
```
whoburg: nice

gpkit/model.py
```diff
@@ -544,6 +544,30 @@ def _latex(self, unused=None):
                                  for var, val in self.constants.items()]) +
                          ["\\end{array}"])
 
+    def interact(self,  fn_of_sol=None, ranges=None, **solveargs):
```
whoburg: I'm ok with the name; would `widget` be another option?

This needs a docstring now that it's part of `Model`, in my opinion.

bqpd: Indeed it does. I like the verbish nature of `m.interact()`

gpkit/model.py
bqpd: Certainly reads better :)

gpkit/model.py
```diff
@@ -544,6 +544,30 @@ def _latex(self, unused=None):
...
+        try:
+            from ipywidgets import interactive, FloatSlider
+        except ImportError:
+            from IPython.html.widgets import interactive, FloatSliderWidget
+            FloatSlider = FloatSliderWidget
+
+        if ranges is None:
+            ranges = {k._cmpstr: FloatSlider(min=v/10.0, max=10*v,
+                                             step=v/10.0, value=v)
```
whoburg: Minor thought -- I've noticed that `10.0` is too aggressive in most of my applications. I wonder if we should try `2.0`?

gpkit/model.py
bqpd: Interactives don't (necessarily) sweep, just ractorjs.

gpkit/model.py
```diff
@@ -544,6 +544,30 @@ def _latex(self, unused=None):
...
+                      for k, v in self.constants.items()}
+        if fn_of_sol is None:
```
whoburg: like `fn_of_sol` as a name

gpkit/model.py
bqpd: Hmm! I like that better than my initial thought of "half an order of magnitude ~ pi"

gpkit/model.py
```diff
@@ -544,6 +544,30 @@ def _latex(self, unused=None):
...
+            def fn_of_sol(solution):
+                print solution.table(["cost", "freevariables"])
```
whoburg: print sensitivities too, since we've been noticing they're often even more useful than the solution itself?

bqpd: Yeah. I just don't want it to get too long...maybe count the variables and choose that way? (prioritize free over senss or the other way around?)

gpkit/model.py
bqpd: haha it's so non-pythonic but good

gpkit/model.py
whoburg: Ah I see. Wonder if it should be called `skipsweepfailures`. But that's super minor, low priority, and unrelated to this PR. 

Thanks for explaining.

gpkit/model.py
whoburg: sounds good.

gpkit/model.py
whoburg: I'd say between 10 and 20? Probably 20, since we're not sweeping?

gpkit/model.py
bqpd: Good point, and I like that name.

gpkit/model.py
whoburg: I don't think most users would want sensitivities and no free variables.

With the modular models we could print only the highest-level layer, right? So that a too-long table would be an indication to the user that they might want to break up their model into pieces? This thought started as "maybe the user should tag the variables they want to see printed" -- but it seems like Modular models give us a clean way to do that already.

gpkit/model.py
bqpd: Made it a step size of v/16 so there'll be 8 to the left and 16 to the right.

whoburg: oh right it's not in log space... sounds good

gpkit/model.py
bqpd: Or, you could write your own display fn? (or except arguments to the `table` method?)

whoburg:
This looks great!! I'm happy with it -- once docstring request in the inlines is complete, I'm for merging.

whoburg:
@pgkirsch, this is super relevant for the models you're currently pulling together (will make interactives work)

whoburg:
does this work for SP models?

bqpd:
Erp, lemme add that.

gpkit/model.py
bqpd: yeah :-/

bqpd:
Has SP support now.

whoburg:
did you push that? I don't see it on origin

bqpd:
Getting the last couple changes.

whoburg:
cool. if it works for SPs and `interact()` has a good docstring, I'd say merge when you're happy.

gpkit/interactive/ractor.py
bqpd: Title can't be optional, but I could add it to `ractorpy`

bqpd:
Pushed! Take a quick look?

whoburg:
looking

Issue #446, opened 2015-11-04
use classmethods to create multiple ways of constructing objects


whoburg:
This is a proposal; comments welcome.

I propose that we follow the design guideline of _not_ writing `__init__` methods that accept multiple calling sequences. Instead, I think we should definitely allow multiple ways to construct an object, but do so using classmethods. Like this:

``` python
    @classmethod
    def frommodel(cls, m):
        return cls(m.cost, m.constraints)
```

This is pretty standard python design style, e.g. there's `dict.fromkeys`, and `date.fromordinal`, and `date.fromtimestamp`, etc.

bqpd:
Oh man. This is such the right way to do this.

bqpd:
relevant: http://pythoncentral.io/difference-between-staticmethod-and-classmethod-in-python/
https://stackoverflow.com/questions/12179271/python-classmethod-and-staticmethod-for-beginner

PR #461, opened 2015-11-12
implement fromnomials for NomialData construction

whoburg:
This is a first stab at the type of changes needed under #446. There is a little awkwardness due to the fact that GeometricProgram implements a different `__init__` signature than `NomialData` (which it inherits from), but this PR gets around that issue.

### DIFF
--- a/gpkit/nomial_data.py
+++ b/gpkit/nomial_data.py
@@ -56,7 +56,8 @@ class NomialData(object):
         """Way to initialize from nomials. Calls __init__.
         Used by subclass __init__ methods.
         """
-        exps, cs = exps_and_cs_from_nomials(nomials)
+        exps = functools_reduce(add, (tuple(s.exps) for s in nomials))
+        cs = np.hstack((mag(s.cs) for s in nomials))
         # nomials are already simplified, so simplify=False
         NomialData.__init__(self, exps, cs, simplify=False)
         self.nomials = nomials  # TODO eliminate constructor-dependent state
@@ -134,13 +135,6 @@ class NomialData(object):
         if self.units != other.units:
             return False
         return True
-
-
-def exps_and_cs_from_nomials(nomials):
-    "Flatten an iterable of nomials into exps and cs"
-    exps = functools_reduce(add, (tuple(s.exps) for s in nomials))
-    cs = np.hstack((mag(s.cs) for s in nomials))
-    return exps, cs
 
 
 def simplify_exps_and_cs(exps, cs, return_map=False):
--- a/gpkit/nomial_data.py
+++ b/gpkit/nomial_data.py
@@ -19,18 +19,11 @@ class NomialData(object):
     varlocs: {VarKey: list} (terms each variable appears in)
     units: pint.UnitsContainer
     """
-    def __init__(self, exps=None, cs=None, nomials=None, simplify=True):
-        if nomials and (exps or cs):
-            raise ValueError("The NomialData initializor accepts either"
-                             " exps and cs, or nomials, but not both.")
-        elif nomials:
-            self.nomials = nomials
-            exps = functools_reduce(add, (tuple(s.exps) for s in nomials))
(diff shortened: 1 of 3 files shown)


whoburg:
this is ready for review.

bqpd:
(is it expected to pass tests?)

gpkit/nomial_data.py
```diff
@@ -54,6 +45,24 @@ def __hash__(self):
             self._hashvalue = hash(tuple(zip(self.exps, self.cs)))
         return self._hashvalue
 
+    @classmethod
+    def fromnomials(cls, nomials):
+        """Construct a NomialData from an iterable of Signomial objects"""
+        nd = cls()  # use pass-through version of __init__
```
bqpd: Interesting! Is this pass-through standard?

whoburg:
yep, expected to pass tests -- just failing El Capitan

gpkit/nomial_data.py
whoburg: possibly, but I just made it up -- easiest solution I could see to not duplicate code.

whoburg: (note the pass through is implemented in `__init__` -- it's not a python standard)

bqpd: Yeah. Any reason to use the pass through rather than just modifying the object in `fromnomials`?

whoburg: modifying which object in `fromnomials`? I need to create one, which is the purpose of calling `cls`. We also need `init_from_monomials` so that children (GeometricProgram, in this case) can call `super.init_from_monomials`. (I don't see an obvious way for subclassers to call the `fromnomials` classmethod).

Does that answer what you were asking?

bqpd: Oh, interesting. Yeah, for subclasses that does seem necessary. Hmm.

bqpd: What about a NomialData class and a NomialsData class, which inherits, for GP? (or similar names)

whoburg: hmm, what I like about this is that it's causing discussion about the right class design / inheritance structure.

It's both impressive and a bit awkward that Signomial and GeometricProgram can both inherit from NomialData. I think it's the right design -- it encodes cs and exps as the right building block. The awkwardness comes in the fact that GP needs a `map` too (i.e. `p_idxs`). And that's related to the fact that it makes sense to create a GP from nomials, but not so much a Signomial from nomials.

bqpd: Not entirely, it's also used for other collections of signomials.

bqpd: (relatedly, how do you feel about Constraints not inheriting from nomials in the other thread?)

whoburg: might be worth a focused in-person discussion on the overall inheritance structure?

bqpd: Great, see you tomorrow :p

whoburg: I'm definitely open to the idea that constraints have nomials as attributes, but that they are not themselves nomials.

whoburg:
Retest this please

galbramc:
retest this please

whoburg:
Shall we merge this?

bqpd:
I think it'll be superseded by NomialCollections, right?

galbramc:
How did these tests pass without the MOSEK fixes in the other pull request?

bqpd:
Good question.

whoburg:
@galbramc, good question. Any guesses?

@bqpd, I think we should merge this -- In my opinion it's better than what's on master right now, and NomialCollections might as well build on this.

galbramc:
My only guess is that .gpkit/expopt.so was lingering somehow... but it should have been deleted at the end of the testing... I moved the delete command to the top of the script instead.

bqpd:
That folder's also deleted as part of the build process

galbramc:
Strange... I don't know why it passed then...

Issue #454, opened 2015-11-10
Setting variable value directly is not equivalent to setting value using '=='


pgkirsch:
I believe the following two code snippets should work the same, but they don't. The first solves just fine. The second returns unknown.

``` python
from gpkit import Variable, Model, SignomialsEnabled, units
import numpy as np

Qv = Variable('Qv', 'N*m', 'Torsion moment imparted by vertical tail')
Lvmax = Variable('Lvmax', 35000, 'N', 'Tail max lift')
bv = Variable('bv', 7, 'm', 'Vertical tail span')
plamv = Variable('plamv', 2.3, '-', 'Tail taper ratio + 2')
qlamv = Variable('qlamv', 1.3, '-', 'Tail taper ratio + 1')
taucone = Variable('taucone', 15000/0.000145,'N/m^2', 'Shear stress in cone')
dRfuse = Variable('dRfuse', 15*0.0254, 'm', 'Offset of lower fuselage cylinder')
Afuse = Variable('Afuse', 'm^2', 'Cross sectional area')
tcone = Variable('tcone', 'm', 'Cone thickness')
Rfuse = Variable('Rfuse', 77*0.0254, 'm', 'Fuselage radius')
plamcone = Variable('plamcone', 1.3, '-', 'Cone taper ratio + 1')
lcone = Variable('lcone', 15*0.3048, 'm', 'Cone length') # added this 
Wcone = Variable('Wcone', 'N', 'Weight of cone')
rhocone = Variable('rhocone', 2700, 'kg/m^3', 'Density of cone material') # [TAS]
g = Variable('g', 9.81, 'm/s^2', 'Gravitational acceleration')

objective = Wcone
constraints = [
                Afuse >= np.pi*Rfuse**2 + 2*Rfuse*dRfuse,
                Qv == 2*Afuse*taucone*tcone,
                Wcone >= rhocone*g*2*Lvmax*bv*plamv/(3*qlamv)*lcone/(taucone*Rfuse*plamcone),
              ]
m = Model(objective, constraints)
m.solve()
```

The only difference is how the value of Rfuse is specified.

``` python
from gpkit import Variable, Model, SignomialsEnabled, units
import numpy as np

Qv = Variable('Qv', 'N*m', 'Torsion moment imparted by vertical tail')
Lvmax = Variable('Lvmax', 35000, 'N', 'Tail max lift')
bv = Variable('bv', 7, 'm', 'Vertical tail span')
[...]

Issue #315, opened 2015-08-10
install process: reduce mucking with user's machine and enable parallel CI


whoburg:
Thanks to @galbramc for input on best practices

 - [x] We should ask for permission to modify the user's `.bashrc` file, (if a modification is necessary* (for CI purposes, the necessary items will already exist in `.bashrc`, so the permission warning won't occur). (Edit: we no longer modify .bashrc or .bash_profile)

 - [x] Add a flag to build.py or setup.py to specify an alternate path/name for `.gpkit`. (Since `.gpkit` needs to be deleted for cvxopt-only testing on linux machines, we are currently in a serial-testing mode for CI. This item, along with #312, will enable parallel testing.

bqpd:
Sounds good! Although `.gpkit` really shouldn't need to be deleted, I agree it should be moveable.

galbramc:
We only have to delete .gpkit so that we can test the cvxopt-only situation. Right now, using mosek actually works on Linux so long as .gpkit/expopt.so exists. Hence I have to delete it to test if the testing assumes mosek is installed when a user only has cvxopt.

bqpd:
Okay. So having gpkit only attempt to use mosek if mosek was found during the install process would solve this issue?

galbramc:
So right now gpkit always finds mosek in the test maching. That is why I have to modify gpkit/env/settings. Similarly, since gpkit finds mosek, it also builds .gpkit/expopt.so. So I have to delete .gpkit/ to prevent a faulty test from using it. Because .gpkit is a "global" thing, a separate test that actually needs to use .gpkit/expopt.so cannot run at the same time as the cvxopt only test that deletes .gpkit/. As a result, the testing is serialized. gpkit is not fundamentally serialized, just the testing.

whoburg:
related to #354

whoburg:
I think this is now resolved thanks to the new `GPKIBUILD` environment var. Feel free to reopen if that is false.

PR #465, opened 2015-11-14
minor fixes

whoburg:
@bqpd, this is ready to review pending Jenkins. Addresses comments in #454

### DIFF
--- a/gpkit/_mosek/cli_expopt.py
+++ b/gpkit/_mosek/cli_expopt.py
@@ -83,7 +83,7 @@ def imize_fn(path=None, clearfiles=True):
         """
 
         with open(filename, "w") as f:
-            numcon = 1+p_idxs[-1]
+            numcon = p_idxs[-1]
             numter, numvar = map(int, A.shape)
             for n in [numcon, numvar, numter]:
                 f.write("%d\n" % n)
--- a/gpkit/model.py
+++ b/gpkit/model.py
@@ -252,6 +252,7 @@ class Model(object):
     def signomials_et_al(self):
         "Get signomials, beforesubs, allsubs in one pass; applies VarKey subs."
         signomials = self.signomials
+        # don't use self.beforesubs here to avoid re-computing self.signomials
         beforesubs = NomialData.fromnomials(signomials)
         allsubs = beforesubs.values
         allsubs.update(self.substitutions)


gpkit/_mosek/cli_expopt.py
```diff
@@ -83,7 +83,7 @@ def imize(c, A, p_idxs, *args, **kwargs):
         """
 
         with open(filename, "w") as f:
-            numcon = 1+p_idxs[-1]
+            numcon = p_idxs[-1]
```
bqpd: Ah, great.

gpkit/model.py
```diff
@@ -252,7 +252,7 @@ def allsubs(self):
     def signomials_et_al(self):
         "Get signomials, beforesubs, allsubs in one pass; applies VarKey subs."
         signomials = self.signomials
-        beforesubs = NomialData.fromnomials(signomials)
+        beforesubs = self.beforesubs
```
bqpd: Err, no, this is slower.

whoburg: really? how? aren't they equivalent?

bqpd: Nope. 

```python
    @property
    def beforesubs(self):
        "Get this Model's NomialData before any substitutuions"
        return NomialData.fromnomials(self.signomials)
```

whoburg: ah, ok. I'll update the PR  -- standby.

whoburg:
ready for re-review

bqpd:
Looks good!

whoburg:
Shoot, I think I broke master. @galbramc the CI on master reports "aborted"

galbramc:
no I am changing things on jenkins. Because of the 'globa' $HOME/.gpkit folder you cannot have multiple tests running at the same time on any given machine. I realized that I need to make sure the gpkit_Push and gpkit_PullRequests could not be running at the same time. I found a plugin that was even better. It allows me to restrict things so no two jobs can run at the same time on a given machine, but jobs can run in parallel across machines. Before your builds were running in serial across all machines, but now you are parallel across machines and serial on a given machine.

whoburg:
awesome!!

bqpd:
Nifty! (looking forward to removing $HOME/.gpkit)

galbramc:
Yes. If you remove $HOME/.gpkit so that each build is completely independent, then we can run all builds completely in parallel. Should you open an issue maybe about this?

whoburg:
I believe that is captured in #315 .

galbramc:
Yes it is.

galbramc:
Actually, I think you can check off that first goal. You are no longer mucking with a users .bashrc or .bash_profile, so no need to ask for permission.

PR #478, opened 2015-11-19
update modular example, fix nested merges, closes #425

### LINKED ISSUE #425: Update Modular example
The modular example is currently out of date -- `docs/source/ipynb/modular/modular.ipynb` still inherits from GP instead of Model. Is this the best example we have of what will ultimately become Models playing together?

bqpd:
This is far too slow during model construction. I noticed this on Tony's examples too, that vector-based construction is faster than object-oriented. This should be fixed! But for the meantime, this works.


bqpd:
@whoburg, ready for review. Note that if you click on 'view' for the notebook file, it'll display it properly.

whoburg:
Reviewed.

 - see inlines
 - check 5ee7db0613c66996aed43ea8148aa78fe20ff67b (feel free to squash commits if you like)
 - I'd rename `Flight` to `FlightCondition`.

Seeing the modularity is super cool!!!

bqpd:
I feel like Flight is a good name; it takes off, takes on fuel, etc. for each case.

bqpd:
@whoburg, responded to code review

whoburg:
Maybe `FlightSegment` then? `AirplaneFlight1` seems like an odd name to me.

Changes look good! I'm ready for you to merge once we decided on that name -- and `Flight` is fine if you insist.

bqpd:
I don't care very much (and don't know the naming conditions in this field); which do you prefer?

whoburg:
Oops! I misunderstood what `Flight` was doing (though I still think the name could be better). Perhaps `Performance`, or `FlightPerformance`?

Here are some relevant names with my interpretation of what they are usually used to mean:
`FlightCondition`: a condition at a single instant in time, like a 4g pullup, or crosswind takeoff, or steady level flight
`FlightSegment`: a portion of a mission over which some quantities change, and others are integrated -- like a cruise segment, or a climb segment, etc
`Mission`: a group of FlightSegments, and sometimes FlightConditions as well
`Performance`: overall name given to the "discipline" associated with analyzing range/payload/weights/fuel burn over a mission

whoburg:
@cjk12 your input welcome on my previous comment

bqpd:
@cjk12 proposes Airplane => Mission, Flight => Segment (or to FlightSegment)

bqpd:
Talking @cjk12 is proposing we adopt an explicit aircraft design taxonomy, a la 
image

bqpd:
also, Spar => WingWeight. Or WeightWing, taking into account that there are various weight models; Fuel => WeightFuel, as well. Really these should be integrated into a Weight module (convenient naming scheme, eh)

bqpd:
@whoburg, given the above, what'd you propose as a renaming?

bqpd:
@whoburg, added a couple commits to this to close v0.3.4; once they're reviewed and Modular naming is finalized, we'll be good to go with that version.

setup.py
```diff
@@ -46,17 +46,18 @@
...
-    url='https://www.github.com/convexopt/gpkit',
-    install_requires=['numpy', 'scipy'],
-    version='0.3.3.0',
-    packages=['gpkit', 'gpkit._mosek', 'gpkit.tests', 'gpkit.interactive'],
-    package_data={'gpkit': ['env/*'],
-                  'gpkit._mosek': ['lib/*']},
+    name="gpkit",
+    description="Package for defining and manipulating geometric "
+                "programming models.",
+    author="MIT Department of Aeronautics and Astronautics",
```
whoburg: should this be kept consistent with "citing GPkit"?

whoburg:
Added a couple commits and one comment -- ready for review and merging. In case the diff is hard to view, the renamings (for now) in 9a3c1d9c030c9286646db7963edff296c579ede9 are:

Fuel --> FuelBurn
Flight --> FlightSegment
Spar --> WingBox

setup.py
bqpd: Hmm, probably! We should meet up and talk about that change at some point.

bqpd:
Looks good! Merging.

setup.py
whoburg: word.

PR #486, opened 2015-11-25
latex table option

pgkirsch:
This is the code I modified to create latex tables directly from solution arrays. Any feedback is welcome.


whoburg:
superseded by #489

PR #490, opened 2015-11-28
Define units from GPkit folder

bqpd:
See https://github.com/aeroa/gpkit-models/pull/24

### DIFF
--- a/gpkit/__init__.py
+++ b/gpkit/__init__.py
@@ -20,6 +20,10 @@ __version__ = "0.3.4"
 UNIT_REGISTRY = None
 SIGNOMIALS_ENABLED = False
 
+from os import sep as os_sep
+from os.path import dirname as os_path_dirname
+UNITDEF_PATH = os_sep.join([os_path_dirname(__file__), "gpkit_units.txt"])
+
 
 def enable_units():
     """Enables units support in a particular instance of GPkit.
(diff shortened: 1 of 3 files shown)


whoburg:
Now why is Jenkins unhappy... ?

whoburg:
retest this please

whoburg:
retest this please, just to be sure. @galbramc, that was a strange testing failure -- temporary loss of internet connection?

bqpd:
Worth noting that if we define currency by the year, we can easily define exchange rates...

whoburg:
I'm ready to merge this, @bqpd please go ahead if you are too.

bqpd:
:+1:

bqpd:
@whoburg, See also the gpkit-models mod; should that be USD2012?

whoburg:
Hmm good point -- would you basically hard-code the inflation rates by defining USD2012, USD2013, etc relative to some baseline, like USD1913?

I guess the other alternative is to expect users to continually update their code to use current labor rates, etc?

bqpd:
Yerp.

galbramc:
Looks like a loss of internet for some odd reason. No idea what happened there.

whoburg:
I'm very open to adding the year-based dollar units, but of course we should also keep `USD` for models that don't want to worry about inflation.

bqpd:
see https://github.com/aeroa/gpkit-models/pull/25

bqpd:
see https://github.com/convexopt/gpkit/pull/494

PR #500, opened 2016-01-13
Modular mike

mjburton11:
Modular with canard configuration


mjburton11:
@whoburg @bqpd I have a few questions about this model: 

The error says that the variables that I'm substituting don't have the same units but I can't find any that don't have the same.  Can you help me figure out what is going on here? 

Also, I'm not sure how to incorporate the spar model.  Will it still apply for a canard configuration?  Will it also apply if I want to use carbon fiber instead of aluminum? 

In my stability submodel I have one constraint equation that is signomial and I'm not sure how to best handle it without just assigning a number to the variable of interest.  Any suggestions? 

Any comments on how to improve this model would be very helpful.  Thanks!

bqpd:
`CanardStability` doesn't solve with MOSEK, btw.

bqpd:
You've got V for both voltages (in VTTricopter) and for flight velocity (everywhere else), is the problem. When you `&` models together it tries to collapse all variables with the same name.

bqpd:
@whoburg, design discussion: we could fix this by only combining things with the same name and units, by requiring users to explicitly whitelist linked variables, or by rewriting the error message to make it clearer what the problem is. Which do you prefer from a user process standpoint?

mjburton11:
Just met with @whoburg and this modular is working.  Weight sizing for VTOL/horizontal flight vehicle.

bqpd:
How did you get around the variable naming?

mjburton11:
I just made a new file, ```canard_VTOL.ipynb```, and added in new submodels one by one making sure they worked as I added them.

bqpd:
Err, are Voltage and Velocity still both named `V`?

bqpd:
Ah, looks like there's no volts in those models yet.

mjburton11:
No they aren't.  In the old file I did change ```V``` to ```Volt``` for the voltage, and yes in the new model I haven't added volts.

bqpd:
Great! Sorry to harp on that, just wanted to make sure, since that was the original bug. :)

mjburton11:
No worries! Thanks for you help.

whoburg:
good debugging.

@bqpd, regarding your options, repeated here for clarity:

"
 we could fix this by
1. only linking variables with the same name and units, 
2. by requiring users to explicitly whitelist linked variables, or
3. by rewriting the error message to make it clearer which variables can't be linked.
Which do you prefer from a user process standpoint?
"

I support 3 for now, with the possibility of implementing 2 in the future. I don't like 1 because it feels too magical and I worry about both false negatives (e.g. where the user has thrust T in newtons and mass T in kg and expects them to be linked, but their units error is not identified because the linking silently does not happen) and about false positives (where unintentional linking occurs [catching this, of course, would require the whitelist]).

bqpd:
3 has been implemented in #503.

whoburg:
I propose closing this PR without merging (and if we want the models they can be submitted to gpkit-models as .py files instead of .ipynb files).

Thoughts?

mjburton11:
Ya that makes sense to me.  I don't think this should live in ~modular.

PR #502, opened 2016-01-21
allow tweaking modelname printing in latex

bqpd:
```python
import gpkit
x = gpkit.Variable("x", model="test")
x  # prints without modelname
gpkit.settings["latex_modelname"] = True
x  # prints with modelname
```

### DIFF
--- a/gpkit/__init__.py
+++ b/gpkit/__init__.py
@@ -156,3 +156,4 @@ def load_settings(path=SETTINGS_PATH):
         settings_ = {"installed_solvers": [""]}
     return settings_
 settings = load_settings()
+settings["latex_modelname"] = False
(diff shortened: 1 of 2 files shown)


bqpd:
@whoburg, ready for review.

gpkit/__init__.py
```diff
@@ -156,3 +156,4 @@ def load_settings(path=SETTINGS_PATH):
         settings_ = {"installed_solvers": [""]}
     return settings_
 settings = load_settings()
+settings["latex_modelname"] = False
```
whoburg: should we make the default `settings["latex_modelname"] = False`, so that this doesn't affect the status quo? I'm willing to make the settings change for my applications where I don't want printing, and I like the idea of leaving printing on to remind us that a longer-term fix is still needed.

whoburg:
@bqpd: thanks for doing this. See one suggestion on flipping the default; otherwise I'm ready to merge.

bqpd:
I think you're right; changed the default back.

whoburg:
this looks great. thanks for the hotfix.

PR #342, opened 2015-08-14
Experimenting with CI testing

galbramc:
I created this pull request so I can mess with getting the CI setup without using other pull requests

PR #504, opened 2016-01-24
reduce duplicate code

whoburg:
@bqpd, ready for review.

### DIFF
--- a/gpkit/substitution.py
+++ b/gpkit/substitution.py
@@ -37,7 +37,6 @@ def get_constants(nomial, substitutions):
 
 def vectorsub(subs, var, sub, varset):
     "Vectorized substitution"
-
     if hasattr(var, "__len__"):
         isvector = True
     elif hasattr(var, "descr"):
@@ -179,8 +178,7 @@ def substitution(nomial, substitutions, val=None):
                 sub = VarKey(name=sub, **descr)
                 exps_[i] += HashVector({sub: x})
                 varlocs_[sub].append(i)
-            elif isinstance(sub, VarKey):
-                sub = VarKey(sub)
+            elif isinstance(sub, (VarKey, Monomial)):
                 if isinstance(var.units, Quantity):
                     try:
                         new_units = var.units/sub.units
@@ -191,23 +189,17 @@ def substitution(nomial, substitutions, val=None):
                                          " those of the original '%s' [%s]." %
                                          (sub, sub.units.units,
                                           var, var.units.units))
-                exps_[i] += HashVector({sub: x})
-                varlocs_[sub].append(i)
-            elif isinstance(sub, Monomial):
-                if isinstance(var.units, Quantity):
-                    try:
-                        new_units = var.units/sub.units
-                        cs_[i] *= new_units.to('dimensionless')
-                    except DimensionalityError:
-                        raise ValueError("units of the substituted monomial"
-                                         " '%s' [%s] are not compatible with"
-                                         " those of the original '%s' [%s]." %
-                                         (sub, sub.units.units,
-                                          var, var.units.units))
-                exps_[i] += x*sub.exp
-                cs_[i] *= mag(sub.c)**x
-                for subvar in sub.exp:
-                    varlocs_[subvar].append(i)
+                if isinstance(sub, VarKey):
+                    sub = VarKey(sub)
+                    exps_[i] += HashVector({sub: x})
+                    varlocs_[sub].append(i)
+                elif isinstance(sub, Monomial):
+                    exps_[i] += x*sub.exp
+                    cs_[i] *= mag(sub.c)**x
+                    for subvar in sub.exp:
+                        varlocs_[subvar].append(i)
+                else:
+                    assert False  # should not be possible to get here
             else:
                 raise TypeError("could not substitute with value"
                                 " of type '%s'" % type(sub))


gpkit/substitution.py
```diff
@@ -179,8 +178,7 @@ def substitution(nomial, substitutions, val=None):
                 sub = VarKey(name=sub, **descr)
                 exps_[i] += HashVector({sub: x})
                 varlocs_[sub].append(i)
```
whoburg: @bqpd: when is the `elif isinstance(sub, Strings):` case invoked for substitution? Reason I ask is: should the Strings case do the same unit checking as the VarKey and Monomial cases?

bqpd: Naw, you don't know what units a string has? Anyway this is all replaced by keydicts.

whoburg: haha - no I don't. What's an example of a string substitution?

bqpd: ```m.sub("x", "y")```

whoburg: ok I think I get it -- string substitution is like a renaming operation? Create something just like `x` but with `.descr["name"]` overwritten by `"y"`?

bqpd: from line 137 of `substitution.py`
```python
elif isinstance(sub, Strings):
                descr = dict(var.descr)
                del descr["name"]
                sub = VarKey(name=sub, **descr)
                exps_[i] += HashVector({sub: x})
                varlocs_[sub].append(i)
```

whoburg:
retest this please

whoburg:
@galbramc: upon retesting this, it is now suddenly failing pylint because of an issue that I just fixed in #529, commit f151c5e5cc23b3f4713871dc802e6067fd333b3a.

That's strange behavior -- I fix an issue in one PR and it breaks another one, presumably because these are somehow seen as 'new' errors.

Any easy solution?

whoburg:
@bqpd, aside from the minor pylint annoyance, can we consider merging this?

galbramc:
if you merge this with pylint errors, the gpkit push will fail.

retest this please

galbramc:
Looks like there are still 3 pylint warnings, but they should be easy to fix

whoburg:
ugh, but those are unrelated to this PR -- Jenkins is only making them an issue because of activity on #529 and #523. I feel we should not need to fix them on this PR.

galbramc:
So the one that came up in ractor.py is likely because I installed ipython, which meant the system changed, and hence a warning about not being able to import got changed to a warning about import order. So that's why that one warning would show up there.

I'm not sure why there are new warnings in tests/t_sub.py. Installing ipython should not have changed the warnings in that file...

I can reset the counters if you don't want to fix these warnings.

galbramc:
Wait, I see it. There were some successful pull requests by #529 and #523. Thats why you see new errors in tests/t_sub.py. You don't have to fix those warnings in that case. The gpkit push job is not aware that those were fixed at any point, so merging this PR will not cause a failure. 

Installing ipython did cause a failure now in gpkit push, I'll reset the counter as this is a result of changing the system. 

I'll also reset the counter based on this pull request as it's a system change that is causing the failure.

galbramc:
retest this please

whoburg:
thanks @galbramc!

@bqpd, please review and merge if happy

galbramc:
ok the counts are reset, so this PR can be merged without problems.

You are going to have this issue where pylint will fail due to activity from multiple different PR until you get the pylint warnings down to zero. I agree it will be a small headache until you get there, but I think it's a worthy goal to strive for.

I still think you have many artificial warnings. Such as limits on the number of methods in a class, and not allowing variables that are only a single character. Spending a couple more minutes on the .pylintrc will likely remove several hundred warnings that I don't think are improving the code.

whoburg:
is there any chance the pylint thresholds could be set to only change (tighten) when PRs are actually merged to master? It seems the problem is that the thresholds decrease based on code that's still in PRs -- that implies, for example, that if I create a new PR that deletes all the code and run it through Jenkins, Jenkins will end up with a cap of 0 pylint warnings -- I claim that cap should only be changed when pylint warnings *on master* decrease.

galbramc:
I agree. The reference really always should be the master branch. Unfortunately, I don't have that option in jenkins to compute the reference based on another branch or another jenkins job. I can only use the previous builds in the current job as a reference.

Another solution would be to get the master branch warnings down to zero. Then all pylint warnings would be legitimate. I know this is too big of a job to do at once, but it should be something to strive for.

Also, tell your students not to create PR's where they have deleted all the code :P

galbramc:
One thing that we can do use use #342 to reset the counts at any time. Jenkins first merges the PR into the master branch, and then runs all the tests. Since #342 does not have any new code in it, the counts will be reset to what ever is in the master branch. I can show you how to reset the counts, it's very easy.

galbramc:
test this please

galbramc:
test this please

gpkit/substitution.py
```diff
@@ -179,8 +178,7 @@ def substitution(nomial, substitutions, val=None):
...
-            elif isinstance(sub, VarKey):
-                sub = VarKey(sub)
+            elif isinstance(sub, (VarKey, Monomial)):
                 if isinstance(var.units, Quantity):
                     try:
                         new_units = var.units/sub.units
```
bqpd: Comparing to the refactor, I think this might be the inverse of what it should be...

gpkit/substitution.py
```diff
@@ -191,23 +189,17 @@ def substitution(nomial, substitutions, val=None):
...
-            elif isinstance(sub, Monomial):
-                if isinstance(var.units, Quantity):
-                    try:
-                        new_units = var.units/sub.units
-                        cs_[i] *= new_units.to('dimensionless')
-                    except DimensionalityError:
-                        raise ValueError("units of the substituted monomial"
-                                         " '%s' [%s] are not compatible with"
-                                         " those of the original '%s' [%s]." %
-                                         (sub, sub.units.units,
```
bqpd: note that for monomials this will print things like

```
units of the substituted monomial x [m] [m] are not compatible with the original y [volt]
```

Because by default monomials print with units and varkeys do not. 

To fix this, replace this line with `(sub.str_without(["units"]), sub.units.units,`

bqpd:
@whoburg, raised some concerns

gpkit/substitution.py
whoburg: hmm, `new_units` isn't a great variable name, since it's dimensionless.

in the line below this one, `cs_[i]` gets multiplied by `new_units.to('dimensionless')`.

I think it's correct as written here -- this causes a `var` in `cm` to have it's `cs` multiplied by 0.01 when it is substituted by a `sub` in meters:

```python
In [9]: Q(1, "cm")/Q(1, "m")
Out[9]: <Quantity(1.0, 'centimeter / meter')>

In [10]: _.to("dimensionless")
Out[10]: <Quantity(0.01, 'dimensionless')>
```

gpkit/substitution.py
whoburg: oh cool, good call - want to just make that change in the refactor now so it's captured?

whoburg:
test this please

gpkit/substitution.py
whoburg: in any case, that line didn't change in this PR

whoburg:
Anything that needs to be done before merging this?

gpkit/substitution.py
bqpd: Right, but if the var is `x [cm]` and the sub is `1 [m]`, then the value of new units should be 100, not 0,01...

whoburg: really? Doesn't a sub cause the units to change to what was substituted? If so, shouldn't the new cs (in [m]) be 0.01 times the old cs (which was in [cm])?

bqpd: The units of the nomial don't change, however.

whoburg: argh, I'm struggling with this.

in your example above, if `a = 1 [m]`, and `y` is in [cm], say 5 cm, then `100 y` is 500... `[m]`? that can't be correct.

whoburg: though GPkit does currently give `100y` for `x.sub(y)` for the example above, so perhaps your point is that it should really be `a = 0.01y [m]`? (which I would agree with)

If that's the case, how on earth are our unit tests passing?

bqpd: Yeah, I mostly gave up and flipped the switch so that it worked properly
for a test I wrote for the refactor...

whoburg: I pushed some work in progress on this to the `subflip` branch (branched off this one). It's failing a model composition test, and haven't been able to track down why. @bqpd, want to take a look?

whoburg: It's shocking that we've had this wrong for so long -- we should add some more unit tests to make sure it remains right. This is when I agree with pylint on too many branches being a bad thing -- we need to test every branch.

galbramc: When was the last time you looked at the coverage information of your unit
tests?

bqpd: Actually we test this but the test was written to my wrong assumptions (as whoburg's above) which I didn't realize until I made model composition use varkey substitution in the refactor.

whoburg: I  believe that the currently-failing test (see `subflip` branch) is written correctly though...

bqpd: You mean `test_model_composition_units` ?

bqpd: In model composition I changed `test_scalar_units` as you have to pass `test_model_composition_units`...

whoburg: Yes, I mean `test_model_composition_units`. I already fixed `test_scalar_units` on the `sublfip` branch.

bqpd: Let's go over this at our next meeting.

bqpd:
Incorporated into refactor.

PR #505, opened 2016-02-02
keep current slider value when max jumps below min during typing

bqpd:
@whoburg

### DIFF
--- a/gpkit/interactive/widgets.py
+++ b/gpkit/interactive/widgets.py
@@ -151,7 +151,7 @@ def create_settings(box):
                 slider.max = new_value
             if attr is "max" and new_value <= slider.value:
                 if slider.max >= slider.min:
-                    slider.value = slider.max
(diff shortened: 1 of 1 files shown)


whoburg:
This is a fix for a bug where changing slider bounds in `controlpanel` caused values to change unintentionally, or worse, to go to zero and cause signomial errors

gpkit/interactive/widgets.py
```diff
@@ -150,7 +150,11 @@ def link_fn(name, new_value):
             elif new_value >= slider.max:
                 slider.max = new_value
             if attr is "max" and new_value <= slider.value:
-                slider.value = new_value
+                if slider.max >= slider.min:
+                    slider.value = slider.max
```
whoburg: should be `new_value`

whoburg:
thanks @bqpd

whoburg:
this works great now

PR #514, opened 2016-02-10
pylint changes as suggested by @galbramc

whoburg:
Minor changes -- makes it easier for users to run pylint without specifying rcfile and to run pylint on `docs/examples/*.py`.

Ready for review.


galbramc:
retest this please

galbramc:
we added the examples directory to the pylint checks, since there are warnings in there the number of warnings increased and hence the failure

galbramc:
@whoburg I could reset the counts or you could fix some of the warnings in the examples directory... your call

galbramc:
retest this please

galbramc:
retest this please

galbramc:
retest this please

galbramc:
The examples give an extra 100 warnings. You should take a look at them and try to adjust the setting of pylint to what you like.

whoburg:
I really like the setting where the number of warnings must decrease or remain constant. I can't go deal with the warnings due to adding `examples/` now though. Could you please increase the thresholds since we've increased the number of files subject to pylint testing?

Also -- do you know if it is possible to locally change settings? For example, I would like to make the invalid-name warning less strict (or remove it entirely) in examples/, but not for the rest of gpkit.

galbramc:
I changed the threshold for you, but your last commit seems to have created real errors.

I think you can just put a different pylintrc file in examples with the settings you want there. Worst comes to worse I'll run pylint twice with the two different rc files.

whoburg:
Well that's what I get for not running the unit tests. Fixed those errors; now getting `[WARNINGS] Plug-in Result: Failed - <a href="warningsResult/new">718 new warnings</a> exceed the threshold of 0 by 718`

galbramc:
I'm guessing I just need to update the counts. Let me look at it.

Can you make a separate rc file and put it in examples?

galbramc:
retest this please

whoburg:
I'm not ready to make the separate examples rc file. We currently have two almost identical rcfiles (gpkit and gpfit). adding one to the examples dir for each of those repos would result in 4 total rcfiles, which makes it a pain to change pylint settings globally.

whoburg:
shall I go ahead and merge this?

galbramc:
Depends on what you want to do with the examples directory. I can turn off something just from the script with the command line call if you tell me what to turn off.

galbramc:
retest this please

whoburg:
We need to carefully think though our pylint settings across all projects but haven't had time to do that yet. I'm against turning off from the script because I want our settings consolidated until we figure out exactly what we want.

Options are
1. Leave this open as a reminder that we need to nail down our pylint settings
2. Merge this and continue with the policy that pylint errors must decrease over time (and we may update our examples settings in the future to make that happen).

I'm in favor of 2. but welcome other opinions

galbramc:
I want option 2, I just want to turn off the things you want me to turn off for examples:

I disabled "invalid-name"

Remaining errors are:

1. import-error
2. no-init
3. too-many-locals
4. redefined-outer-name
5. missing-docstring
6. no-member
7. no-self-use
8. too-few-public-methods
9. undefined-variable
10. superfluous-parens

Look at the end here: https://acdl.mit.edu/csi/job/gpkit_PullRequest/buildnode=reynolds,optimizer=mosek/681/console

galbramc:
Once you tell me what to dissable, I need to run the test one last time to set the counts, then we have option 2 in effect where the counts always have to stay the same or decrease

galbramc:
If you don't want to suppress anything else now, you can go ahead and merge this branch.

galbramc:
retest this please

galbramc:
retest this please

galbramc:
retest this please

whoburg:
Below is your list with my current opinion on whether it should be suppressed in `examples/` only. I'm not fully satisfied with suppressing these in Jenkins settings though, because since we don't have them captured in a pylintrc, our developers will still see the pylint errors in their lint-checkers and when running pylint locally.

I know the separate rc file could work, but I wish we didn't have to maintain two nearly-identical files.

'''''''''''''''
I disabled "invalid-name" (wh: sounds good)

Remaining errors are:

1. unused-wildcard-import (wh: keep, these should be resolved)
2. no-init (wh: keep, these should be resolved)
3. too-many-locals (wh: suppress for examples)
4. redefined-outer-name (wh: keep, these should be resolved)
5. missing-docstring (wh: keep, these should be resolved)
6. no-member (wh: keep for now, these may be possible to resolve)
7. no-self-use (wh: keep for now, these may be possible to resolve)
8. too-few-public-methods (wh: keep for now, these may be possible to resolve)
9. undefined-variable (wh: keep, these should be resolved)
10. superfluous-parens (wh: keep, these should be resolved)

galbramc:
I agree with the dilemma. I would imagine that the examples files will not be touched as frequently as the main code. I think this is a reasonable compromise. I'll let you know when it's ready to merge.

galbramc:
retest this please

whoburg:
I think I might know how to handle everything gracefully using rcfiles. Trying it now. Would mean that we don't need to disable any warnings in Jenkins config.

galbramc:
Oh that would be great. Let me know how I should change the script

galbramc:
retest this please

galbramc:
so it looks like you removed some things in the "ignore" setting of the rc file, and we got 32,566 new warnings... that's actually pretty impressive :P

whoburg:
Hmm I tried and failed. I really thought that a minimalistic pylintrc file in `examples` that linked back to the main rcfile via something like

```
# Specify a configuration file.
rcfile=../../../pylintrc
disable=invalid-name
```
... would do the trick. But my entry in `rcfile=` doesn't seem to have any affect.

galbramc:
Are you executing pyling inside the examples directory?

whoburg:
yes

galbramc:
yeah it's not working for me either....

galbramc:
I did find out that you can dissable pylint warnings from within the python file, e.g.

# pylint: disable=wildcard-import, method-hidden
# pylint: enable=too-many-lines

though I would not recommend using this for the examples, but might be good for that function you can't document

whoburg:
This closed issue over at pylint seems to suggest using a PYLINTRC environment variable for this purpose:
https://github.com/PyCQA/pylint/issues/175

Do you understand what that's suggesting and how we would implement it?

galbramc:
The environment variable will only work if you have just one pylintrc file. It won't help in this situation. You want two rc files.

galbramc:
So the only think I can see is to disable things on the command line, or you can put comments in the example files to suppress the pylint warnints. Since these are examples, I would lean toward the command line option, unless you find something else.

whoburg:
asked for help at PyCQA/pylint#175

galbramc:
did you restore the old pylintrc file?

galbramc:
if you restore the old pylintrc file we can go ahead and merge this branch and just use the jenkins script arguments for now. again, people wont be working in the examples directory often

galbramc:
New idea. How about we just make a bash script that runs pylint for both directories with the separate options for the examples and put that in the repo? That way you just run ./pylint.sh to run pylint on both directors.

galbramc:
Here is a suggested script: pylint.sh

```
#!/bin/bash

#This means you can run the script anywhere in the repo, but not outside
WORKSPACE=$(git rev-parse --show-toplevel)

#Calling pylint directly will not work correcly in a virtualenv if pylint is not installed in the venv
#Using python with the pylint script will always work properly in a virtualenv
PYLINT=`which pylint`

python $PYLINT --rcfile=$WORKSPACE/pylintrc $@ $WORKSPACE/gpkit/

python $PYLINT --rcfile=$WORKSPACE/pylintrc --disable=invalid-name,too-many-locals $@ $WORKSPACE/docs/source/examples/*.py
```

Using a script like this also means you can rename and move the pylintrc file where ever you want to have it. You can also put this script anywhere you want in the repository, because we explicitly say what rc file to use.

I will need you to add:

msg-template={path}:{line}: [{msg_id}({symbol}), {obj}] {msg}

to the pylintrc file as well (after you revert back to the old one).

whoburg:
@galbramc: pushed the pylint script and updated pylintrc. I think we're ready for you to reset the threshold.

galbramc:
test this please

galbramc:
I don't see a pylintrc file anymore...

galbramc:
test this please

galbramc:
test this please

galbramc:
test this please

galbramc:
test this please

galbramc:
ok, merge!

galbramc:
All other pull requests will fail until this is merged, because the pull request script assumes pylint.sh exists in the repo.

galbramc:
test this please

galbramc:
I had a thought. You get a bunch of errors importing cvxopt when I test with only mosek, and vice versa. I'm going try to suppress that

galbramc:
test this please

galbramc:
test this please

galbramc:
test this please

galbramc:
Ok, now the mosek testing ignores _cvxopt.py, and the cvxopt testing ignores _mosek. This way you wont get warnings where pylint can't import cvxop because we are testing mosek by it self, and vice versa.

Go ahead and merge. Again, all other pull reqeusts will fail until this is merged.

whoburg:
Thanks @galbramc, and apologies for the delay!

whoburg:
Shoot, master CI is failing now.

galbramc:
Let me take a look

galbramc:
Looks like to mosek license expired

whoburg:
sent pm

galbramc:
all back in the green now

Issue #507, opened 2016-02-02
weight breakdown visualization


whoburg:
Automatically generate these as output

<img width="523" alt="screen shot 2016-02-02 at 5 29 03 pm" src="https://cloud.githubusercontent.com/assets/6042395/12766554/844577cc-c9d2-11e5-888f-8cd185368749.png">
<img width="463" alt="screen shot 2016-02-02 at 5 29 11 pm" src="https://cloud.githubusercontent.com/assets/6042395/12766555/844e0248-c9d2-11e5-98f0-3c70008abd9f.png">

whoburg:
@mayork

whoburg:
@mayork if you have initial thoughts on how to design this before getting started on coding, post them here - I'm sure you'll get lots of feedback and ideas.

mayork:
I think taking inputs as a series of nested tuples would be easiest...so if we had three weights only, a total and w1 and w2 one level down the input would look like this ('w',['w1','w2'])...my only question is do @whoburg what weights do you want to take as given? I was thinking have a numerical value input for the lowest level weights makes the most sense but am not sure what you want

bqpd:
What about nested (potentially ordered) dicts? `{"w": {"w1": 1, "w2": 3}}

whoburg:
Oh I like the nested dicts idea. Here's a more complicated one:

`{"W": {"W_pay": 100*units("lbf"), "W_fuel": None, "W_wing": {"W_spar": None, "W_skin": None, "W_sys": 100*units("N")}}}`

This allows non-numeric base cases (via None), which is going to be important, e.g. if W_fuel is bounded by another model.

A nice thing is that the logic for deciding what to do with each (key, value) pair in each dict is simple:

```python
if isinstance(value, dict):
    # recurse
else:
    # pass value to `Variable` constructor as the `value` arg.
```
From my limited testing `Variable` deals just fine with values that are numeric, values that have units, and values that are None.

bqpd:
Yup, and the strings allow convenient access to a solution KeyDict.

mayork:
I'm having a little trouble creating constraints after I create all the Var objects...currently I'm trying to put the Vars into a nested list and recursively loop over the list...anyone have a better idea?@bqpd

whoburg:
Might be good to push what you have so far to a branch so we can take a look at the code. If you need help with the git syntax to do so let me know, happy to help.

whoburg:
Design discussion:
[...]

PR #522, opened 2016-02-24
Initial attempt at weight Breakdown (issue #507)

mayork:
@whoburg initial breakdown class ready for review. Should be fully functional. Let me know if you want to tweak how the diagram looks, I could possibly make it colored, change fonts/sizes, dash some lines, etc.

current problems:
improper indent warning from pylint on continued lines


whoburg:
test this please

whoburg:
@mayork, I'm going to rebase and force push to eliminate all the commits related to a merge gone wrong. You'll need to update your local branch -- I'll post again with instructions on how to do that

whoburg:
ok, rebased and force pushed (and I have a backup of the branch in case there is any issue)

anyone (including @mayork ) that has a local checkout of this branch needs to do the following:

(stash any uncommitted changes)
```
git checkout master
git branch -D breakdown
git checkout breakdown
git pull origin breakdown
```

whoburg:
@mayork, it looks like your text editor is still set up to insert tabs instead of 4 spaces -- google how to set the `tabstop` and `expandtab` so that tab means 4 spaces for python files.

gpkit/breakdown.py
```diff
@@ -0,0 +1,92 @@
...
+    def __init__(self, input_dict):
+        """
+        model to make graphical breakdowns of weights, etc.
+        takes input as a dict, variable arguments provided in a list
+        """
+        #create the list of variables to make constraints out of
+        self.constr = []
+        self.varlist = []
+        #call recursive function to create gp constraints
+        total = self.recurse(input_dict)
```
whoburg: `total` is never used -- should it be?

gpkit/breakdown.py
```diff
@@ -0,0 +1,92 @@
...
+
+        self.sol = self.solve_method()
+
+    def getSolution(self):
+        return self.sol
+
+    def solve_method(self):
+        m = Model(self.make_objective(), self.make_constraints())
+        return m.solve(verbosity=0)
```
whoburg: I don't think this method is needed. no need to solve the model on initialization -- let a user cal `solve()` instead.

gpkit/breakdown.py
```diff
@@ -0,0 +1,92 @@
...
+        #create the list of variables to make constraints out of
+        self.constr = []
+        self.varlist = []
+        #call recursive function to create gp constraints
+        total = self.recurse(input_dict)
+
+        self.sol = self.solve_method()
+
+    def getSolution(self):
+        return self.sol
```
whoburg: no need for this either -- see below

gpkit/breakdown.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def solve_method(self):
+        m = Model(self.make_objective(), self.make_constraints())
+        return m.solve(verbosity=0)
+
+    def recurse(self, input_dict):
```
whoburg: need docstring

gpkit/breakdown.py
```diff
@@ -0,0 +1,92 @@
...
+    def make_constraints(self):
+        i = 0
+        constraints = []
+        while i < len(self.constr):
+            j = 0
+            posy = 0
+            while j < len(self.constr[i][1]):
+                posy = posy+self.constr[i][1][j]
+                j = j+1
+            constraints.append(1 >= posy/self.constr[i][0])
```
whoburg: might be clearer to do `constraints.append(self.constr[i][0] >= posy)` ?

gpkit/tests/t_model.py
```diff
@@ -354,6 +355,23 @@ def test_signomials_not_allowed_in_objective(self):
             with self.assertRaises(TypeError):
                 sol = m.localsolve(verbosity=0)
 
+    def test_breakdown(self):
```
whoburg: this should not be inside the `TestSP` class -- make a separate `TestBreakdown` class, separate from `TestGP` and `TestSP`

whoburg:
@mayork I did an initial code review with inline comments. You can add additional commits to address them. I think there are some ways to further simplify this - but you can start with the comments so far.

gpkit/breakdown.py
mayork: that makes sense as well...so we don't need to have everything in "standard" format (I think in one of Boyd's lectures he said standard format was to have a one on one side)?

gpkit/breakdown.py
mayork: @whoburg this is the default variable description?

mayork:
disregard the large chunk of deleted lines in the above commits...I was getting a merge error when I tried to commit, I think this was an artifact of moving the branch around. It's fixed now.

mayork:
@whoburg addressed all of the corrections and am up and running with pylint

mayork:
@whoburg talked with @bqpd about the graphics...he thought SVG was the way to go. If you concur I'll start on that.

mayork:
@whoburg 
image

this is my current diagram...thoughts?

bqpd:
test this please

bqpd:
retest this please

bqpd:
verify this please (is that how I do this)

bqpd:
Looks like there's merge conflicts?

mayork:
@bqpd not sure how to fix a merge conflict...also, what I just committed doesn't pass pylint because of improper indent on a continued line. What is the protocol for that? I couldn't figure it our from the config file

whoburg:
retest this please

whoburg:
@mayork can you update the very first comment to give an overview of the current status? What's there now is out of date.

whoburg:
ok failing unit tests because `svgwrite` is not on the Jenkins machines. @bqpd any strategy thoughts here? Should part of this be moved to interactive?

bqpd:
err, is `breakdown.py` the only changed code?

bqpd:
I think this should inherit from `ConstraintSet` and have a method which imports the svg drawing logic from interactive, like `Model` and `ControlPanel`.

gpkit/breakdown.py
whoburg: nope, that conversion is handled internally by gpkit

whoburg:
retest this please

gpkit/breakdown.py
```diff
@@ -188,16 +188,16 @@ def test(self):
...
+        npt.assert_almost_equal(self.sol('w').magnitude, 15, decimal=5)
+        npt.assert_almost_equal(self.sol('w1').magnitude, 11, decimal=5)
+        npt.assert_almost_equal(self.sol('w2').magnitude, 3, decimal=5)
+        npt.assert_almost_equal(self.sol('w3').magnitude, 1, decimal=5)
+        npt.assert_almost_equal(self.sol('w11').magnitude, 3, decimal=5)
+        npt.assert_almost_equal(self.sol('w12').magnitude, 8, decimal=5)
+        npt.assert_almost_equal(self.sol('w121').magnitude, 2, decimal=5)
+        npt.assert_almost_equal(self.sol('w122').magnitude, 6, decimal=5)
+        npt.assert_almost_equal(self.sol('w21').magnitude, 1, decimal=5)
+        npt.assert_almost_equal(self.sol('w22').magnitude, 2, decimal=5)
```
bqpd: `mag(sol(...))` will work even if units are disabled.

mayork:
@bqpd @whoburg I still am not sure what exactly needs to be moved to svg.py inside of interactive....do I need to move every method that touches any part of the drawing object (i.e. dwgrecurese and drawsegment)? I have tried a few things and can't something which works

when I try and do this:

from .interactive.svg import importsvgwrite
        from svgwrite import cm

i get this error
"ValueError: Attempted relative import in non-package"

I just pushed with the svg.py file I made inside interactive and an updated breakdown.py...the svg.py has some commented out areas which reflect a previous attempt to get things working

whoburg:
Try `from gpkit.interactive.svg import` ...

I don't fully understand why the relative import isn't working here - @bqpd might know.

In any case, try to follow the pattern currently used by `gpkit/constraints/costed`'s `controlpanel` method, which imports `modelcontrolpanel` from interactive.

bqpd:
I think the problem is that this isn't being imported as part of the module, and so the relative import won't work.

mayork:
@bqpd could we meet to look at this? I tried a few other things...if I try to do from gpkit.interactive.svg import _____ i get the error:

importerror: no module named svg

bqpd:
I think for now you should just import svgwrite in `breakdown.py`; when the breakdown class is feature-complete, I can sit down with you and we can figure out how to integrate it into gpkit.

mayork:
@bqpd I think that's the point we're at, woody was hoping to merge it into master and close this branch out this week/next week

bqpd:
Okay! For this to be in GPkit, Breakdown should probably be a ConstraintSet which generates its constraints on init, and its name should end with ConstraintSet. Then should be a function in interactive which accepts a Breakdown and does all the plotting; much of what are attributes of `self` now will be instead gotten from the passed Breakdown.

bqpd:
I'll take a stab at starting this.

bqpd:
Ready for review by @mayork

mayork:
@bqpd @whoburg I think this is ready for final review...I moved all the unit tests to t_constraints, all unit tests pass on my machine. Also fixed some diagram drawing problems that came up after moving things to interactive

not sure where the branch conflicts are...i can't find them

gpkit/constraints/breakdown.py
```diff
@@ -31,8 +31,17 @@ def __init__(self, input_dict, units):
         self.height = 20
 
         constraints = [1 >= sum(constr[1])/constr[0] for constr in self.constr]
+        self.constraints = constraints
         ConstraintSet.__init__(self, constraints)
 
+    def solve_method(self):
```
bqpd: What's wrong with Breakdown just being a constraintset?

gpkit/tests/t_constraints.py
```diff
@@ -125,8 +126,31 @@ def test_init(self):
         self.assertTrue(isinstance(sc, SignomialInequality))
         self.assertFalse(isinstance(sc, Posynomial))
 
-
-TESTS = [TestConstraint, TestMonomialEquality, TestSignomialInequality]
+class TestBreakdown(unittest.TestCase):
```
bqpd: Niiice

bqpd: Unfortunately putting the test here leads to it being tested once for every installed solver...

bd.py
```diff
@@ -3,11 +3,13 @@
...
-TEST = {'w': {'w1': {'w11':[3, "N"], 'w12':{'w121':[2, "N"], 'w122':[6, "N"]}},
-                'w2': {'w21':[1, "N"], 'w22':[2, "N"]}, 'w3':[1, "N"]}}
+TEST ={'w': {'w1': {'w11':[3, "N"], 'w12':{'w121':[2, "N"], 'w122':[6, "N"]}},
+              'w2': {'w21':[1, "N"], 'w22':[2, "N"]}, 'w3':[1, "N"]}}
+              
+
 BD = Breakdown(TEST, "N")
 m = Model(BD.varlist[0], BD)
-sol = m.solve()
+sol = BD.solve_method()#m.solve()
```
bqpd: m is now unused.

bqpd:
@mayork the conflicts are probably from some PRs recently merged to master, I can handle that. Breakdown still doesn't match the gpkit code style in some significant ways; let's meet up in person and go over the code before final review by @whoburg.

mayork:
@bqpd sounds good, when would you like to meet?

bqpd:
retest this please

bqpd:
Ready to merge once TODOs are cleaned up and `bd.py` is removed. There should potentially be a test that the correct SVG is produced?

mayork:
@whoburg @bqpd took care of all of the TODOs, just need to remove bd.py

Also, what kind of test did you want to verify the correct SVG is produced? Should I just save a sample file into gpkit somewhere?

gpkit/constraints/breakdown.py
```diff
@@ -43,11 +46,8 @@ def _recurse(self, input_dict, units):
 
     def make_diagram(self, sol, sidelength=12, height=20):
```
bqpd: I think filename should be an argument here, not in `__init__`

bqpd:
Awesome! Reviewing...

gpkit/interactive/svg.py
```diff
@@ -4,85 +4,74 @@
...
     calls the draw function
     """
-    order = sorted(input_dict.keys())
-    i = 0
     totalheight = 0
     currentlevel = currentlevel+1
-    while i < len(order):
-        height = int(round((((bd.height/bd.total)*sol(order[i])))))
-        name = order[i]
+    for key in sorted(input_dict):
```
bqpd: if you use `for key, value in sorted(input_dict.items()):` you should be able to replace `input_dict[key]` below with `value`.

mayork: Yeah I think the most current version already has that...is it possible you're looking at an outdated diff?

gpkit/interactive/svg.py
```diff
@@ -0,0 +1,78 @@
...
+    calls the draw function
+    """
+    totalheight = 0
+    currentlevel = currentlevel+1
+    for key in sorted(input_dict):
+        height = int(round((((sheight/total)*sol(key)))))
+        currentcoord = (initcoord[0], initcoord[1]+totalheight)
+        drawsegment(key, height, currentcoord, dwg, elementlength)
+        totalheight = totalheight+height
+        if isinstance(input_dict[key], dict):
```
bqpd: @mayork, this still seems to be in the latest changes. Perhaps you've a local commit you haven't pushed yet?

mayork: @bqpd oh wow sorry I'm an idiot and totally misunderstood what the first comment was about. For some reason my git is saying you'r commenting on an outdated diff...but best as I can tell this is the most current thing there is, it seems the same as what I have locally

bqpd: No worries!

mayork: @bqpd  just made line 36 of svg.py read this on my local copy

if isinstance(value, dict):

and I get this error: NameError: global name 'value' is not defined

I just tried to replace input_dict[key] with value...what did i do wrong?

bqpd: Line 31 should be `for key, value in sorted(input_dict.items()):`

mayork:
@bqpd just pushed a new version with your changes

mayork:
@whoburg I looked over this again and think it's ready for review...do you want any sort of test method? I'm not entirely sure what that would/should look like

whoburg:
just rebased this with @mayork, force pushing

whoburg:
retest this please

mayork:
screen shot 2016-08-30 at 2 39 06 pm
this is the diagram the test method draws...seems potentially useful to have somewehre

bqpd:
Nice, this looks good.

mayork:
@whoburg this is ready for a final review. I think we should disable the remaining pylint warnings, but I didn't want to do that before it was reviewed

whoburg:
retest this please

whoburg:
@mayork I'd like to do  this review in two stages:
1. high-level comments/discussion (now)
2. a more detailed line-by-line review by me or @bqpd

For now, the high-level comments/discussion:

1. I want us to try using this on the 1682 model and see how it goes. Could you please work with @mjburton11 to try this out on a gpkit-models branch. My suspicion is that it won't make sense to specify a single weight breakdown input dict in one location -- rather, it will be more natural to have multiple Breakdown objects (wing, fuselage, etc) that are local to their submodels/components -- but not sure how this will really play out.

2. Related to the above, does it work for one of the leaves of Breakdown A to itself be a Breakdown?

3. Further related to the above, should we implement it that way, i.e. each Breakdown is only one level deep, its leaves always being either a Variable or a Breakdown?

4. Can we switch the font to math font? Have you tested a latex math symbol like `\rho`?

(@bqpd, any comments?)

mayork:
1. I'll get with Mike and set up a 1682 breakdown. 

2. Have not tested that...however, if it doesn't work should be an easy feature to add simply by appending the constraints from the other breakdown

3. What do you see as the benefit of that method? I'm not sure I fully understand what you're going for 

4. Changing font is possible...I looked into it before, it was a while ago so I don't recall why I didn't do it. I'm not sure on math type, I'm guessing there's a way to do it but I need to investigate

whoburg:
The benefit of 3 would be a cleaner implementation in that the data structures and parsing would match up. To parse an input dict, for all values in a level of the dict you would a) create a variable, if the value is a string or Variable, or b) create a Breakdown, if the value is a dict.

Additional benefit is that it would be always possible to draw the breakdown diagram associated with a subset of a Breakdown, e.g. Aircraft has a Breakdown, but its Wing is also a breakdown.

Start by addressing 1 and I suspect 3 may end up being related.

mayork:
@whoburg more detailed answer to 4 from above. new fonts and styles can be defined as a new CSS style...I'm not venturing outside my knowledge base but from what I know/can tell from some quick web research CSS has limited support for math, so I don't think we can go display math font. How big of a problem is that? I think it would be nice to be able to do, but I don't think it's that big of an issue

whoburg:
I would be surprised if the svg package does not support latex math. But I've been surprised before.

In other news, looks like we have some conflicts; I'll attempt a rebase.

mayork:
screen shot 2016-09-06 at 8 54 56 pm
screen shot 2016-09-06 at 8 55 09 pm
screen shot 2016-09-06 at 8 55 33 pm
screen shot 2016-09-06 at 8 55 19 pm

@mjburton11 @whoburg 16.82 gas_male breakdowns. the code to make them is below.

```python
    #create the centbreakdown
    BDcent = {'Center': {'Fuel': sol('W_{fuel-tot}'), 'Payload': sol('W_{pay}'),
                                'Engine': sol('W_{eng-tot}'), 'Avionics': sol('W_{avionics}'),
                                'Skid': sol('W_{skid}')}}
    centbd = Breakdown(BDcent, "lbf")
    mcent = Model(centbd.root, centbd)
    centsol = mcent.solve()
    centbd.make_diagram(centsol,  "16.82_cent_weight_breakdown.svg", 7, 20)

    #create the fuselage breakdown
    BDfuse = {'Fuselage': {'Ribs': [sol('m_{rib}')*9.8*units('m/s^2'), "N"],
                           'Skin': [sol('m_{fuse}')*9.8*units('m/s^2'), "N"]}}
    fusebd = Breakdown(BDfuse, "lbf")
    mfuse = Model(fusebd.root, fusebd)
    fusesol = mfuse.solve()
    fusebd.make_diagram(fusesol, "16.82_fuse_weight_breakdown.svg", 7, 12)

    #create the wing breakdown
    BDwing = {'Wing': {'Caps': [sol('m_{cap}')*9.8*units('m/s^2'), "N"],
                       'Skin':[sol('m_{skin}')*9.8*units('m/s^2'), "N"]}}
    wingbd = Breakdown(BDwing, "lbf")
    mwing = Model(wingbd.root, wingbd)
    wingsol = mwing.solve()
    wingbd.make_diagram(wingsol, "16.82_wing_weight_breakdown.svg", 7, 12)

    #create the total breakdown
    total = dict(BDwing.items() + BDfuse.items() + BDcent.items())
    BDtotal = {'Total': total}
    fullbd = Breakdown(BDtotal, "lbf")
    mfull = Model(fullbd.root, fullbd)
    bdsol = mfull.solve(solver="mosek")
    fullbd.make_diagram(bdsol, "16.82_weight_breakdown.svg", 10, 22)
```

mayork:
@whoburg you cannot give a Breakdown another Breakdown as an arugment. However, I think the way it currently works is still close to your vision.

Breakdown is really just a plotting tool. The user supplies a Breakdown with a nested input dict which defines the breakdown. Breakdown provides a convenient method to sum weights in this input dict, but that does not need to be used if the values for all the weights are already known. If the user wants multiple breakdowns like I made above, you can just make multiple sub dicts and combine them before making a Breakdown. What are your thoughts on this? I don't think its too user intrusive.

bqpd:
screen shot 2016-02-02 at 5 29 03 pm
screen shot 2016-02-02 at 5 29 11 pm

whoburg:
I still feel like this is dying to be a ConstraintSet that can be included as a model building block. The drawing functionality is a natural use for `process_result`.

In the current form I'm worried that models will get out of sync from their corresponding Breakdown.

I think all the work done here on drawing will be super useful, and I don't think defining the recursive class will be too bad. I can take a shot at this at some point if it's confusing what I'm suggesting here.

mayork:
What do you want the input to the building block constraint set to be?

mayork:
Also, I can adjust line styles etc. is there a preference on what they should look like?

whoburg:
1. I would think there would be two main inputs:

```python
Arguments
---------
root (varkey):
    top-level variable to break down
leaves (iterable):
    list of (variables or other breakdowns) that sum to form root
```

The resulting constraint pseudocode is something like

```python
class Breakdown(Model):
    def __init__(self, **kwargs):
        constraints = [root >= sum(getattr(x, "root", x) for x in leaves)]
        constraints.extend([x for x in leaves if hasattr(x, "root")])  # add sub-breakdown constraints
        self.root = root
        Model.__init__(root, constraints, **kwargs)
```
... and then the only hard part is drawing. Unless I'm missing something hard?

On style, I think we should try to make them look like the latex examples posted above.

mayork:
that should be a pretty easy refactor on making the breakdown constraint set the way you want...I think most of the current code can actually still be used.

As far as the drawings do we want to actually make them tin latex? I'm rather amateur with latex so I'm not sure how hard it would be to make the switch.  I know @bqpd suggested svg, but there might be advantages to latex (namely simpler printing of math symbols)

bqpd:
Yeah you probably do want to make them in latex, though might be worth
doing svg as well for faster interactive display. Latex is just too slow
for interactive use.

On Sep 13, 2016 6:25 PM, "mayork" <notifications@github.com> wrote:

that should be a pretty easy refactor on making the breakdown constraint
set the way you want...I think most of the current code can actually still
be used.

As far as the drawings do we want to actually make them tin latex? I'm
rather amateur with latex so I'm not sure how hard it would be to make the
switch. I know @bqpd <https://github.com/bqpd> suggested svg, but there
might be advantages to latex (namely simpler printing of math symbols)

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/hoburg/gpkit/pull/522#issuecomment-246845317>, or mute
the thread

whoburg:
yeah tough call there. Possibly useful: http://tex.stackexchange.com/questions/51757/how-can-i-use-tikz-to-make-standalone-svg-graphics

PR #531, opened 2016-03-02
pylint: allow single character variable names

whoburg:
pylint by default requires variable names to be three or more characters. This reduces that requirement to one character.


whoburg:
I'm going to go ahead and merge this in an effort to address pylint issues in several other PR's.

If anyone disagrees with the decision please let me know.

PR #532, opened 2016-03-02
reorganize GeometricProgram.solve(); fixes #526

### LINKED ISSUE #526: unexpected exp overflow
The following problem is dual-infeasible:
```python
import numpy as np

CL = Variable("CL")
CD = Variable("CD")
V = Variable("V", "m/s", "cruise speed")
W = Variable("W", 200, "lbf", "weight")
rho = Variable(r"\rho", 1.2, "kg/m^3", "air density")
S = Variable("S", 190, "ft^2", "wing area")
m = Model(V*(W/(CL/CD)),
          [0.5*rho*CL*S*V**2 >= W,
           CL**1.51/CD <= 20])
sol = m.solve("mosek")
```
Mosek indeed returns a `DUAL_INFEAS_CER`. Surprisingly though, running the code above results in an overflow error,
```
gpkit/geometric_program.py:156: RuntimeWarning: overflow encountered in exp
  np.exp(primal)))
```
It seems to me that we should not be taking the `exp` of the primal solution if the solver returns `DUAL_INFEAS_CER`.

We should track this down and fix it to make sure it's not representative of a bigger issue.

whoburg:
- avoids calling `np.exp` for non-optimal solutions (which resulted in a `RuntimeWarning` in #526)
 - still allows user to create the results dict for non-optimal models as described in error message edited in this PR

### DIFF
--- a/gpkit/geometric_program.py
+++ b/gpkit/geometric_program.py
@@ -148,9 +148,41 @@ class GeometricProgram(NomialData):
             print("Solving took %.3g seconds." % (soltime,))
             tic = time()
 
-        result = {}
-        # confirm lengths before calling zip
+        if solver_out.get("status", None) not in ["optimal", "OPTIMAL"]:
+            raise RuntimeWarning(
+                "final status of solver '%s' was '%s', not 'optimal'."
+                "\n\nThe infeasible solve's result is stored in"
+                " model.program.solver_out. A result dict can be generated "
+                " via program._compile_result(program.solver_out)."
+                " If the problem was Primal Infeasible, you can generate a"
+                " feasibility-finding relaxation with model.feasibility()." %
+                (solver, solver_out.get("status", None)))
+
+        result = self._compile_result(solver_out)
+
+        self.result = result  # NOTE: SIDE EFFECTS
+
+        if verbosity > 1:
+            print ("result packing took %.2g%% of solve time" %
+                   ((time() - tic) / soltime * 100))
+            tic = time()
+
         primal = np.ravel(solver_out['primal'])
+        self.check_solution(result["cost"],
+                            primal,
+                            nu=result["sensitivities"]["monomials"],
+                            la=result["sensitivities"]["posynomials"])
+
+        if verbosity > 1:
+            print ("solution checking took %.2g%% of solve time" %
+                   ((time() - tic) / soltime * 100))
+        return result
+
+    def _compile_result(self, solver_out):
+        """Compile result dict for returning in solve() from solver ouput"""
+        result = {}
+        primal = np.ravel(solver_out['primal'])
+        # confirm lengths before calling zip
         assert len(self.varlocs) == len(primal)
         result["variables"] = dict(zip(self.varlocs,
                                        np.exp(primal)))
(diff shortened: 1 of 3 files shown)


whoburg:
test this please

whoburg:
@bqpd or @pgkirsch, please review

gpkit/geometric_program.py
```diff
@@ -103,14 +105,16 @@ def solve(self, solver=None, verbosity=1, *args, **kwargs):
...
             from . import settings
             if settings['installed_solvers']:
-                solver = settings['installed_solvers'][0]
+                return settings['installed_solvers'][0]
             else:
                 raise ValueError("No solver was given; perhaps gpkit was not"
                                  " properly installed, or found no solvers"
                                  " during the installation process.")
+        if solver is None:
+            solver = _default_solver()
```
bqpd: Interesting, why do it like this?

gpkit/geometric_program.py
```diff
@@ -148,9 +152,41 @@ def solve(self, solver=None, verbosity=1, *args, **kwargs):
...
+                            np.ravel(solver_out['primal']),
+                            nu=result["sensitivities"]["monomials"],
+                            la=result["sensitivities"]["posynomials"])
+
+        if verbosity > 1:
+            print ("solution checking took %.2g%% of solve time" %
+                   ((time() - tic) / soltime * 100))
+        return result
+
+    def _compile_result(self, solver_out):
```
bqpd: Oh, interesting, so you can run it yourself if you want?

gpkit/geometric_program.py
whoburg: heh, if you want the truth, I did this because I was trying to the the damn pylint `too-many-branches` under the threshold

whoburg: but one could argue we might want the `default_solver` functionality all the way up at the gpkit level for other uses at some point?

gpkit/geometric_program.py
whoburg: right -- by making this a function we can avoid calling it when a solution is not optimal -- and we leave it up the the user whether they want to try passing `solver_out` into `_compile_result` -- with exp overflows being one possible result

My feeling is that for a user that is not debugging, getting an exp overflow feels icky.

whoburg: but the point is that before a user could look up the `results` dict even for failed solves -- so we want that to remain possible via this method.

gpkit/geometric_program.py
bqpd: Ehhhhh that's kind of the point of settings. I could see an argument that we have a `settings["default_solver"]` tho, which would also fix the branching issue.

gpkit/geometric_program.py
bqpd: Sounds good! I'll add it to the list of things to document.

gpkit/geometric_program.py
whoburg: Just implemented that -- see 6afb5b0623276bfe0d036709c4769eba5615717d and please review

whoburg:
retest this please

whoburg:
@bqpd, confirm you're happy with merging this?

bqpd:
How about confirming I'm happy by merging it? :p

PR #533, opened 2016-03-02
assorted pylint tweaks

whoburg:
- increase max branches from 12 to 15
 - line length 80 char
 - docstrings not required for unit test methods

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -169,7 +169,7 @@ method-name-hint=[a-z_][a-z0-9_]{2,30}$
 
 # Regular expression which should only match function or class names that do
 # not require a docstring.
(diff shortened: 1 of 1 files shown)


whoburg:
yarg... @galbramc I decreased the max line length and now I want to get this on master

galbramc:
All we need to do is reset the reference on both the pull request build and the push build. If you are happy with these settings you can go ahead and merge, and then we just reset the counts like I showed you.

galbramc:
Did you look at the number of methods restriction? There are also things about not defining all variables in the init method that pylint warns about.

whoburg:
ok will look at that some other time, too many other fires buring

PR #545, opened 2016-03-04
strawman resolution to issue534

whoburg:
here's a simple fix we can discuss. disadvantage is that perhaps the user should have to put calls to diff in a `SignomialsEnabled` environment

### DIFF
--- a/gpkit/nomials.py
+++ b/gpkit/nomials.py
@@ -151,7 +151,7 @@ class Signomial(NomialData):
         Signomial (or Posynomial or Monomial)
         """
         deriv = super(Signomial, self).diff(wrt)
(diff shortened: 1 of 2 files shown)


whoburg:
failing in part cause of the pint update

whoburg:
I wonder if there is a more clever way to compute a mono_approximation of a posynomial that avoids calling diff on terms with negative exponents? Like pass in the reciprocal for vars w/ negative exps?

galbramc:
test this please

galbramc:
test this please

bqpd:
@whoburg I'm ready to merge this.

PR #493, opened 2015-11-30
Newconstr and keydict

bqpd:
@pgkirsch, this should work for your needs...

todo:
  - [x] ConstraintSet docstrings
  - [x] pass tests
  - [x] unify and fix ConstraintSet/GeometricProgram/etc printing
  - [x] hand test `gpkit.interactive` features still work
  - [x] make sure runtimes are competitive with master branch even for large models

PR #547, opened 2016-03-05
Markdown parsing

bqpd:
This PR allows a markdown file to be run as python code. For example: 

```python
import gpkit; exec gpkit.mdmake("gas_hale.md")
```

will run the code in this [example markdown file](https://raw.githubusercontent.com/hoburg/gpkit-models/1682literate/1682/gas_hale.md).

This provides immediate model documentation (including plots) on Github. [Check it out!](https://github.com/hoburg/gpkit-models/blob/1682literate/1682/gas_hale.md)

A secondary advantage is that running `pandoc -o gas_hale.pdf gas_hale.md` produces [this nice pdf](https://github.com/hoburg/gpkit/files/159718/gas_hale.pdf)

### DIFF
--- a/gpkit/__init__.py
+++ b/gpkit/__init__.py
@@ -28,16 +28,26 @@ SETTINGS_PATH = os_sep.join([os_path_dirname(__file__), "env", "settings"])
 
 def mdparse(filename):
     with open(filename) as f:
-        python_lines = []
+        lines = []
         in_block = False
         for line in f:
+            content = "\n"
             if line == "```python\n":
                 in_block = True
-            elif line == "```\n":
+            elif in_block and line == "```\n":
                 in_block = False
             elif in_block:
-                python_lines.append(line)
-        return "".join(python_lines)
+                content = line
+            elif line != "\n":
+                content = "# " + line
+            lines.append(content)
+        return "".join(lines)
+
+
+def mdmake(filename):
+    with open(filename+".py", "w") as f:
+        f.write(mdparse(filename))
+    return open(filename+".py")
 
 
 def enable_units(path=UNITDEF_PATH):
--- a/gpkit/__init__.py
+++ b/gpkit/__init__.py
@@ -26,6 +26,20 @@ UNITDEF_PATH = os_sep.join([os_path_dirname(__file__), "gpkit_units.txt"])
 SETTINGS_PATH = os_sep.join([os_path_dirname(__file__), "env", "settings"])
 
 
+def mdparse(filename):
+    with open(filename) as f:
+        python_lines = []
+        in_block = False
+        for line in f:
+            if line == "```python\n":
+                in_block = True
+            elif line == "```\n":
+                in_block = False
+            elif in_block:
+                python_lines.append(line)
+        exec("\n".join(python_lines))
+
+
 def enable_units(path=UNITDEF_PATH):
     """Enables units support in a particular instance of GPkit.
 
--- a/gpkit/__init__.py
+++ b/gpkit/__init__.py
@@ -37,7 +37,7 @@ def mdparse(filename):
                 in_block = False
             elif in_block:
                 python_lines.append(line)
-        exec("\n".join(python_lines))
+        return "".join(python_lines)
 
 
 def enable_units(path=UNITDEF_PATH):
--- a/gpkit/__init__.py
+++ b/gpkit/__init__.py
@@ -26,27 +26,51 @@ UNITDEF_PATH = os_sep.join([os_path_dirname(__file__), "gpkit_units.txt"])
 SETTINGS_PATH = os_sep.join([os_path_dirname(__file__), "env", "settings"])
 
 
-def mdparse(filename):
+def mdparse(filename, return_tex=False):
     with open(filename) as f:
-        lines = []
-        in_block = False
+        py_lines = []
+        texmd_lines = []
+        block_idx = 0
+        in_replaced_block = False
         for line in f:
-            content = "\n"
-            if line == "```python\n":
-                in_block = True
-            elif in_block and line == "```\n":
-                in_block = False
-            elif in_block:
-                content = line
-            elif line != "\n":
-                content = "# " + line
-            lines.append(content)
-        return "".join(lines)
+            line = line[:-1]  # remove newline
+            texmd_content = line if not in_replaced_block else ""
+            texmd_lines.append(texmd_content)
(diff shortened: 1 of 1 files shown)


bqpd:
@whoburg, @galbramc, curious for your feedback on this. Useful for 16.82?

bqpd:
Thinking of adding the following as a shell script or makefile: @whoburg, thoughts?

```bash
for var in "$@"
do
    python -c "import gpkit; exec gpkit.mdmake($var)"
    if [ -e "default.latex"]; then
        pandoc -c "pandoc -o $var.pdf  $var --template default.latex
    else
        pandoc -c "pandoc -o $var.pdf  $var
    done
done
```

bqpd:
Also keeping an eye open for ways we could indicate that for the PDF a particular code block should be removed or replaced by a particular `.tex` file, so that `model.md` could compile directly to one of the subsections of @pgkirsch's paper, etc.

bqpd:
Implemented the above in latest commit.

bqpd:
I'm digging the idea of starting with code and incrementally replacing it with latex that describes it: the bits of the PDF that are still code are the bits that need more documentation work... @pgkirsch, you're the expert on gpkit papers, does this seem reasonable? (with an overall tex file that imports all the submodel texs, etc.)

pgkirsch:
hmmm this is an interesting idea, and seems like it could be really helpful for documentation and paper writing.

A couple initial reactions:
- It feels odd that we are trying to optimize/automate the GPkit paper-writing process when the group only has a couple paper's worth of experience. Isn't this a little premature?
- I'm worried about adding syntax to GPkit scripts that makes them less easy to understand. One of the nice things about GPkit at the moment is that it's all very lightweight and easy to pickup. It's not like your proposed additions are overly complex, but it's just more things for a new user to learn/understand. I also like the idea of keeping the GPkit interface static for just a little while to let the user base grow before adding loads of bells and whistles. I feel like the flavor of GP changes so often that it might be frustrating for people to have to keep getting up to speed with new interfaces.
- I'm also worried about making files longer than they need to be, and adding overhead.

That probably sounds really negative, sorry. I think it's cool stuff and really nice for auto-generating documentation, but is there really a demand for it yet when we can't even debug models that don't work/don't have that many models in our model library?

bqpd:
Good points!

re 1, this experiment is definitely a little premature. :) But Woody and I were looking at the 16.82 repo and saw latex documentation of models which was just copy-pasted comments variables and constraints, so it might be immediately useful.

re 2, One nice thing about this is that it doesn't add any new python syntax.

But, re 3, it does increase the complexity of the model file by incorporating more documentation into it. The hope is that this reduces the overall difficulty of documenting a model (see for instance the picture you posted for the atmospheric model) and the likelihood that documentation and code will slip out of sync.

whoburg:
retest this please

whoburg:
Great to see an initial proposal for how to tie together model implementation with documentation / derivation.

A few thoughts:
 - is it possible to put raw latex math in the `.md`? This is critical -- e.g. someone might want to give the derivation that led up to some constraint in a model -- those derivation equations won't come from Python.
 - probably belongs in gpkit-models, but I think the latex template should `\usepackage{fullpage}` -- right now many code blocks are running off to the right. An 80 char block should look good.
 - Impressive that this is implemented in 24 lines -- let's keep it clean instead of implementing our own markdown language
 - on that note, not sure the code implementing `inPDF:` got pushed to the remote
 - how would imported models work, e.g. `from gpkit.models.atmosphere import Trposphere`?
 - At a high level I like the direction but I still have questions about the implementation. This proposal implements one option: models and documentation in `.md` files. I'm also wondering about a second option, where models live in `.py` files and explicitly create `.tex` output files, which can then be included in a separate latex file describing the model. Not saying this is right -- just want to think about pros / cons.

gpkit/__init__.py
```diff
@@ -26,6 +26,30 @@
 SETTINGS_PATH = os_sep.join([os_path_dirname(__file__), "env", "settings"])
 
 
+def mdparse(filename):
```
whoburg: docstring

gpkit/__init__.py
```diff
@@ -26,6 +26,30 @@
...
+                in_block = False
+            elif in_block:
+                content = line
+            elif line != "\n":
+                content = "# " + line
+            lines.append(content)
+        return "".join(lines)
+
+
+def mdmake(filename):
```
whoburg: docstring

whoburg:
one more comment: it feels a bit awkward to have the python implementation of a particular class (`class GasPoweredHALE(Model):`, in this case) broken up by text and documentation. E.g. it looks odd to see code indented so far. I imagine this would improve with model composition though, because everything will be split into separate models.

bqpd:
- Yup it's possible to put raw latex math in the `.md`; it's done in the example a couple of times

(moved second bullet to gpkit-models)

- Added a few more lines of code when
- pushing up the `inPDF` flag

- The easiest way to implement importing would be to give `gpkit.models` a function that imported the appropriate model with the docstring set to the md file (so that one could do `help(Troposphere)` to get the documentation)

- The second option seems quite reasonable, and requires no new lines of code! Note that the `.py` and `.tex.md` file written by `mdmake` are good starts for the two separate files, so this one-file method isn't removing that information. I think the biggest con for me is the concern that model documentation (explaining constants or constraints) is more likely to slip behind or be inconsistent with the actual model if you can't see them both on the same screen.

bqpd:
retest this please

galbramc:
test this please

whoburg:
```
whoburg$ pandoc -o aircraft_design.pdf aircraft_design.md.tex.md
! LaTeX Error: Environment longtable undefined.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...

l.7 \begin{longtable}

pandoc: Error producing PDF
```

whoburg:
I think it would make sense to move `mdparse` and `mdmake` to `gpkit.tools`, and then implement @bqpd's proposed shell script / make file, which can handle the import from `gpkit.tools`

bqpd:
Sounds like a good plan!

The pandoc error is because you didn't use the template including `longtable`

whoburg:
should that template be added to version control?

bqpd:
In gpkit-models? Quite possibly. In gpkit? I don't know...

bqpd:
This commit was merged into #493, closing

PR #555, opened 2016-03-08
get build dir from environment variable GPKITBUILD if it exists

bqpd:
@galbramc, see if this fits your needs...

### DIFF
--- a/gpkit/build.py
+++ b/gpkit/build.py
@@ -196,7 +196,10 @@ class Mosek(SolverBackend):
             return
 
         lib_dir = replacedir(pathjoin("gpkit", "_mosek", "lib"))
-        solib_dir = replacedir(pathjoin(os.path.expanduser("~"), ".gpkit"))
+        if "GPKITBUILD" in os.environ:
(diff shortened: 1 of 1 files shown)


whoburg:
retest this please

galbramc:
test this please

galbramc:
Hmmm, I added GPKITBUILD to the build scripts, but the directory was not created, and the unit tests failed.

galbramc:
I don't see in the build script where you actually create the 'solib_dir' directory. The compiler will not create the directory with the -o switch, at least I don't think it will.

galbramc:
I think you need something like this in build.py

```
        if !os.path.isdir(solib_dir):
            log("# Creating %s" % solib_dir)
            os.makedirs(solib_dir)
```

bqpd:
Oops, you're right. I forgot to also make the GPKITBUILD folder.

galbramc:
This might be why there are problems with pip install as well.

Once this is merged, we can enable full parallel builds.

PR #558, opened 2016-03-08
remove gpkit.models from setup.py

bqpd:
@galbramc


whoburg:
yup I expect this will do the trick

Issue #258, opened 2015-05-07
refactor multi-solver unit testing


whoburg:
We should revisit the structure of `t_model.py`, which currently runs every unit test for every solver, as well as other unit tests.
- [ ] Determine which tests should be run for all solvers (anything that requires a solver?), and which should not involve a solver at all.
- [ ] Reorganize unit tests according to above.
- [ ] As suggested by @bqpd, in tests where the outcome should not require a solver to be called, pass in some sort of null solver object to ensure this is the case.
- [ ] test that infeasible GP raises `RuntimeWarning` for all solvers 
- [ ] have a way for tests to specify which solvers they should be run for

bqpd:
Solver runtime on personal computers should stay below 5s; tests can be moved from running for all solvers to running for only one solver as necessary.

bqpd:
We have all this! Somewhat ad-hoc, but this isn't the issue to organize it...

PR #571, opened 2016-03-10
add boundedness debugging tool

bqpd:
- [x] Turn into `BoundedConstraintSet` and move functions into `__init__` and `sens_from_dual/sens_from_gpconstr`, with a verbosity flag to print during solve.

### DIFF
--- a/docs/source/advancedcommands.rst
+++ b/docs/source/advancedcommands.rst
@@ -233,3 +233,24 @@ Example Usage
 The ``normsub`` argument specifies an expected value for your solution to normalize the different :math:`g_i` (you can also do this by hand). The feasibility of the problem should not depend on the normalization, but the spacing of the sweep will.
 
 The ``sweep`` argument specifies what points between 0 and 1 you wish to sample the weights at. If you want different resolutions or spacings for different weights, the ``sweeps`` argument accepts a list of sweep arrays.
+
+
+Debugging
+=========
+
+Unbounded variables
+-------------------
+In some cases a solver will return ``unknown`` when a free variable is
+unbounded by the model. ``gpkit.tools.determine_unbounded_variables`` is a
+simple tool that attempts to detect unbounded variables by adding extremely
+large upper bounds and extremely small lower bounds to all variables in a model,
+resolving, and checking whether any variables slide off to the bounds.
+Mosek returns unknown when attempting to solve the following model:
+
+.. literalinclude:: examples/unbounded.py
+
+Upon viewing the output from ``determine_unbounded_variables``,
+
+.. literalinclude:: examples/unbounded_output.txt
+
+it becomes clear that ``A`` is unbounded below in the original model.
--- a/docs/source/advancedcommands.rst
+++ b/docs/source/advancedcommands.rst
@@ -243,14 +243,16 @@ Unbounded variables
 In some cases a solver will return ``unknown`` when a free variable is
 unbounded by the model. ``gpkit.constraints.bounded.BoundedConstraintSet`` is a
 simple tool that attempts to detect unbounded variables by adding extremely
-large upper bounds and extremely small lower bounds to all variables in a ConstraintSet,
-resolving, and checking whether any variables slide off to the bounds.
-Mosek returns unknown when attempting to solve the following model:
+large upper bounds and extremely small lower bounds to all variables in a ConstraintSet.
+
+When a model if with a BoundedConstraintSet is solved, it checks whether any variables slid off to the bounds, notes this in the solution dictionary and prints a warning (if verbosity is greater than 0).
+
+For example, Mosek returns unknown when attempting to solve the following model:
 
 .. literalinclude:: examples/unbounded.py
 
-Upon viewing the output from ``determine_unbounded_variables``,
+Upon viewing the printed output,
 
 .. literalinclude:: examples/unbounded_output.txt
 
-it becomes clear that ``A`` is unbounded below in the original model.
+it becomes clear that ``x`` is unbounded below in the original model.
--- a/docs/source/advancedcommands.rst
+++ b/docs/source/advancedcommands.rst
@@ -240,14 +240,14 @@ Debugging
 
 Unbounded variables
 -------------------
-In some cases a solver will return ``unknown`` when a free variable is
-unbounded by the model. ``gpkit.constraints.bounded.BoundedConstraintSet`` is a
-simple tool that attempts to detect unbounded variables by adding extremely
-large upper bounds and extremely small lower bounds to all variables in a ConstraintSet.
+In some cases a model will not solve because its variables are pushing to 0 or infinity. If the solver catches such behaviour it will return ``dual infeasible`` (or equivalent), but sometimes solvers do not catch it and return ``unknown``.
 
-When a model if with a BoundedConstraintSet is solved, it checks whether any variables slid off to the bounds, notes this in the solution dictionary and prints a warning (if verbosity is greater than 0).
+``gpkit.constraints.bounded.BoundedConstraintSet`` is a
+simple tool that attempts to detect unbounded variables and get unbounded models to solve by adding extremely large upper bounds and extremely small lower bounds to all variables in a ConstraintSet.
 
-For example, Mosek returns unknown when attempting to solve the following model:
+When a model with an BoundedConstraintSet is solved, it checks whether any variables slid off to the bounds, notes this in the solution dictionary and prints a warning (if verbosity is greater than 0).
+
+For example, Mosek returns ``DUAL_INFEAS_CER`` when attempting to solve the following model:
 
 .. literalinclude:: examples/unbounded.py
 
@@ -255,4 +255,4 @@ Upon viewing the printed output,
 
 .. literalinclude:: examples/unbounded_output.txt
 
-it becomes clear that ``x`` is unbounded below in the original model.
+it becomes clear that the problem is, unsurprisingly, that ``x`` is unbounded below in the original model.
--- a/docs/source/advancedcommands.rst
+++ b/docs/source/advancedcommands.rst
@@ -241,9 +241,9 @@ Debugging
 Unbounded variables
 -------------------
 In some cases a solver will return ``unknown`` when a free variable is
-unbounded by the model. ``gpkit.tools.determine_unbounded_variables`` is a
+unbounded by the model. ``gpkit.constraints.bounded.BoundedConstraintSet`` is a
 simple tool that attempts to detect unbounded variables by adding extremely
-large upper bounds and extremely small lower bounds to all variables in a model,
+large upper bounds and extremely small lower bounds to all variables in a ConstraintSet,
 resolving, and checking whether any variables slide off to the bounds.
 Mosek returns unknown when attempting to solve the following model:
 
--- a/docs/source/examples/unbounded.py
+++ b/docs/source/examples/unbounded.py
@@ -1,0 +1,20 @@
+from gpkit import Variable, Model
+from gpkit.tools import determine_unbounded_variables
+
+A = Variable("A")
+D = Variable("D")
+F = Variable("F")
+mi = Variable("m_i")
+
+Fs = Variable("Fs", 0.9)
+mb = Variable("m_b", 0.4)
+V = Variable("V", 300)
+
+m = Model(F, [F >= D + V**2,
+              D == V**2*A,
+              V >= mi + mb,
+              Fs <= mi,
+             ])
+# sol = m.solve("mosek")  # returns UNKNOWN
+bounds = determine_unbounded_variables(m, "mosek")
+print bounds
--- a/docs/source/examples/unbounded.py
+++ b/docs/source/examples/unbounded.py
@@ -1,5 +1,5 @@
 from gpkit import Variable, Model, settings
-from gpkit.tools import determine_unbounded_variables
+from gpkit.tools import BoundedConstraintSet
 
 A = Variable("A")
 D = Variable("D")
@@ -10,12 +10,13 @@ Fs = Variable("Fs", 0.9)
 mb = Variable("m_b", 0.4)
 V = Variable("V", 300)
 
-m = Model(F, [F >= D + V**2,
-              D == V**2*A,
-              V >= mi + mb,
-              Fs <= mi,
-             ])
-# sol = m.solve("mosek")  # returns UNKNOWN
+constraints = [F >= D + V**2,
+               D == V**2*A,
+               V >= mi + mb,
+               Fs <= mi,
+               ]
+
+# Model(F, constraints).solve("mosek")  # returns UNKNOWN
 if "mosek" in settings["installed_solvers"]:
-    bounds = determine_unbounded_variables(m, "mosek")
-    print bounds
+    sol = Model(F, BoundedConstraintSet(constraints)).solve("mosek")
+    print sol["boundedness"]
--- a/docs/source/examples/unbounded.py
+++ b/docs/source/examples/unbounded.py
@@ -1,4 +1,4 @@
-from gpkit import Variable, Model
+from gpkit import Variable, Model, settings
 from gpkit.tools import determine_unbounded_variables
 
 A = Variable("A")
@@ -16,5 +16,6 @@ m = Model(F, [F >= D + V**2,
               Fs <= mi,
              ])
 # sol = m.solve("mosek")  # returns UNKNOWN
-bounds = determine_unbounded_variables(m, "mosek")
-print bounds
+if "mosek" in settings["installed_solvers"]:
+    bounds = determine_unbounded_variables(m, "mosek")
+    print bounds
--- a/docs/source/examples/unbounded.py
+++ b/docs/source/examples/unbounded.py
@@ -1,13 +1,13 @@
 "Demonstrate a trivial unbounded variable"
-from gpkit import Variable, Model, settings
+from gpkit import Variable, Model
 from gpkit.constraints.bounded import BoundedConstraintSet
 
 x = Variable("x")
 
-constraints = [x <= 1]
+constraints = [x >= 1]
 
 # Model(x, constraints).solve()  # does not solve
-m = Model(x, BoundedConstraintSet(constraints))
+m = Model(1/x, BoundedConstraintSet(constraints))
 # by default, prints bounds warning during solve
 sol = m.solve(verbosity=0)
 # bound waring is available in sol["boundedness"]:
--- a/docs/source/examples/unbounded.py
+++ b/docs/source/examples/unbounded.py
@@ -6,7 +6,7 @@ x = Variable("x")
 
 constraints = [x >= 1]
 
-# Model(x, constraints).solve()  # does not solve
+# Model(1/x, constraints).solve()  # does not solve
 m = Model(1/x, BoundedConstraintSet(constraints))
 # by default, prints bounds warning during solve
 sol = m.solve(verbosity=0)
--- a/docs/source/examples/unbounded.py
+++ b/docs/source/examples/unbounded.py
@@ -10,5 +10,5 @@ constraints = [x >= 1]
 m = Model(1/x, BoundedConstraintSet(constraints))
 # by default, prints bounds warning during solve
 sol = m.solve(verbosity=0)
-# bound waring is available in sol["boundedness"]:
-print sol["boundedness"]
+print sol.table()
+print "sol['boundedness'] is:", sol["boundedness"]
--- a/docs/source/examples/unbounded.py
+++ b/docs/source/examples/unbounded.py
@@ -1,5 +1,5 @@
 from gpkit import Variable, Model, settings
-from gpkit.tools import BoundedConstraintSet
+from gpkit.constraints.bounded import BoundedConstraintSet
 
 A = Variable("A")
 D = Variable("D")
@@ -16,7 +16,10 @@ constraints = [F >= D + V**2,
                Fs <= mi,
                ]
 
-# Model(F, constraints).solve("mosek")  # returns UNKNOWN
+# Model(F, constraints).solve("mosek")  # does not solve
 if "mosek" in settings["installed_solvers"]:
-    sol = Model(F, BoundedConstraintSet(constraints)).solve("mosek")
+    m = Model(F, BoundedConstraintSet(constraints))
+    # by default, prints bounds warning during solve
+    sol = m.solve("mosek", verbosity=0)
+    # bound waring is also accessible in sol:
     print sol["boundedness"]
--- a/docs/source/examples/unbounded.py
+++ b/docs/source/examples/unbounded.py
@@ -1,25 +1,14 @@
+"Demonstrate a trivial unbounded variable"
 from gpkit import Variable, Model, settings
 from gpkit.constraints.bounded import BoundedConstraintSet
 
-A = Variable("A")
-D = Variable("D")
-F = Variable("F")
-mi = Variable("m_i")
+x = Variable("x")
 
-Fs = Variable("Fs", 0.9)
-mb = Variable("m_b", 0.4)
-V = Variable("V", 300)
+constraints = [x <= 1]
 
-constraints = [F >= D + V**2,
-               D == V**2*A,
-               V >= mi + mb,
-               Fs <= mi,
-               ]
-
-# Model(F, constraints).solve("mosek")  # does not solve
-if "mosek" in settings["installed_solvers"]:
-    m = Model(F, BoundedConstraintSet(constraints))
-    # by default, prints bounds warning during solve
-    sol = m.solve("mosek", verbosity=0)
-    # bound waring is also accessible in sol:
-    print sol["boundedness"]
+# Model(x, constraints).solve()  # does not solve
+m = Model(x, BoundedConstraintSet(constraints))
+# by default, prints bounds warning during solve
+sol = m.solve(verbosity=0)
+# bound waring is available in sol["boundedness"]:
+print sol["boundedness"]
--- a/gpkit/constraints/bounded.py
+++ b/gpkit/constraints/bounded.py
@@ -1,0 +1,68 @@
+"Implements BoundedConstraintSet"
+from collections import defaultdict
+import numpy as np
+
+from .. import Variable
+from .set import ConstraintSet
+from ..small_scripts import mag
+
+
+def add_bounds(varkeys, lower, upper):
+    "Returns constraints list bounding all varkeys"
+    constraints = []
+    for varkey in varkeys:
+        units = varkey.descr.get("units", 1)
+        constraints.append([upper*units >= Variable(**varkey.descr),
+                            Variable(**varkey.descr) >= lower*units])
+    return constraints
+
+
+class BoundedConstraintSet(ConstraintSet):
+    "Bounds contained variables so as to ensure dual feasibility."
+
+    def __init__(self, constraints, substitutions=None, verbosity=1,
+                 eps=1e-30, lower=None, upper=None):
+        self.bound_las = None
+        self.verbosity = verbosity
+        self.lowerbound = lower if lower else eps
+        self.upperbound = upper if upper else 1/eps
+        constraintset = ConstraintSet([constraints])
+        self.bounded_varkeys = tuple(vk for vk in constraintset.varkeys
+                                     if "value" not in vk.descr)
+        bounding_constraints = add_bounds(self.bounded_varkeys,
+                                          self.lowerbound, self.upperbound)
+        constraints = [bounding_constraints, constraintset]
+        super(BoundedConstraintSet, self).__init__(constraints, substitutions)
+
+    def sens_from_dual(self, las, nus):
+        "Return sensitivities while capturing the relevant lambdas"
+        self.bound_las = las[:2*len(self.bounded_varkeys)]
+        return super(BoundedConstraintSet, self).sens_from_dual(las, nus)
+
+    def process_result(self, result):
+        "Creates (and potentially prints) a dictionary of unbounded variables."
+        lam = self.bound_las
+        out = defaultdict(list)
+        for i, varkey in enumerate(self.bounded_varkeys):
+            lam_gt, lam_lt = lam[2*i], lam[2*i+1]
+            if abs(lam_gt) >= 1e-7:  # arbitrary threshold
+                out["sensitive to upper bound"].append(varkey)
+            if abs(lam_lt) >= 1e-7:  # arbitrary threshold
+                out["sensitive to lower bound"].append(varkey)
+            value = mag(result["variables"][varkey])
+            distance_below = np.log(value/self.lowerbound)
+            distance_above = np.log(self.upperbound/value)
+            if distance_below <= 3:  # arbitrary threshold
+                out["value near lower bound"].append(varkey)
+            elif distance_above <= 3:  # arbitrary threshold
+                out["value near upper bound"].append(varkey)
+        if self.verbosity > 0:
+            if out:
+                print
+                print "UNBOUNDED VARIABLES"
+                for key, value in out.items():
+                    print "% 25s: %s" % (key, value)
+                print
+        if not "boundedness" in result:
+            result["boundedness"] = {}
+        result["boundedness"].update(out)
--- a/gpkit/constraints/bounded.py
+++ b/gpkit/constraints/bounded.py
@@ -7,18 +7,54 @@ from .set import ConstraintSet
 from ..small_scripts import mag
 
 
-def add_bounds(varkeys, lower, upper):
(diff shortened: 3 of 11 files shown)


bqpd:
@whoburg, ready for review.

gpkit/tools.py
```diff
@@ -112,3 +115,53 @@ def mdmake(filename, make_tex=True):
...
+# pylint: disable=too-many-locals
+def determine_unbounded_variables(model, solver=None, verbosity=0,
+                                  eps=1e-30, lower=None, upper=None, **kwargs):
+    "Returns labeled dictionary of unbounded variables."
+    m = bound_all_variables(model, eps, lower, upper)
+    sol = m.solve(solver, verbosity, **kwargs)
+    lam = sol["sensitivities"]["la"][1:]
+    out = defaultdict(list)
+    for i, varkey in enumerate(m.bound_all["varkeys"]):
+        lam_gt, lam_lt = lam[2*i], lam[2*i+1]
```
whoburg: hmm, is the iteration order keys in a dict stable? Could this indexing be fragile as a result?

bqpd: ["varkeys"] is a list, though that issue comes up in geometric_program.py

gpkit/tools.py
```diff
@@ -112,3 +115,53 @@ def mdmake(filename, make_tex=True):
...
+            distance_above = np.log(m.bound_all["ub"]/value)
+            if distance_below <= 3:
+                out["hit eps lower bound"].append(varkey)
+            elif distance_above <= 3:
+                out["hit 1/eps upper bound"].append(varkey)
+           #  if distance_below <= 3 or distance_above <= 3:
+           #      # if we're within ~10x of a boundary (arbitrary threshold)
+           #      raise AttributeError("%s appears to have hit a boundary"
+           #                           " but it's not showing up in the"
+           #                           " sensitivities." % varkey)
```
whoburg: should this commented out code stay?

bqpd: @pgkirsch and I are still discussing what constitutes a boundary "hit":
just sensitivities, or sensitivities and position?

gpkit/tools.py
```diff
@@ -112,3 +115,53 @@ def mdmake(filename, make_tex=True):
...
+        if lam_gt and lam_lt:  # to avoid divide-by-zero
+            sens_ratio = lam_gt/lam_lt
+        elif lam_gt:
+            sens_ratio = np.inf
+        else:
+            sens_ratio = 0.0
+        if sens_ratio >= 2:  # arbitrary threshold
+            out["pushing up"].append(varkey)
+        elif sens_ratio <= 0.5:  # arbitrary threshold
+            out["pushing down"].append(varkey)
```
whoburg: Wondering whether comparing the sensitivity ratio is the right thing. Why not just a hard cutoff on sensitivity instead?

(could imagine entries in `lam` floating around in the numerical noise and crossing the ratio threshold, e.g. `1.5e-13` and `2.0e-14`)

gpkit/tools.py
whoburg: really?  

```python
In [16]: from gpkit import Variable, Model

In [17]: x = Variable('x')

In [18]: m = Model(x, [x >= 1])

In [19]: m.varkeys
Out[19]: {x: {x}}
```

bqpd: bound_all["varkeys"] is declared above, I think?

bqpd: Oh whoops you're right, this is what I get for just emailing.

bqpd: It's stable in Python 2, we use this technique in geometric_program too

gpkit/tools.py
whoburg: The sensitivity to the bounds in your unit test example is `1.00697474e-06` - and this decreases as the exponent `1.0000001` is moved closer to 1.

So, the necessary absolute threshold on sensitivity is in some sense a measure of how dual infeasible the model is -- how close that exponent is to making the model dual feasible.

I think as a practical matter setting a bound of `1e30` and looking for absolute position might be most reliable. It's also perhaps more intuitive for users to hear that a variable in their model is optimally near `1e30` -- that might be more understandable than hearing it exceeded the sensitivity threshold of `1e-6`, even if the latter is a stronger condition.

gpkit/tools.py
whoburg: hmph - ok for this pull request if it's used in GP too, but that sounds scary and potentially fragile to me.

whoburg:
initial review complete. I tend to think this could be even simpler than it is now -- just a dumb but reliable check on the value of the variables in the solution and whether they slid out to the bounds -- no sensitivities needed.

That would reduce the number of tolerance knobs and I think make this more intuitive.

?

gpkit/tools.py
bqpd: The problem with position is that sometimes the product of two numbers is
what's upper unbounded, and so they might not hit the boundary
individually, but are still sensitive to the upper bound.

whoburg:
I see the checkbox above - are we aiming to merge this and do that later, or get that done before merging? I'm fine with either, once the other review items are addressed/decided on.

bqpd:
I believe the current test is an example of a place where two of the
variables don't hit the bounds but are sensitive to them.

gpkit/tools.py
whoburg: huh really? Aren't all the bounds individual, an upper and a lower bound per variable? How can there be a sensitivity to the constraint `x <= 1e30` if x is not equal to `1e30`?

whoburg: hmm yeah, I see that on the test cvxopt doesn't push the variables all the way to the bounds. I think this is a tolerance issue, where cvxopt is calling the solution good enough. Mosek pushes all the way to the bounds.

I don't currently think it's a coupling issue or related to the product of two numbers being bounded.

But it does raise an interesting issue in how to determine that the bound has been hit.

bqpd:
@whoburg, I made the output dictionary clear as to what the entries mean. Ready for review.

docs/source/examples/unbounded.py
```diff
@@ -0,0 +1,20 @@
...
+mb = Variable("m_b", 0.4)
+V = Variable("V", 300)
+
+m = Model(F, [F >= D + V**2,
+              D == V**2*A,
+              V >= mi + mb,
+              Fs <= mi,
+             ])
+# sol = m.solve("mosek")  # returns UNKNOWN
+bounds = determine_unbounded_variables(m, "mosek")
```
whoburg: @bqpd, note that removing `"mosek"` from the line above now causes the unit tests to fail. In a fun twist, it seems that cvxopt returns unknown on the problem with all variables bounded (even though it solved the original problem).

whoburg:
@bqpd, review my latest commit to this branch, and then I think this is ready to merge

gpkit/tests/t_examples.py
```diff
@@ -76,6 +76,9 @@ def test_simpleflight(self, example):
             sol_rat = sol["sensitivities"]["constants"][key]/consenscheck[key]
             self.assertTrue(abs(1-sol_rat) < 1e-2)
 
+    def test_unbounded(self, example):
+        pass
```
whoburg: this gets run for all solvers currently -- we need better control over this -- see #258

docs/source/examples/unbounded.py
bqpd: D: maybe it needs a lower `eps`?

whoburg: changing it to 1e-20 (a 10^10 fold increase) does not fix it.

changing it to 1e-10 results in a different error,

```
FAIL: test_unbounded_debugging (gpkit.tests.t_model.TestSP_mosek_cli)
Test nearly-dual-feasible problems
----------------------------------------------------------------------
Traceback (most recent call last):
  File "gpkit/tests/t_model.py", line 360, in test_unbounded_debugging
    self.assertEqual(bounds["sensitive to lower bound"], [x.key])
AssertionError: Lists differ: [] != [x]

Second list contains 1 additional elements.
First extra element 0:
x

- []
+ [x]
?  +
```

bqpd: But does it fix cvxopt?

bqpd:
Do we want to make this a constraint set before merging? Also there's a missing file docstring for `unbounded.py`

docs/source/examples/unbounded.py
bqpd: Seems to work for me with `eps=1e-5`

docs/source/advancedcommands.rst
```diff
@@ -280,3 +280,23 @@ Example Usage
...
+resolving, and checking whether any variables slide off to the bounds.
+Mosek returns unknown when attempting to solve the following model:
+
+.. literalinclude:: examples/unbounded.py
+
+Upon viewing the output from ``determine_unbounded_variables``,
+
+.. literalinclude:: examples/unbounded_output.txt
+
+it becomes clear that ``A`` is unbounded below in the original model.
```
whoburg: add note here: in our initial testing, mosek appears to solve the bounded problem better for small eps.

Also add info on how to access `unbounded_variables` and use `BoundedConstraintSet`.

pgkirsch:
Is this ready to merge @bqpd?

bqpd:
I was thinking of turning it into a ConstraintSet first, sorry.

whoburg:
also just noticed that `determine_unbounded_variables` seems to call `localsolve` -- need to make it work with gps that need `solve`?

bqpd:
Yeah, that's why it should be a ConstraintSet. :p

mayork:
I get this error when running this line of code bounds = determine_unbounded_variables(design, "cvxopt") where design is a model

```Traceback (most recent call last):
  File "C:\Users\mayork\Documents\GitHub\gpkit-models\gpkitmodels\aircraft\engine_component_testing.py", line 182, in <module>
    bounds = determine_unbounded_variables(design, "cvxopt")
  File "C:\Python27\lib\site-packages\gpkit\tools.py", line 139, in determine_unbounded_variables
    sol = m.solve(solver, verbosity, **kwargs)
  File "C:\Python27\lib\site-packages\gpkit\constraints\prog_factories.py", line 79, in solvefn
    self.program, solvefn = genfunction(self, verbosity-1)
  File "C:\Python27\lib\site-packages\gpkit\constraints\prog_factories.py", line 34, in programify
    prog = program(self.cost, self, substitutions, verbosity)
  File "C:\Python27\lib\site-packages\gpkit\geometric_program.py", line 72, in __init__
    super(GeometricProgram, self).init_from_nomials(self.posynomials)
  File "C:\Python27\lib\site-packages\gpkit\nomials\data.py", line 78, in init_from_nomials
    exps = functools_reduce(add, (tuple(s.exps) for s in nomials))
  File "C:\Python27\lib\site-packages\gpkit\nomials\data.py", line 78, in <genexpr>
    exps = functools_reduce(add, (tuple(s.exps) for s in nomials))
AttributeError: 'numpy.float64' object has no attribute 'exps'```

**Comment by whoburg** [comment]
shall we try to rebase this onto master?

**Comment by mayork** [comment]
if this could be a priority that would be awesome, I think both would really help debug the engine code

**Comment by bqpd** [comment]
Sure thing! I'll rewrite it as a constraint set

**Comment by mayork** [comment]
I know ned's writing the gpkit paper right now so this could take a little bit, but is there a fast local hack I could do to get this running? If not its not a big deal

**Comment by whoburg** [comment]
I'll take a quick look at this. Getting on an airplane so response will be ~3 hours from now

**Comment by whoburg** [comment]
I'm going to force push a rebased version of this (with a bug fix) to this PR. I still have a local copy of the previous version in case @bqpd needs it.

**Comment by whoburg** [comment]
conflicts when rebasing were limited to import errors.

**Comment by whoburg** [comment]
@mayork, give this a try. I wasn't sure exactly what to run to reproduce the error you saw, so not sure yet whether or not this will fix it.

**Comment by whoburg** [comment]
Another todo before this is closed -- the example `docs/source/examples/unbounded.py` has an inline comment

`# sol = m.solve("mosek")  # returns UNKNOWN`

... but this isn't what happens on my machine. I get a solver cost did not match computed cost error from gpkit's solution checking. We should 

- [ ] figure out why this changed, and
- [ ] unit test that the inline comment is accurate

**Comment by whoburg** [comment]
The example should probably be changed to not depend on mosek -- this fails the windows machine that only has cvxopt installed.

**Comment by mayork** [comment]
@whoburg I think this fixed the error. I actually had that weird importing problem again but worked around it. Might be worth discussing that problem at some point

Also, I was thinking that before this is merged in to master there should be a way to handle signomial programs. In determine_unbounded_variables solve is called, maybe an if else to check if signomials are enabled then call localsolve?

**Comment by whoburg** [comment]
I'm going to rebase this again once #751 is merged -- that should allow me to clear up the windows unit test failure

**Comment by whoburg** [comment]
I'm going to rebase force push this again to get #751

**Comment by whoburg** [comment]
@galbramc curious to get your thoughts on how to deal with this. Look at my last commit to this PR and you'll see I'm just trying to run this test only if mosek is installed. We're now failing because the output changed. I think it has something to do with the fact that the CI script does `echo 'installed_solvers : cvxopt'`. Any thoughts on best way to make this work?

**Comment by whoburg** [comment]
I'm really confused about why all tests but windows succeeded here: https://acdl.mit.edu/csi/job/gpkit_PullRequest_Unit/1331/ -- but are now failing.

**Comment by galbramc** [comment]
Ahh, the ```echo 'installed_solvers : cvxopt'``` was a hack that @bqpd suggested we use so we would only test cvxopt even when both solvers are installed. I'm all for changing the hack if you have a better solution.

**Comment by galbramc** [comment]
In regards to 1331, I don't know why you get ```ImportError: Could not load MOSEK library: ImportError('No module named lib',)```. I don't think mosek is even installed on that VM...

**Comment by whoburg** [comment]
right, we got it in 1331 because `docs/source/examples/unbounded.py` explicitly called `solve("mosek")`. I tried to fix that in the last commit, but now we have this output changing issue.

I guess maybe we need to refactor the example so that it's solver-independent. that's probably the right solution.

**Comment by galbramc** [comment]
You know better than me how to fix that particular problem...

**Comment by whoburg** [comment]
yep -- sorry for the distraction

**Comment by galbramc** [comment]
No problem. Do you want me to change the script in regards to ```echo 'installed_solvers : cvxopt'```?

**Comment by whoburg** [comment]
Not yet. That should happen as part of the refactor proposed in #258. Maybe @aechevarria1 is interested.

**Comment by galbramc** [comment]
There is also an issue (I can't remember the number) with storing the settings file in the pip cache while you are at it.

**Comment by whoburg** [comment]
#679

**Comment by mayork** [comment]
I was talking to Woody one day and mentioned I once saw a model that solved when it wasn't bounded but didn't solve bounded....I just ran into something similar while doing some debugging. When bounded the model at commit https://github.com/hoburg/gpkit-models/pull/73/commits/891bded301f7a05625937297c88c66941aedeaea solves, unbounded it remains unsolved after 200 iterations. I'm not sure what's going on but I thought maybe one of y'all would want to look at it to understand better why this occurs

**Comment by galbramc** [comment]
Sounds like you should create an issue with enough information ( including an example that reproduces the problem ) so they can sort out the possible bug.

**Comment by whoburg** [comment]
Noticed using the boundedness debugging tool that it's not possible to access the solution of the bounded problem (since the method returns the boundedness dict and doesn't save the solution array).

Also, not sure if this is intended, but (on the new unit test test_unbounded_debugging, for example), after running `bounds = determine_unbounded_variables(m)` on a model `m`, `m.solution` is None.

Also, for an SP model, after a determine_unbounded_variables solve, `m.program` ends up being None as well, and there is no attribute `m.gps`.

Not sure if some of these are intentional design decisions, but currently it's difficult to query the values of variables after a boundedness solve.

**Comment by bqpd** [comment]
@whoburg, ready for review

**Comment by mayork** [comment]
I haven't switched over to this yet so I've been modifying a local copy of this...one thing I just did that seems potentially useful is to only lower bound or only upper bound variables. I think this is helpful for debugging...if y'all think this could be useful I can take a stab at adding a keyword to allow only upper or lower bounding once this is merged into master

**Comment by bqpd** [comment]
sounds great! Make an issue for it now so we remember?

File: docs/source/advancedcommands.rst [hunk 7]
```diff
@@ -233,3 +233,24 @@ Example Usage
...
+=========
+
+Unbounded variables
+-------------------
+In some cases a solver will return ``unknown`` when a free variable is
+unbounded by the model. ``gpkit.tools.determine_unbounded_variables`` is a
+simple tool that attempts to detect unbounded variables by adding extremely
+large upper bounds and extremely small lower bounds to all variables in a model,
+resolving, and checking whether any variables slide off to the bounds.
+Mosek returns unknown when attempting to solve the following model:
```
whoburg: this documentation should be changed to use `BoundedConstraintSet`

docs/source/examples/unbounded.py
```diff
@@ -0,0 +1,22 @@
...
+mb = Variable("m_b", 0.4)
+V = Variable("V", 300)
+
+constraints = [F >= D + V**2,
+               D == V**2*A,
+               V >= mi + mb,
+               Fs <= mi,
+               ]
+
+# Model(F, constraints).solve("mosek")  # returns UNKNOWN
```
whoburg: comment above is not true on my machine -- I get a `RuntimeWarning`

docs/source/examples/unbounded.py
```diff
@@ -0,0 +1,22 @@
...
+if "mosek" in settings["installed_solvers"]:
+    sol = Model(F, BoundedConstraintSet(constraints)).solve("mosek")
+    print sol["boundedness"]
```
whoburg: `docs/source/examples/unbounded_output.txt` appears to be an empty file on this branch, which isn't what we want (@bqpd maybe because you're running on a machine without mosek?)

gpkit/tests/t_model.py
```diff
@@ -405,6 +405,20 @@ def test_partial_sub_signomial(self):
...
 
+    def test_unbounded_debugging(self):
+        "Test nearly-dual-feasible problems"
+        from gpkit.tools import BoundedConstraintSet
+        x = Variable("x")
+        y = Variable("y")
+        m = Model(x*y, [x*y**1.000001 >= 100])
+        with self.assertRaises((RuntimeWarning, ValueError)):
+            m.solve(self.solver, verbosity=0)
+        m = Model(x*y, BoundedConstraintSet([x*y**1.000001 >= 100], verbosity=0))
```
whoburg: cleaner to use `m` or `m.constraints` instead of `[x*y**1.000001 >= 100]`?

gpkit/tools/tools.py
```diff
@@ -113,3 +115,63 @@ def mdmake(filename, make_tex=True):
         with open(filename+".tex.md", "w") as f:
             f.write(texmd)
     return open(filename+".py")
+
+
+def bound_all_variables(varkeys, lb, ub):
+    "Returns model with additional constraints bounding all free variables"
```
whoburg: looks like this returns a list of constraints, not a Model. Also I think the name could be better -- `bounded_variable_constraints` or `bounded_variables` or `get_bounds`?

gpkit/tools/tools.py
```diff
@@ -113,3 +115,63 @@ def mdmake(filename, make_tex=True):
...
+        constraints = [bounding_constraints, constraints]
+        super(BoundedConstraintSet, self).__init__(constraints, substitutions)
+
+    def sens_from_dual(self, las, nus):
+        "Return sensitivities while capturing the relevant lambdas"
+        self.bound_las = las[:2*len(self.bounded_varkeys)]
+        return super(BoundedConstraintSet, self).sens_from_dual(las, nus)
+
+    def process_result(self, result):
+        "Creates (and potentially prints) a dictionary of unbounded variables."
```
whoburg: description would be more precise if it said something like `"... a dictionary of variables affected by the bounds"` ?

pylint.sh
```diff
@@ -12,4 +12,4 @@ export PYTHONPATH=$PYTHONPATH:$WORKSPACE/gpkit/
 
 python $PYLINT --rcfile=$WORKSPACE/.pylintrc --disable=fixme $@ $WORKSPACE/gpkit/
 
-python $PYLINT --rcfile=$WORKSPACE/.pylintrc --disable=invalid-name,too-many-locals,redefined-outer-name,wrong-import-position $@ $WORKSPACE/docs/source/examples/*.py
+python $PYLINT --rcfile=$WORKSPACE/.pylintrc --disable=invalid-name,too-many-locals,redefined-outer-name,wrong-import-position,missing-docstring $@ $WORKSPACE/docs/source/examples/*.py
```
whoburg: I disagree with disabling this -- we're so close to having docstrings on all our methods, we should strive to do so.

There's a separate pylint threshold for allowing trivial methods shorter than a certain length to not have a dicstring.

whoburg:
code review complete.

pgkirsch:
Is there a path towards this getting merged?

mayork:
I wrote some documentation for this and am creating test files for the code in the documentation but will leave the part to run the tests commented out until this is merged

pylint.sh
bqpd: This isn't in the pylint rc file, I just wanted to stop seeing those when I ran the script :p

docs/source/advancedcommands.rst
bqpd: Ah, yes.

docs/source/examples/unbounded.py
bqpd: changing to "does not solve"

docs/source/examples/unbounded.py
bqpd: Indeed, good call!

gpkit/tests/t_model.py
bqpd: I like it.

gpkit/tools/tools.py
bqpd: Good call.

bqpd:
@whoburg, responded to code review and rebased on master

docs/source/advancedcommands.rst
bqpd: @whoburg, looks like this note never got added. Still desired?

bqpd:
retest this please (passed dependency but not unit tests???)

bqpd:
This is failing unit tests because of the mosek-only example.

bqpd:
retest this please

docs/source/advancedcommands.rst
```diff
@@ -233,3 +233,24 @@ Example Usage
...
 The ``sweep`` argument specifies what points between 0 and 1 you wish to sample the weights at. If you want different resolutions or spacings for different weights, the ``sweeps`` argument accepts a list of sweep arrays.
+
+
+Debugging
+=========
+
+Unbounded variables
+-------------------
+In some cases a solver will return ``unknown`` when a free variable is
+unbounded by the model. ``gpkit.constraints.bounded.BoundedConstraintSet`` is a
```
whoburg: are we going to add it to the `__init__` so that the above can be changed to `gpkit.constraints.BoundedConstraintSet`?

docs/source/advancedcommands.rst
```diff
@@ -233,3 +233,24 @@ Example Usage
...
+simple tool that attempts to detect unbounded variables by adding extremely
+large upper bounds and extremely small lower bounds to all variables in a ConstraintSet,
+resolving, and checking whether any variables slide off to the bounds.
```
whoburg: technically, the BoundedConstraintSet does not resolve; the above should say something more like "if a model involving a BoundedConstraintSet is resolved, warnings will be raised for any variables that slide off to the bounds."

docs/source/advancedcommands.rst
whoburg: First, we should check whether or not it's still true -- do cvxopt and mosek have the same behavior on `unbounded.py`?

docs/source/examples/unbounded.py
```diff
@@ -0,0 +1,14 @@
...
+x = Variable("x")
+
+constraints = [x >= 1]
+
+# Model(x, constraints).solve()  # does not solve
+m = Model(1/x, BoundedConstraintSet(constraints))
+# by default, prints bounds warning during solve
+sol = m.solve(verbosity=0)
+# bound waring is available in sol["boundedness"]:
+print sol["boundedness"]
```
whoburg: cut or comment print statement, so we don't get the output twice?

gpkit/constraints/bounded.py
```diff
@@ -0,0 +1,68 @@
...
+from ..small_scripts import mag
+
+
+def add_bounds(varkeys, lower, upper):
+    "Returns constraints list bounding all varkeys"
+    constraints = []
+    for varkey in varkeys:
+        units = varkey.descr.get("units", 1)
+        constraints.append([upper*units >= Variable(**varkey.descr),
+                            Variable(**varkey.descr) >= lower*units])
```
whoburg: Does it make sense to make `Variable(**varkey.descr)` once and then use it instead of making two Variable objects here?

gpkit/constraints/bounded.py
```diff
@@ -0,0 +1,68 @@
...
+    return constraints
+
+
+class BoundedConstraintSet(ConstraintSet):
+    "Bounds contained variables so as to ensure dual feasibility."
```
whoburg: add arg and kwarg descriptions to docstring

gpkit/constraints/bounded.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def __init__(self, constraints, substitutions=None, verbosity=1,
+                 eps=1e-30, lower=None, upper=None):
+        self.bound_las = None
+        self.verbosity = verbosity
+        self.lowerbound = lower if lower else eps
+        self.upperbound = upper if upper else 1/eps
+        constraintset = ConstraintSet([constraints])
```
whoburg: what if `constraints` is a `TightConstraintSet` or other child of `ConstraintSet`? Won't this demote it to a `ConstraintSet`?

Another potential way: should `ConstraintSet` have an `add_constraints` method, so this could call super init without casting `constraints` to `ConstraintSet`, and then add the bounding constraints?

Alternatively, if the cast is just a handy way to get the varkeys, then I think the line below 
```constraints = [bounding_constraints, constraintset]```
should use `constraints` in place of `constraintset`.

File: gpkit/constraints/bounded.py [hunk 20]
```diff
@@ -0,0 +1,68 @@
...
+                out["sensitive to lower bound"].append(varkey)
+            value = mag(result["variables"][varkey])
+            distance_below = np.log(value/self.lowerbound)
+            distance_above = np.log(self.upperbound/value)
+            if distance_below <= 3:  # arbitrary threshold
+                out["value near lower bound"].append(varkey)
+            elif distance_above <= 3:  # arbitrary threshold
+                out["value near upper bound"].append(varkey)
+        if self.verbosity > 0:
+            if out:
```
whoburg: `if self.verbosity > 0 and out:` ?

gpkit/tools/tools.py
```diff
@@ -2,7 +2,6 @@
 import numpy as np
 from ..nomials import Variable, VectorVariable
 from ..nomials import NomialArray
-# pylint: disable=duplicate-code
```
whoburg: nice!

whoburg:
@bqpd, re-review complete.

@mayork, your Jul 12 comment above mentions a commit that solved unbounded but not bounded. want to test again using the latest on this branch and open an issue if you're still seeing that behavior?

gpkit/constraints/bounded.py
bqpd: No, if constraints is a TightConstraintSet, we'll have `BCS(CS(TCS))`

docs/source/advancedcommands.rst
bqpd: We haven't done that with `TightConstraintSet`, so I was following that pattern.

docs/source/examples/unbounded.py
bqpd: Sounds good. Also noticed a typo.

docs/source/advancedcommands.rst
bqpd: Done.

gpkit/constraints/bounded.py
bqpd: Some, for sure! Done.

gpkit/constraints/bounded.py
bqpd: Done.

gpkit/constraints/bounded.py
bqpd: Done.

bqpd:
@whoburg, responded to code review.

gpkit/constraints/bounded.py
whoburg: ok got it, thanks for clarifying.

docs/source/advancedcommands.rst
whoburg: ok sounds good

docs/source/advancedcommands.rst
```diff
@@ -233,3 +233,26 @@ Example Usage
...
+Unbounded variables
+-------------------
+In some cases a solver will return ``unknown`` when a free variable is
+unbounded by the model. ``gpkit.constraints.bounded.BoundedConstraintSet`` is a
+simple tool that attempts to detect unbounded variables by adding extremely
+large upper bounds and extremely small lower bounds to all variables in a ConstraintSet.
+
+When a model if with a BoundedConstraintSet is solved, it checks whether any variables slid off to the bounds, notes this in the solution dictionary and prints a warning (if verbosity is greater than 0).
+
+For example, Mosek returns unknown when attempting to solve the following model:
```
whoburg: Not true, for me mosek raises a `RuntimeWarning: final status of solver 'mosek' was 'DUAL_INFEAS_CER', not 'optimal'.`.

docs/source/examples/unbounded.py
```diff
@@ -0,0 +1,14 @@
+"Demonstrate a trivial unbounded variable"
+from gpkit import Variable, Model
+from gpkit.constraints.bounded import BoundedConstraintSet
+
+x = Variable("x")
+
+constraints = [x >= 1]
+
+# Model(x, constraints).solve()  # does not solve
```
whoburg: `x` should be `1/x` above

gpkit/constraints/bounded.py
```diff
@@ -0,0 +1,103 @@
...
+
+    eps : float
+        default lower bound is eps, upper bound is 1/eps
+
+    lower : float
+        lower bound for all varkeys, replaces eps
+
+    upper : float
+        upper bound for all varkeys, replaces 1/eps
+    """
```
whoburg: 👍 👍

whoburg:
Almost there! Re-review found two final issues to address before merging.

docs/source/advancedcommands.rst
whoburg: I think the above should mention dual infeasibility -- in particular,

 - A model that is DUAL_INFEAS_CER is always a good candidate for debugging with a BoundedConstraintSet
 - A model that is UNKNOWN is sometimes a possible to debug with a BoundedConstraintSet -- when the unknown is actually caused by dual infeasibility undetected by the solver.

mjburton11:
I got this error while trying to build a LinkedConstraintSet on this branch.  May be relevant. 

```python
/Users/mjburton11/Documents/SuperUROP/gpkit/gpkit/constraints/set.pyc in subinplace(self, subs, value)
    126         "Substitutes in place."
    127         for constraint in self:
--> 128             constraint.subinplace(subs, value)
    129 
    130     @property

/Users/mjburton11/Documents/SuperUROP/gpkit/gpkit/constraints/set.pyc in subinplace(self, subs, value)
    126         "Substitutes in place."
    127         for constraint in self:
--> 128             constraint.subinplace(subs, value)
    129 
    130     @property

AttributeError: 'bool' object has no attribute 'subinplace'
```

docs/source/advancedcommands.rst
bqpd: Done.

docs/source/examples/unbounded.py
bqpd: Whoops.

bqpd:
Well, it looks like there's a bool in your model, potentially a posynomial equality? I don't think that's related to boundedconstraintset, though...

mjburton11:
I just double checked my model and I didn't see a boolean or posynomial equality.  Maybe we can look at it when we catch up this afternoon.

bqpd:
retest this please

bqpd:
@whoburg, ready for rereview

whoburg:
Was there a resolution of where @mjburton11's bool came from?

Assuming so, I'm ready to merge.

whoburg:
oh, add this to the notes here: https://github.com/hoburg/gpkit/releases/tag/untagged-94bb89ada8a10826054e

bqpd:
retest this please

PR #580, opened 2016-03-15
simplify gp after substitution

### LINKED ISSUE #525: GP not simplified post-substitution
Here is a simple GP (the fact that it has unbounded variables is irrelevant for this issue):

```python

D  = Variable('D')
mi = Variable('m_i')
V  = Variable('V', 1)

m = Model(D + V,
          [V >= mi + 0.4])
#sol = m.solve("mosek")
print m.gp().A.todense()
print m.gp().varlocs
```
The output from the print statements above is:
```
D has no lower bound
m_i has no lower bound
[[0 0]
 [0 1]
 [0 0]
 [1 0]]
D has no lower bound
m_i has no lower bound
{m_i: [3], D: [1]}
```

Note that the `value` of `V` is `1`. Here is the same GP with `1` manually substituted for `V`:
```python

D  = Variable('D')
mi = Variable('m_i')
V  = Variable('V', 1)

m = Model(D + 1,
          [1 >= mi + 0.4])
#sol = m.solve("mosek")
print m.gp().A.todense()
print m.gp().varlocs
```
But... the output from the print statements is now completely different:
```
D has no lower bound
m_i has no lower bound
[[0 0]
 [0 1]
 [1 0]]
D has no lower bound
m_i has no lower bound
{m_i: [2], D: [1]}
```
The fact that gpkit's substitution fails to produce the same result as a manual substitution means that the solver is passed a different problem. Indeed for the MWE's above, mosek returns `UNKNOWN` for the first (`V` has value=1) version, and has an issue #361 error for the second (manually-substituted) version.

bqpd:
closes #525

### DIFF
--- a/gpkit/nomials/nomial_math.py
+++ b/gpkit/nomials/nomial_math.py
@@ -428,6 +428,23 @@ class ScalarSingleEquationConstraint(SingleEquationConstraint):
         self.varkeys.update(self.right.varlocs)
 
 
+def simplify_posy_ineq(exps, cs):
+    "Simplify a posy <= 1 posynomial by moving constants to the right side."
+    coeff = 1.0
+    count = 0
+    exps_ = list(exps)
+    for i, exp in enumerate(exps):
+        if not exp:
+            coeff -= mag(cs[i-count])
+            cs = np.hstack((cs[:i-count], cs[i+1-count:]))
+            exps_.pop(i-count)
+            count += 1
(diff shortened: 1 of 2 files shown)


galbramc:
test this please

whoburg:
note rebased and force pushed -- update any local copies of this branch

bqpd:
Ready to merge, once it passes tests!

PR #581, opened 2016-03-16
remove old docstring info, unneeded pylint suppression

whoburg:
minor cleanups...

ready for review.

### DIFF
--- a/gpkit/__init__.py
+++ b/gpkit/__init__.py
@@ -69,7 +69,6 @@ def disable_units():
 
     class DummyUnits(object):
         "Dummy class to replace missing pint"
-        # pylint: disable=too-few-public-methods
         class Quantity(object):
             "Dummy Quantity instead of pint"
             pass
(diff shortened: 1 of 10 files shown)


bqpd:
Looks good to me.

whoburg:
@galbramc, it looks like this is failing due to recently added git status check

galbramc:
Indeed it is. So do you consider this a real or fake error. From what you told me earlier this means there is some sort of error with the examples. I have not added the test to gpki_Push, but the error is in the master branch so all PRs will fail with this test.

galbramc:
What order are the solvers tested? I'm guessing it's either the mosek interface or mosek_ci. That is falling. I added the check to depends as well, but it looks like the last solver is overwriting the incorrect results with correct ones.

whoburg:
Oh, yeah, that's the problem -- we generate the examples output with a specific solver (cvxopt). So the git status check should happen on the cvxopt machines only.

the ouput can be different for mosek due to numerical precision.

galbramc:
test this please

Only the cvxopt solver now checks for modified files

galbramc:
Sines this is a cleanup PR, maybe the handful of pylint warnings in the gpkit/test folder could be fixed? Suppress or move import statement warnings?

galbramc:
Nice. Fixed another 10 warnings!

Of course, pylint will fail for all other PRs until this is merged...

whoburg:
haha. ok - @bqpd was happy, so merging.

PR #582, opened 2016-03-16
faster simplify_posy_ineq, add test

whoburg:
@bqpd, I think this implementation is a bit faster -- it should be O(len(cs)) instead of O(len(cs)*len(0exps)) for large exps and cs (if we ever encounter them)

also added a test


bqpd:
:clap:  I'm down to merge this. Want to rebase `walden` afterwards?

bqpd:
Note that there was already a very similar test, although not this exact one.

whoburg:
ok, I'll merge this and rebase.

PR #591, opened 2016-03-17
Tight Constraint Set

pgkirsch:
Implements a tight constraint set that checks whether the left and right side of an equality are equal after substituting in the solved values.

@whoburg ready for review


bqpd:
retest this please

PR #604, opened 2016-03-25
add error for tautological constraints

### LINKED ISSUE #601: equality constraints with no variables
This is a carry-over from a behavior noticed by @pgkirsch in https://github.com/hoburg/gpkit-models/pull/45. MWE:

```python
x = Variable("x")
y = Variable("y", 2)
m = Model(x,
          [x >= 1,
           y == 2])
m.solve()
```
Note that after substitution, the second constraint becomes `2 == 2`. This currently results in a `ValueError: infeasible constraint: constant term too large.`.

This is confusing because the constraint is actually feasible. Can we raise a more informative error?

bqpd:
Closes #601

### DIFF
--- a/gpkit/nomials/nomial_math.py
+++ b/gpkit/nomials/nomial_math.py
@@ -463,11 +463,12 @@ class PosynomialInequality(ScalarSingleEquationConstraint):
             else:
                 coeff -= mag(cs[i])
         if len(exps_) < len(exps):
-            if coeff < 0:
+            if coeff > 0:
+                cs = cs[nonzero_exp_ixs]
+            elif coeff < 0:
                 raise ValueError("infeasible constraint: %s" % self)
             elif coeff == 0:
                 raise ValueError("tautological constraint: %s" % self)
-            cs = cs[nonzero_exp_ixs]
         return tuple(exps_), cs/coeff
 
     def _gen_unsubbed(self):
--- a/gpkit/nomials/nomial_math.py
+++ b/gpkit/nomials/nomial_math.py
@@ -451,7 +451,7 @@ class PosynomialInequality(ScalarSingleEquationConstraint):
         self.nomials = [self.left, self.right, self.p_lt, self.m_gt]
         self.nomials.extend(self._unsubbed)
 
-    def simplify_posy_ineq(self, exps, cs):
+    def _simplify_posy_ineq(self, exps, cs):
         "Simplify a posy <= 1 by moving constants to the right side."
         coeff = 1.0
         exps_ = []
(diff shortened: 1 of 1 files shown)


gpkit/nomials/nomial_math.py
```diff
@@ -469,6 +451,25 @@ def __init__(self, left, oper, right):
         self.nomials = [self.left, self.right, self.p_lt, self.m_gt]
         self.nomials.extend(self._unsubbed)
 
+    def simplify_posy_ineq(self, exps, cs):
+        "Simplify a posy <= 1 by moving constants to the right side."
```
whoburg: I find it confusing that this is a method of `self` but also takes `exps` and `cs` as arguments. I see that it was needed for printing, but, can exps and cs be anything here? Or do they need to relate to self in some way?

gpkit/nomials/nomial_math.py
```diff
@@ -469,6 +451,25 @@ def __init__(self, left, oper, right):
...
+                nonzero_exp_ixs.append(i)
+                exps_.append(exp)
+            else:
+                coeff -= mag(cs[i])
+        if len(exps_) < len(exps):
+            if coeff < 0:
+                raise ValueError("infeasible constraint: %s" % self)
+            elif coeff == 0:
+                raise ValueError("tautological constraint: %s" % self)
+            cs = cs[nonzero_exp_ixs]
```
whoburg: this is probably silly, but would it be slightly faster to do a `if coeff > 0` case first, for the standard case, and then move into the cases that raise ValueErrors?

bqpd: Yes! At least a microsecond per posynomial faster.

gpkit/nomials/nomial_math.py
bqpd: They can be anything, it's really only for printing.

whoburg: But I'm still confused. What is required such that what is printed for `self` in the error is in any way consistent with the arguments that were passed in?

bqpd: Right now it's just how it's used. My thought was that it would be helpful for the user to know just which constraint was triggering the error.

whoburg: Agree with that.

The confusion is that this method only makes sense when called with the `exps` and `cs` that arise in `self._gen_unsubbed` and `self.as_posyslt1`. maybe it should be private? If it's part of the user-facing API then I want to know what are valid inputs, and it doesn't seem to make sense to print `self` in the errors for arbitrary inputs.

bqpd: Oh! Yeah, you're right, it should be private.

bqpd: Done!

gpkit/nomials/nomial_math.py
bqpd: Done.

bqpd:
@whoburg, I addressed the code review. Is this ready to merge?

whoburg:
retest this please

PR #631, opened 2016-04-05
print warning about unused substitutions

### LINKED ISSUE #609: Checking whether a substitution key is actually in varkeys
There have been a couple times now where I add a key:value pair to my substitution dict, but I accidentally misspell the varkey or forget an underscore or something, and then the problem solves (without the substitution I want) and I go on, blissfully unaware. One (hopefully easy) check that might be nice and would help to avoid silly user mistakes is something that ensures that all of your substitution keys are actually valid varkeys.

bqpd:
Closes #609

### DIFF
--- a/gpkit/geometric_program.py
+++ b/gpkit/geometric_program.py
@@ -52,6 +52,13 @@ class GeometricProgram(NomialData):
         ## Create list of substituted posynomials <= 1
         self.posynomials = [cost.sub(self.substitutions)]
         self.constr_idxs = []
+        self.unusedsubkeys = set(key for key in self.substitutions
+                                 if (key not in constraints.varkeys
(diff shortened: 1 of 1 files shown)


gpkit/geometric_program.py
```diff
@@ -52,6 +52,12 @@ def __init__(self, cost, constraints, substitutions=None, verbosity=1):
         ## Create list of substituted posynomials <= 1
         self.posynomials = [cost.sub(self.substitutions)]
         self.constr_idxs = []
+        self.unusedsubkeys = set(key for key in self.substitutions
+                                 if key not in constraints.varkeys)
```
whoburg: `self.unusedsubkeys = set(self.substitutions) - set(constraint.varkeys)`?

PR #638, opened 2016-04-07
don't create meaningless sensitivities, closes #450

### LINKED ISSUE #450: sens only gives variable sensitivities, but accepts signomial input
Is doing so even possible? I am skeptical.

bqpd:
@whoburg, ready for review

### DIFF
--- a/gpkit/solution_array.py
+++ b/gpkit/solution_array.py
@@ -1,7 +1,7 @@
 """Defines SolutionArray class"""
 from collections import Iterable
 import numpy as np
-from .nomials import NomialArray, Monomial
+from .nomials import NomialArray
 from .small_classes import Strings, DictOfLists
 from .small_scripts import unitstr, mag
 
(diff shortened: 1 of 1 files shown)


gpkit/solution_array.py
```diff
@@ -80,17 +80,7 @@ def sens(self, nomial):
```
whoburg: should the docstring be updated? What are the valid input types for `nomial`? Variable? Monomial? Signomial?

bqpd: Anything that's in the keymap or has a .key attribute.

whoburg: (I'm referring to the `sens` method -- looks like this inline comment is showing up in an unexpected place)

Then should `sens(self, nomial)` change to `sens(self, key)` and the docstring be updated to remove the part about signomial input?

whoburg:
@bqpd, this is waiting on your reply to the docstring comment/question.

bqpd:
Responded to code review!

bqpd:
@whoburg please merge this if it passes your review, it's an issue in Next Release

PR #639, opened 2016-04-07
add user-friendly ConstraintSet class for inheriting from, closes #607

### LINKED ISSUE #607: Document when to inherit from ConstraintSet vs Model
Re https://github.com/hoburg/gpkit-models/pull/45#issuecomment-201454600

bqpd:
@whoburg, ready for review.

### DIFF
--- a/gpkit/__init__.py
+++ b/gpkit/__init__.py
@@ -119,6 +119,7 @@ from .nomials import Monomial, Posynomial, Signomial
 from .nomials import Variable, VectorVariable, ArrayVariable
 from .geometric_program import GeometricProgram
 from .constraints.signomial_program import SignomialProgram
+from .constraints.set import ConstraintSet
 from .constraints.model import Model
 from .constraints.linked import LinkedConstraintSet
 
--- a/gpkit/constraints/set.py
+++ b/gpkit/constraints/set.py
@@ -26,6 +26,17 @@ class ConstraintSet(list):
         # initializations for attributes used elsewhere
         self.posymap = []
 
+    def __getitem__(self, key):
+        if isinstance(key, int):
+            return list.__getitem__(self, key)
+        else:
+            from ..nomials import Variable
+            variables = [Variable(**key.descr) for key in self.varkeys[key]]
+            if len(variables) == 1:
+                return variables[0]
+            else:
+                return variables
+
     __str__ = _str
     __repr__ = _repr
     _repr_latex_ = _repr_latex_
--- a/gpkit/constraints/costed.py
+++ b/gpkit/constraints/costed.py
@@ -1,6 +1,5 @@
 "Implement CostedConstraintSet"
 from .set import ConstraintSet
-from ..nomials import Variable
 
 
 class CostedConstraintSet(ConstraintSet):
@@ -19,16 +18,6 @@ class CostedConstraintSet(ConstraintSet):
         if substitutions:
             subs.update(substitutions)
         ConstraintSet.__init__(self, constraints, subs)
-
-    def __getitem__(self, key):
-        if isinstance(key, int):
-            return list.__getitem__(self, key)
-        else:
-            variables = [Variable(**key.descr) for key in self.varkeys[key]]
-            if len(variables) == 1:
-                return variables[0]
-            else:
-                return variables
 
     def subinplace(self, subs, value=None):
         "Substitutes in place."


bqpd:
retest this please

bqpd:
@galbramc: says it failed dependencies but there's no failure in the jenkins output.

galbramc:
when did I say what failed where?

bqpd:
oops, sorry! _Jenkins_ says it failed dependencies but there's no failure in the parsed console output.

galbramc:
The parsed console is only looking for a set of regular expressions which I have specified in a file for Jenkins. Turns out pip barfed with an error I have not seen before, so it was not in the list of regular expressions. You should check the regular console when this thing happens.

I'll see if I can add a regular expression to catch this error

gpkit/constraints/set.py
```diff
@@ -1,222 +1,24 @@
...
-        for constraint in self:
-            if not isinstance(constraint, ConstraintSet):
-                yield constraint
+    def __getitem__(self, key):
+        if isinstance(key, int):
+            return list.__getitem__(self, key)
+        else:
+            variables = [Variable(**key.descr) for key in self.varkeys[key]]
+            if len(variables) == 1:
+                return variables[0]
```
whoburg: the only user-facing method is `__getitem__`? Why not just move this to `Constraint` and rename `Constraint` to `ConstraintSet`? That feels conceptually clean and I would like that naming a lot.

gpkit/__init__.py
```diff
@@ -120,7 +120,10 @@ def __exit__(self, type_, val, traceback):
 from .geometric_program import GeometricProgram
 from .constraints.signomial_program import SignomialProgram
 from .constraints.model import Model
+from .constraints.set import ConstraintSet
+from .constraints.costed import CostedConstraintSet
```
whoburg: not sure `CostedConstraintSet` should be in `__init__.py`. Is the guidance for users to use `Model`, or `CostedConstraintSet`? The name `CostedConstraintSet` feels a bit awkward for user consumption.

gpkit/constraints/base.py
```diff
@@ -0,0 +1,222 @@
+"Implements Constraint"
+from ..small_classes import HashVector
+from ..keydict import KeySet, KeyDict
+from ..small_scripts import try_str_without
+from ..repr_conventions import _str, _repr, _repr_latex_
+
+
+class Constraint(list):
```
whoburg: The name `Constraint` seems a bit too general/broad. Below is a list of names I like better, in rough order of preference. Do you like any of these?

`RecursiveConstraintContainer`
`ConstraintContainer`
`BaseConstraint` (potentially confusing because 'base' sounds like base case)

... and now I'm wondering: why can't we just combine this with `ConstraintSet`? (sorry if this is a repeat question)

gpkit/constraints/set.py
```diff
@@ -1,222 +1,24 @@
-"Implements ConstraintSet"
-from ..small_classes import HashVector
-from ..keydict import KeySet, KeyDict
-from ..small_scripts import try_str_without
-from ..repr_conventions import _str, _repr, _repr_latex_
+"Implement CostedConstraintSet"
```
whoburg: Shouldn't the above line say `ConstraintSet`, not `CostedConstraintSet`?

whoburg:
review complete.

gpkit/constraints/set.py
bqpd: Because `__getitem__` requires Variable which requires Monomial which requires Constraint; it's another circular loop.

gpkit/constraints/set.py
bqpd: Oops, yup.

gpkit/constraints/base.py
bqpd: Yeah, I'm not a huge fan of the name, though hopefully an end user would never have to see it...

gpkit/__init__.py
bqpd: Good call.

gpkit/constraints/set.py
whoburg: ah. hmm, there has to be a good way. let me stew on this

bqpd: (also, the user-facing `setup` method should be added as well)

whoburg: Does this need to return a list of `Variable`s instead of a list of varkeys? It seems returning `self.varkeys[key]` would solve the circular dependency issue.

whoburg: I thought we killed `setup` methods in favor of `Model.__init__`?

bqpd: Oh, sorry, not `setup`; the method that renames all the varkeys in a model.

bqpd: Yeah, this could return varkeys, but the way everyone wants to use its results is as variables.

bqpd:
@whoburg  ready for review

Issue #633, opened 2016-04-06
don't pass the solver tautological monomial constraints?


bqpd:
per #601

bqpd:
Closed by the commit referenced above.

PR #640, opened 2016-04-07
skip tautological monomial constraints, closing #633

bqpd:
@whoburg, waiting on review.

### DIFF
--- a/gpkit/nomials/nomial_math.py
+++ b/gpkit/nomials/nomial_math.py
@@ -476,7 +476,7 @@ class PosynomialInequality(ScalarSingleEquationConstraint):
                 else:
                     # allow tautological monomial constraints (cs[0] <= 1)
                     # because they allow models to impose requirements
-                    return [], []
+                    return (), np.array([])
             return exps, cs
         coeff = 1.0
         exps_ = []
--- a/gpkit/nomials/nomial_math.py
+++ b/gpkit/nomials/nomial_math.py
@@ -473,8 +473,10 @@ class PosynomialInequality(ScalarSingleEquationConstraint):
             if not exps[0]:
                 if cs[0] > 1:
                     raise ValueError("infeasible constraint: %s" % self)
-            # for now, we allow tautological monomial constraints (cs[0] <= 1)
-            # because they allow models to impose requirements on variables
+                else:
+                    # allow tautological monomial constraints (cs[0] <= 1)
+                    # because they allow models to impose requirements
+                    return [], []
             return exps, cs
         coeff = 1.0
         exps_ = []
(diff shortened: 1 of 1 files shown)


gpkit/nomials/nomial_math.py
```diff
@@ -473,8 +473,10 @@ def _simplify_posy_ineq(self, exps, cs):
             if not exps[0]:
                 if cs[0] > 1:
                     raise ValueError("infeasible constraint: %s" % self)
-            # for now, we allow tautological monomial constraints (cs[0] <= 1)
-            # because they allow models to impose requirements on variables
+                else:
+                    # allow tautological monomial constraints (cs[0] <= 1)
+                    # because they allow models to impose requirements
+                    return [], []
```
whoburg: should it return `(), ()` to match the tuple return type below?

gpkit/nomials/nomial_math.py
```diff
@@ -523,9 +525,11 @@ def as_posyslt1(self):
             # remove any cs that are just nans and/or 0s
             nans = np.isnan(cs)
             if np.all(nans) or np.all(cs[~nans] == 0):
-                return []  # skip nan'd or 0'd constraint
+                continue  # skip nan'd or 0'd constraint
```
whoburg: was this a bug catch?

whoburg:
looks great. See one inline comment; make a change if you agree, and merge either way.

gpkit/nomials/nomial_math.py
bqpd: Yup.

gpkit/nomials/nomial_math.py
bqpd: Hmm, it's really tuple/ndarray

gpkit/nomials/nomial_math.py
whoburg: :+1:

gpkit/nomials/nomial_math.py
whoburg: I guess so, if they're not expensive to create. I think array not ndarray though -- I for one don't know how to create an empty ndarray:

```python
In [1]: import numpy as np

In [2]: np.ndarray()
TypeError                                 Traceback (most recent call last)
<ipython-input-2-98b15771a5de> in <module>()
----> 1 np.ndarray()

TypeError: Required argument 'shape' (pos 1) not found

In [3]: np.ndarray([])
Out[3]: array(9.223372036854776e+18)

In [4]: np.array([])
Out[4]: array([], dtype=float64)
```

whoburg:
#633 is closed but this hasn't been merged yet -- should we merge it?

bqpd:
Oh, derp. I think so!

bqpd:
Responded to code review, ready to merge.

PR #642, opened 2016-04-07
add tight and linked constraint errors

### LINKED ISSUE #629: Create new error classes
A couple of times now I'm seeing `ValueErrors` caught when they shouldn't be because we use that type of error too much in the code. We should create new errors and use them instead!

bqpd:
preliminary, will close #629

### DIFF
--- a/gpkit/constraints/linked.py
+++ b/gpkit/constraints/linked.py
@@ -4,6 +4,11 @@ from ..varkey import VarKey
 from .. import SignomialsEnabled
 
 
+class SubstitutionConflictError(Exception):
+    "Raised when linked variables have differing values."
+    pass
+
+
 class LinkedConstraintSet(ConstraintSet):
     """A ConstraintSet that links duplicate variables in its constraints
 
@@ -25,6 +30,7 @@ class LinkedConstraintSet(ConstraintSet):
     exclude: set
         blacklist of variable names, supercedes include_only
     """
+    # pylint: disable=too-many-locals
     def __init__(self, constraints, include_only=None, exclude=None):
         ConstraintSet.__init__(self, constraints)
         varkeys = self.varkeys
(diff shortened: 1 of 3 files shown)


gpkit/constraints/tight.py
```diff
@@ -3,6 +3,11 @@
 from ..nomials import PosynomialInequality, SignomialInequality
 
 
+class TightnessError(Exception):
+    "Raised when linked variables have differing values."
+    pass
```
whoburg: should it inherit from `ValueError`?

gpkit/constraints/linked.py
```diff
@@ -4,6 +4,11 @@
 from .. import SignomialsEnabled
 
 
+class SubstitutionConflictError(Exception):
+    "Raised when linked variables have differing values."
+    pass
```
whoburg: maybe just `SubstitutionError`?

Or, if this is not referring to gpkit.substitution, then `LinkingError`?

gpkit/constraints/tight.py
bqpd: Maybe! The disadvantage of that would be that catching ValueError (which we/cvxopt sometimes use to indicate e.g. solver failure) would also catch this kind of error; being able to tell the difference between those two types of error would be useful.

gpkit/constraints/linked.py
bqpd: I kind of like having the specific problem in the error name; do you think it's unclear?

gpkit/constraints/tight.py
whoburg: right. so I guess the high-level design question is whether

1. we write special error types that inherit from standard types for us to catch internally in special circumstances. Advantage: external code can more easily discriminate errors without knowing our special types, thanks to inheritance. Disadvantage: these might inadvertently be cause by other code catching standard types.

2. We write special error types to avoid detection by things catching errors. Advantage: easy to implement internally. Disadvantage: doesn't play nicely externally.

gpkit/constraints/linked.py
whoburg: No it's fine as is

gpkit/constraints/tight.py
bqpd: I think the general MO is to create new types of errors for things that a user might want to catch alone, and feel that TightnessErrors and SubstitutionConflicts are both good examples of that, along with Signomials in GPs and lack of in SPs, etc.

bqpd:
Closing this, but the idea remains in #629

PR #652, opened 2016-04-13
Fixes in build.py for Windows

galbramc:
This should fix the builds on the Windows 10 VM


galbramc:
test this please

bqpd:
Oh, good catch! It seems to still not be working?

galbramc:
test this please

galbramc:
It's working on windows 10 now, but windows 7 is broken because I followed the recommended install process where python(x,y) is the prefered version of python. Python(x,y) is only 32 bit, so mosek must be 32 bit, but the build script is not setup to find the 32 bit mosek. I'm trying to fix that now.

galbramc:
@bqpd @whoburg all testing on the windows machines will fail until this is merged. Please review and adjust to your liking.

We will also need to reset the pylint warnings, as this is the last file with a bazillion of them...

bqpd:
@whoburg, I'm ready to merge this.

whoburg:
sounds good. @galbramc and I are going to fix some pylint errors before merging

PR #654, opened 2016-04-13
update install instructions with recommendations from @galbramc

whoburg:
This is ready for review and could use some work from someone with a good eye for graphics / markdown.

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -1,103 +1,140 @@
 Installation Instructions
 *************************
 
-If you encounter any bugs during installation, please email ``gpkit@mit.edu``.
+If you encounter any bugs during installation, please email ``gpkit@mit.edu``,
+or raise a `new issue <http://github.com/hoburg/gpkit/issues/new>`_.
 
-Mac OS X
-========
+Install dependencies
+====================
+GPkit's dependencies are as follows.
+ - ``pip``,
+ - ``numpy`` version 1.8 or newer,
+ - ``scipy``,
+ - ``pint``,
 
-1. Install Python and build dependencies
-++++++++++++++++++++++++++++++++++++++++
-  - Install the Python 2.7 version of `Anaconda <http://continuum.io/downloads>`_.
-    - Check that Anaconda is installed: in a Terminal window, run ``python`` and check that the version string it prints while starting includes "Anaconda".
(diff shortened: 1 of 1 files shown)


galbramc:
I wonder if you should mention something about installing in your system vs as a user. Technically you need to run "sudo pip install" on OS X and Linux to install things into the system. Otherwise you should run "pip install --user" to just install for a user. Not sure what the easiest way is to document this.

galbramc:
I wonder if you should mention something about installing in your system vs as a user. Technically you need to run "sudo pip install" on OS X and Linux to install things into the system. Otherwise you should run "pip install --user" to just install for a user. Not sure what the easiest way is to document this.

galbramc:
I think we need to update the "unintstall" instructions, or instructions when you install a new solver. Currently the settings file gets saved in the cache, that means you only get the solvers you have the first time you install gpkit. Any subsequent times you run pip it will simply restore the cached settings file, even if you have installed other solvers since then. I think we should tell people to blow away the cache:

Linux
find ~/.cache/pip/wheels/ -name "*gpkit*" -delete

OS X
find ~/Library/Caches/pip/wheels -name "*gpkit*" -delete

Windows
C:\Users\(your_user_name)\AppData\Local\pip\cache\wheels
Search for gpkit and delete any files with the name gpkit*.whl

galbramc:
test this please

whoburg:
@galbramc thanks -- are you sure the cache needs to be deleted even with the `--no-cache-dir` option for `pip install`?

galbramc:
Well, you can either tell people to delete the cache, or recommend that they use ```--no-cache-dir``` absolutely every time they install gpkit. This is really a hole in pip that there is no easy way to force a "rebuild".

whoburg:
OK thanks -- we think our "rebuild GPkit" instructions are correct since they list --no-cache-dir

galbramc:
Yes, but it can be deceptive

install cvxopt
install gpkit
uninstall gpkit
install mosek
install gpkit  --no-cache-dir
unisntall gpkit
install gpkit      <-- only gets cvxopt because that's what was in the cache

the last install lost mosek because gpkit was restored from the cache

galbramc:
I think the right thing to do really is the blow away the cache. It's the most robust solution.

galbramc:
Though it's the most awkward solution.

galbramc:
Maybe if you put the settings file in .gpkit rather than gpkit/env. That might solve the problem. Then suggesting --no-cache-dir would work correctly.

PR #656, opened 2016-04-15
(in progress) SignomialEquality constraints and methods

whoburg:
This is a PR in progress for @mopg and bscohen to work on signomial programming algorithms -- with a particular focus on dealing with equality constraints.

@bqpd, I hacked together an initial prototype but this could use your eyes -- suggestions welcome on how to do this right.

The code we used this morning to test was

```python
from gpkit import Variable, Model, SignomialsEnabled
x = Variable("x")
y = Variable("y")
ymin = Variable("y_{min}", 4)

with SignomialsEnabled():
    c = SignomialEquality(x, y + 1)
    
m = Model(x, [c,
              y >= ymin])

#m.localsolve(x0={"x": 1, y: 1})
sol = m.localsolve()
```


whoburg:
@bqpd, what's the best way to look at the convergence of the SP iterations? There are (surprisingly?) None Types mixed in with the gps:
`[type(gp.result) for gp in sol.program.gps]`
returns
`[NoneType, dict, NoneType, dict, NoneType, dict, dict, dict]`

galbramc:
test this please

galbramc:
test this please

gpkit/nomials/nomial_math.py
```diff
@@ -698,5 +698,101 @@ def sens_from_gpconstr(self, posyapprox, pa_sens, var_senss):
...
+# TODO do this right
+#         constr_sens = dict(pa_sens)
+#         del constr_sens[str(posyapprox.m_gt)]
+#         _, negy = self._unsubbed.posy_negy()
+#         constr_sens[str(negy)] = pa_sens["overall"]
+#         pa_sens[str(posyapprox)] = pa_sens.pop("overall")
+#         constr_sens["posyapprox"] = pa_sens
+#         return constr_sens
+
+
```
whoburg: hey @bqpd, in the implementation above,  `SignomialEqualityTriv` and  `SignomialEqualityLin` are two different versions of a signomial equality constraint -- the former solves by monomializing the negynomial only; the latter solves by monomializing both the posynomial and negynomial sides.

From a modeler's perspective, I might not know which method I want to solve with when I write down my problem, so it might make more sense to have a single `SignomialEquality` class with different methods of producing `as_gpconstr`.

Any suggestions on how to actually implement that? Would the user specify the options in their call to `localsolve` and then have these get passed as `**kwargs` to `as_gpconstr`?

gpkit/nomials/nomial_math.py
```diff
@@ -698,5 +698,101 @@ def sens_from_gpconstr(self, posyapprox, pa_sens, var_senss):
...
+    
+    def as_gpconstr(self, x0):
+        "Returns GP apprimxation of an SP constraint at x0"
+        #return self.left >= self.right
+        def _force_mono(posy, x0):
+            if isinstance(posy, Monomial):
+                return posy
+            return posy.mono_lower_bound(x0)
+        return [_force_mono(self.left, x0) >= _force_mono(self.right, x0),
+                _force_mono(self.left, x0) <= _force_mono(self.right, x0)]
```
whoburg: could replace the two lines above with `return _force_mono(self.left, x0) == _force_mono(self.right, x0)`.

bqpd:
Should we implement all four kinds of eq constraints, or just the linearize-both-sides one?

mopg:
In a production version I would choose one (the linearize-both-sides). However, I think a little bit more testing needs to be done on how well that one performs, before deciding to use it.

bqpd:
Hmm, that seems like an argument for implementing all four (or at least a couple) so that the user can choose...

bqpd:
If the implementation in #743 of just the full linearizer looks alright, I'll close this PR (but keep the branch)

PR #660, opened 2016-04-15
Added secant taylor expansion tool

cjk12:
Added a `te_secant` function which mirrors `te_exp_minus1`

### DIFF
--- a/gpkit/tools/tools.py
+++ b/gpkit/tools/tools.py
@@ -67,7 +67,7 @@ def te_secant(var, nterm):
         E2n = np.append(E2n, E2n_add)
 
     res = 1
-    factorial_denom = 1
+    factorial_denom = 1.
     for i in range(1, nterm + 1):
         factorial_denom *= ((2*i)*(2*i-1))
         res += E2n[i-1] / factorial_denom * var**(2*i)
--- a/gpkit/tools/tools.py
+++ b/gpkit/tools/tools.py
@@ -29,6 +29,49 @@ def te_exp_minus1(posy, nterm):
         res += posy**i / factorial_denom
     return res
 
+def te_secant(x, nterm):
+    """Taylor expansion of secant(x).
+
+    Arguments
+    ---------
+    x : gpkit.monomial
+      Variable or expression argument
+    nterm : int
+        Number of terms in resulting Taylor expansion
+
+    Returns
+    -------
+    gpkit.Posynomial
+        Taylor expansion of secant(x), carried to nterm terms
+    """
+    if nterm < 1:
+        raise ValueError("Unexpected number of terms, nterm=%s" % nterm)
+
+    # The first 12 Euler Numbers
+    E2n = np.asarray([ 1, 
+                       5,
+                       61,
+                       1385,
+                       50521,
+                       2702765,
+                       199360981,
+                       19391512145,
+                       2404879675441, 
+                       370371188237525,
+                       69348874393137901,
+                       15514534163557086905] )
+    if nterm > 12:
+        n_extend = np.asarray(range(13, nterm+1))
+        E2n_add = 8 * np.sqrt(n_extend/np.pi) * (4*n_extend/(np.pi * np.exp(1)))**(2*n_extend)
+        E2n = np.append(E2n, E2n_add)
+
+    res = 1
+    factorial_denom = 1
+    for i in range(1, nterm + 1):
+        factorial_denom *= ((2*i)*(2*i-1))
+        print factorial_denom
+        res +=  E2n[i-1]/ factorial_denom * x**(2*i)
+    return res
 
 def composite_objective(*objectives, **kwargs):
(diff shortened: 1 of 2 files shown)


cjk12:
Going to need some help here... Apparently I failed the pylint tests, but I don't know what those are...

galbramc:
Make sure you have pylint installed, then run ./pylint.sh in the gpkit folder. That will reproduce all pylint warnings. You should only fix the ones related to your changes though.

I don't see any unit tests accompanying this new function....

cjk12:
Cool, thanks.   Ok, guess I need to figure out how to write unit tests...

galbramc:
I have not looked, but I would hope there are some tests for ```te_exp_minus1``` that you can mimic.

galbramc:
Yep, ```tests/t_tools.py```.

Are you familiar with the concept and intent of unit testing? Do you know how to use code coverage to make sure you are testing what you think you are testing?

cjk12:
Cool, thanks for looking that up for me.  I'm familiar with the concept of unit testing, but have never written one successfully.  Familiar with the code coverage concept as well, but not how to actually implement in python.

galbramc:
So a good place to look up how to write unit tests in pyton is [here](https://docs.python.org/2/library/unittest.html). For coverage information you should use [coverage.py](https://coverage.readthedocs.org/en/coverage-4.0.3/). Let me know if you have trouble using either one.

You can also check the coverage information in jenkins as well by clicking on the covearge.py report linke [here](https://acdl.mit.edu/csi/job/gpkit_PullRequest_Unit/).

I'm sure @whoburg and @bqpd have lots of good tips for you as well, and they will make sure your code is well tested before merging this pull request.

cjk12:
Cool, thanks a lot for the help!

whoburg:
one idea for a unit test here would be to minimize y subject to y >= te_secant(Variable("x", np.pi/6.), nterm=4) for a given value of x, and make sure the resulting cost is close to 2./sqrt(3).

gpkit/tools.py
```diff
@@ -28,6 +28,94 @@ def te_exp_minus1(posy, nterm):
...
+        Number of terms in resulting Taylor expansion
+
+    Returns
+    -------
+    gpkit.Posynomial
+        Taylor expansion of secant(x), carried to nterm terms
+    """
+    if nterm < 1:
+        raise ValueError("Unexpected number of terms, nterm=%s" % nterm)
+
```
bqpd: Convert x to `radians` for automatic conversion in case x is in degrees or revolutions.

bqpd:
is this just waiting for a unit test, pylint, and unit conversion?

whoburg:
I rebased (fixing conflicts), force pushed, and fixed the pylint errors. Main remaining todo is adding unit tests for the two new methods. This sounds like a great task for @aechevarria1. The tests can go in `gpkit/tests/t_tools.py`.

gpkit/tools.py
whoburg: Is conversion really necessary here? Won't pint handle that?

gpkit/tools/tools.py
```diff
@@ -29,6 +29,95 @@ def te_exp_minus1(posy, nterm):
...
+        n_extend = np.asarray(range(13, nterm+1))
+        E2n_add = (8 * np.sqrt(n_extend/np.pi)
+                   * (4*n_extend/(np.pi * np.exp(1)))**(2*n_extend))
+        E2n = np.append(E2n, E2n_add)
+
+    res = 1
+    factorial_denom = 1
+    for i in range(1, nterm + 1):
+        factorial_denom *= ((2*i)*(2*i-1))
+        res += E2n[i-1] / factorial_denom * var**(2*i)
```
whoburg: this looks like a possible python integer division error to me. We should try to write a unit test that catches it before fixing.

galbramc: It should be an integer division problem, unless np.asarray converts things to floating point numbers. I don't know if it will or not.

whoburg: `np.asarray` converts to floating point if any entry is floating point, but leaves the dtype integer if all entries are integer. In an interesting twist, however, it seems the data becomes float if the integers are too large for `int64`. So I think the above actually works, for a somewhat convoluted reason.

galbramc: Lol. Well, probably best to just put period after all those numbers to be safe. Adding a float to the denominator calculation on line 72 would also be good. It would be good to remove some of the numbers from the table and make sure your unit test catches the integer division error.

aechevarria1:
I'll work on unit tests

aechevarria1:
Looking into the test failure issues

bqpd:
@aechevarria1, is this ready for review?

bqpd:
@whoburg, could you review this?

whoburg:
@bqpd, I'm ready for this to be merged. Note two added commits. Thanks @cjk12 for the PR.

Issue #661, opened 2016-04-16
Gallons not in pint


cjk12:
Any chance we could add gallons to our version of pint?  It appears to not be in regular pint and I know we have added stuff before...

whoburg:
Seems to work for me:

```python
In [1]: from gpkit import units

In [2]: units("gallon")
Out[2]: <Quantity(1, 'gallon')>
```

They even have units for both dry and liquid gallons.

cjk12:
Oh, I was using `gal` and didn't see it in the documentation...  Oops.  Thanks.

whoburg:
That's a good point about "gal" -- pint 0.6 did not have that shortcut, but 0.7.2 does. Once #665 is merged, gpkit will have "gal" too.

PR #665, opened 2016-04-17
update unit definitions to pint 0.7.2

whoburg:
This should make it easier to update unit definitions in track with pint.

The 0.6 --> 0.7.2 update includes `"gal"` as requested in #661.

Ready for review.


whoburg:
Hmm, even though this passes the unit tests on pint 0.6 and 0.7.2, if one has pint 0.6 installed and opens an ipython notebook and executes `import gpkit`, then a bunch of errors/warnings occur (only reproducable in ipython notebook).

I'm guessing that the issue is that pint 0.6 does not know how to process parts of the new `default_en.txt`.

whoburg:
until https://github.com/hgrecco/pint/issues/361 is resolved, it's probably easiest to derive the unit definitions directly from the user's pint installation, and then tack on our extras. I added a commit that does that.

Ready for review.

galbramc:
Should I add ipython in the unit testing scripts so something like this does not slip by?

whoburg:
it doesn't occur in ipython, only in an ipython notebook...

galbramc:
So is this what we should use: https://pypi.python.org/pypi/pytest-ipynb

whoburg:
brilliant, I'll add a new issue for using that to test various things

galbramc:
Let me know what you need me to do

PR #675, opened 2016-04-19
remove circular import

bqpd:
@whoburg ready for review


whoburg:
@galbramc, this is failing windows 10 unit tests due to "unable to access github.com" -- any ideas?

whoburg:
So glad to have this circular import gone! Going to merge because it seems the windows 10 failure is not caused by this.

galbramc:
I suspect the windows 10 failure was a network fluke, though we seem to have too many of those...

PR #677, opened 2016-04-20
[ll] to {ll} in array latex

whoburg:
Doing this made me notice we have triply-repeated code in latex array generation

ready for review.


bqpd:
Looks good! Geometric program will lose it once it inherits from CostedConstraintSet.

PR #685, opened 2016-04-21
add autodocs

bqpd:
@whoburg, ready for review.


whoburg:
Let's auto-generate these (via the sphinx Makefile?)

bqpd:
It is done!

bqpd:
(and pushed to master so I could confirm it was working...)

Issue #691, opened 2016-04-21
Tidy up the TightConstraintSet warnings and error messages


pgkirsch:
- [x] don't repeat both sides of the constraint twice
- [x] probably abbreviate each side of the constraint to ~30 characters
- [x] write something about the target reltol and the actual error
- [x] blank lines between warnings

bqpd:
@pgkirsch, I think this was covered by the PR?

PR #697, opened 2016-04-23
Cleaner tight constraint message

pgkirsch:
Addresses issue #691 

Ready for review @whoburg @bqpd


bqpd:
Any idea why this is failing tests?

pgkirsch:
Hmm @bqpd is there something like `.magnitude` that works for both variables with units and without units? Seems to be causing the the units tests to fail.

bqpd:
Yes! Use the `mag` function from `small_scripts`

pgkirsch:
eurgh I was worried you would say that :P

pgkirsch:
Should be good to go

whoburg:
retest this please

PR #706, opened 2016-04-30
A more useful plot_convergence

pgkirsch:
Only plots cost convergence instead of trying to plot variables convergence

Ready for review @whoburg  @bqpd

### DIFF
--- a/gpkit/interactive/plotting.py
+++ b/gpkit/interactive/plotting.py
@@ -275,22 +275,15 @@ def plot_convergence(model):
     Returns
     -------
     matplotlib.pyplot Figure
-        Semilogy plot of variable values as functions of SP iteration #
+        Plot of cost as functions of SP iteration #
     """
-    newdict = {}
-
     fig, ax = plt.subplots()
 
-    for key in model.program.gps[0].result['variables']:
-        a = np.array([])
-        for j in range(len(model.program.gps)):
-            a = np.append(a, model.program.gps[j].result['variables'][key])
-        newdict[key] = a
-        ax.semilogy(np.arange(len(newdict[key])), newdict[key], label=key.name)
-
-    box = ax.get_position()
-    ax.set_position([box.x0, box.y0, box.width * 0.8, box.height])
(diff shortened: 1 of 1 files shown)


pgkirsch:
(Bump)


(Should be a really easy review)

bqpd:
Attach an example plot and I'll merge it. :D

pgkirsch:
convergence_wi

How's this?

bqpd:
:+1:

whoburg:
sweet. minor design discussion: should we plot this on a semilogy scale?

bqpd:
Hmm! I'd argue no; we're checking the effects of negynomials, which I wouldn't expect to span too many orders of magnitude?

bqpd:
Actually, gonna change my mind and argue yes, since the terminating distinction is reltol...

pgkirsch:
The downside being a lot of these plots end up being flat on a semiology scale (as suggested by @bqpd)

bqpd:
Can't we autoscale the y axis to prevent that?

pgkirsch:
Hm, yes. But why does the log scale help in this case?

On May 14, 2016, at 12:19, bqpd <notifications@github.com<mailto:notifications@github.com>> wrote:


Can't we autoscale the y axis to prevent that?

bqpd:
Oh, because the sp solve terminates when the previous solution is close enough to the current solution in  logspace, so plotting that directly might be more easily readable?

bqpd:
Clearly there's arguments both ways, though. :p

pgkirsch:
Well then perhaps it makes most sense to plot the absolute difference from the final cost. Otherwise I'm not sure you can "see" the convergence condition being satisfied.

On May 14, 2016, at 12:25, bqpd <notifications@github.com<mailto:notifications@github.com>> wrote:


Clearly there's arguments both ways, though. :p

bqpd:
Hmm, any multiplicative distance (like reltol) should show up as a fixed delta-y.

PR #708, opened 2016-05-02
band-aid fixes for sensitivity map -- needs more thorough refactor

whoburg:
@bqpd ready for review but more importantly this needs a more thorough refactor

sensitivity map currently does not work for the 1682 model due to vector constants, and this does not yet address that problem


bqpd:
Looks good to me! Do you want to merge it, or to get it working with vector variables first?

bqpd:
@whoburg, there's a push to merge this because it fixed some pylint errors that now show up on all other pull requests.

Issue #713, opened 2016-05-09
A couple sol.table(latex=2) bugs


pgkirsch:
- [x] Creates rows with three ampersands (instead of two) between sub-models.
- [ ] Should put square brackets around units

pgkirsch:
In latex, too few columns is ok, whereas too many is not, so the easiest proposed fix to item 1 is replace 
```python
lines.append(["","", "", ""])
```
with 
```python
lines.append(["", "", ""])
```
on line 228.

pgkirsch:
Actually, I realize that @bqpd and @whoburg might disagree with the second one (I think the code was changed away from square brackets?) so I'll leave that open for discussion. In the meantime the proposed change is ready for a super fast merge.

pgkirsch:
Also piggy-backing on this issue: sensitivity table shouldn't print for `latex=2`.

pgkirsch:
Sorry realized that there is a kwarg for achieving that last thing.

bqpd:
Great! Re: square brackets around units, I'm fine with however you decide to represent them.

whoburg:
I vote for no square braces around units in latex tables, because it seems like unnecessary clutter. Square brackets seem useful in text printouts where there are no columns to disambiguate.

If others agree, then I think this is closed by #714

pgkirsch:
Sounds good

pgkirsch:
You probably disagree with this, but I do wish that unitless variables would have a hyphen or something in the units column to indicate that they have no units, so that the table doesn't look incomplete...

bqpd:
Yeah, for a table a thin horizontal line probably makes sense.

whoburg:
What's a clean way to do this that avoids adding special cases?

the latex table unit string is generated by calling `var.unitstr`, which in turn calls `small_scripts.unitstr`, which has a `dimless` kwarg.

Should we add a `dimless` kwarg to `VarKey.unitstr`?

whoburg:
or just pass through `**kwargs` in `VarKey.unitstr`?

bqpd:
The latter sounds best.

whoburg:
also should `VarKey.unitstr` be renamed to `VarKey.latex_unitstr`? Or alternatively, should it have a `latex=True/False` kwarg, so it can be used for non-latex printouts too?

bqpd:
Yeah, it should probably be renamed.

bqpd:
[...]

PR #714, opened 2016-05-09
simple fix to #713

pgkirsch:


bqpd:
Annoying that this triggers pylint errors...

galbramc:
The pylint warnings seem legitimate though:

```
sensitivity_map.py:35, PyLint, Priority: Normal
Access to a protected member _latex of a client class

sensitivity_map.py:106, PyLint, Priority: Normal
Access to a protected member _latex of a client class
```
Looks like you are calling a protected member outside of the class. If you intend to have latex() called from outside the class, it should not have an underscore in the name. At least that's my understanding.

bqpd:
Oh, the calls are legit, but they're not in the single line of code this PR alters, and fixing them entails much more work / design discussion than the actual fix that motivated this PR.

galbramc:
Was there a pull request that has not been merged yet that fixed these warnings? I think that's the only way why these warnings would not be ignored right now. Any successful pull request effectively resets the counts.

bqpd:
Oh! Yeah, it looks like #708 fixed them but hasn't been merged yet. Will merging this not actually cause pylint errors, then?

galbramc:
Yeah, so if you merge this it will not cause any errors on gpkit_push.

We can reset the counts, or you can merge #708 to fix the fact that you get these pylint warnings on all pull requests at the moment. This is why it is problematic to have unfixed pylint warnings.

PR #719, opened 2016-05-14
allow tautological monomial equalities (with unit test)

### LINKED ISSUE #601: equality constraints with no variables

bqpd:
Closes #601

### DIFF
--- a/gpkit/tests/t_model.py
+++ b/gpkit/tests/t_model.py
@@ -42,6 +42,16 @@ class TestGP(unittest.TestCase):
                                self.ndig)
         self.assertAlmostEqual(sol["cost"], 2*math.sqrt(2), self.ndig)
 
+    def test_601(self):
+        # tautological monomials should solve but not pass to the solver
+        x = Variable("x")
+        y = Variable("y", 2)
+        m = Model(x,
+                  [x >= 1,
+                   y == 2])
+        m.solve(verbosity=0)
+        self.assertEqual(len(m.program.constraints), 2)
+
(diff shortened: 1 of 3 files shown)


bqpd:
@whoburg please merge this if it passes your review, it's an issue in Next Release

gpkit/tests/t_model.py
```diff
@@ -42,6 +42,16 @@ def test_trivial_gp(self):
...
 
+    def test_601(self):
+        # tautological monomials should solve but not pass to the solver
+        x = Variable("x")
+        y = Variable("y", 2)
+        m = Model(x,
+                  [x >= 1,
+                   y == 2])
+        m.solve(verbosity=0)
+        self.assertEqual(len(m.program.constraints), 2)
```
whoburg: the length 2 assertion seems to conflict with the inline comment above -- if "tautological monomials should solve but not pass to the solver", then shouldn't the length of `m.program.constraints` be 1?

bqpd: Oh, the first constraint is the cost

Issue #561, opened 2016-03-09
unit rendering in latex


whoburg:
Latex for units should use `\textnormal` or similar to keep them from being printed in italics math. Here's an example screenshot from a solution table produced with `SolutionArray.table(latex=True)`.

![screen shot 2016-03-09 at 4 48 08 pm](https://cloud.githubusercontent.com/assets/6042395/13651793/e9cfe00e-e616-11e5-9c98-5bd54a3a333c.png)

bqpd:
It uses mathrm now, I believe.

PR #722, opened 2016-05-26
Solution table improvements

### LINKED ISSUE #713: A couple sol.table(latex=2) bugs

pgkirsch:
Somewhat related to the recently closed #713. I know less is more but I truly believe these changes make latex tables look better. These changes are:
- A dash for unitless variables (instead of just whitespace)
- Square brackets around units (clear distinction between variables and units)
- Removed the extra line of whitespace that was printing below model names

This is the code used for generating the tables in my thesis.

Ready for review (and inevitable discussion) @whoburg @bqpd

### DIFF
--- a/gpkit/solution_array.py
+++ b/gpkit/solution_array.py
@@ -231,16 +231,18 @@ def results_table(data, title, minval=0, printunits=True, fixedcols=True,
             lines.append([varstr, valstr, units, label])
         else:
             varstr = "$%s$" % varstr.replace(" : ", "")
-            unitstring = var.unitstr()
-            if not unitstring: unitstring = "-"
+            boxunit = var.unitstr()
+            if not boxunit:
+                boxunit = "-"
             if latex == 1:  # normal results table
-                lines.append([varstr, valstr, "$\\left[%s\\right]$" % unitstring, label])
+                lines.append([varstr, valstr, "$\\left[%s\\right]$" % boxunit,
+                              label])
                 coltitles = [title, "Value", "Units", "Description"]
             elif latex == 2:  # no values
-                lines.append([varstr, "$\\left[%s\\right]$" % unitstring, label])
+                lines.append([varstr, "$\\left[%s\\right]$" % boxunit, label])
                 coltitles = [title, "Units", "Description"]
             elif latex == 3:  # no description
-                lines.append([varstr, valstr, "$\\left[%s\\right]$" % unitstring])
+                lines.append([varstr, valstr, "$\\left[%s\\right]$" % boxunit])
                 coltitles = [title, "Value", "Units"]
             else:
                 raise ValueError("Unexpected latex option, %s." % latex)
--- a/gpkit/solution_array.py
+++ b/gpkit/solution_array.py
@@ -216,7 +216,7 @@ def results_table(data, title, minval=0, printunits=True, fixedcols=True,
                     lines.append([model+" | ", "", "", ""])
                 else:
                     lines.append([r"\multicolumn{3}{l}{\textbf{" +
-                                  model + r"}} \\"])
+                                  model + r"}}"])
             oldmodel = model
         label = var.descr.get('label', '')
         units = unitstr(var, into=" [%s] ", dimless="") if printunits else ""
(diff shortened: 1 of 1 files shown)


bqpd:
retest this please

bqpd:
@whoburg I'm ready to merge this

bqpd:
Does this close #561?

bqpd:
@whoburg disagrees

gpkit/solution_array.py
```diff
@@ -231,14 +231,18 @@ def results_table(data, title, minval=0, printunits=True, fixedcols=True,
             lines.append([varstr, valstr, units, label])
         else:
             varstr = "$%s$" % varstr.replace(" : ", "")
+            boxunit = var.unitstr()
```
bqpd: unitstr has a `dimless` argument that may prove useful

bqpd: also there's the `into` arg; square brackets with `[-]` for dimensionless quantities could be as simple as a couple of arguments that get passed to unitstr

gpkit/solution_array.py
```diff
@@ -216,7 +216,7 @@ def results_table(data, title, minval=0, printunits=True, fixedcols=True,
                     lines.append([model+" | ", "", "", ""])
                 else:
                     lines.append([r"\multicolumn{3}{l}{\textbf{" +
-                                  model + r"}} \\"])
+                                  model + r"}}"])
```
whoburg: I'm fine with this change

gpkit/solution_array.py
```diff
@@ -231,14 +231,18 @@ def results_table(data, title, minval=0, printunits=True, fixedcols=True,
             lines.append([varstr, valstr, units, label])
         else:
             varstr = "$%s$" % varstr.replace(" : ", "")
+            boxunit = var.unitstr()
+            if not boxunit:
+                boxunit = "-"
```
whoburg: I don't think this change is necessary -- or at least it should be a user option what to print in the `dimless` case.

gpkit/solution_array.py
```diff
@@ -231,14 +231,18 @@ def results_table(data, title, minval=0, printunits=True, fixedcols=True,
...
+                lines.append([varstr, valstr, "$\\left[%s\\right]$" % boxunit,
+                              label])
                 coltitles = [title, "Value", "Units", "Description"]
             elif latex == 2:  # no values
-                lines.append([varstr, "$%s$" % var.unitstr(), label])
+                lines.append([varstr, "$\\left[%s\\right]$" % boxunit, label])
                 coltitles = [title, "Units", "Description"]
             elif latex == 3:  # no description
-                lines.append([varstr, valstr, "$%s$" % var.unitstr()])
+                lines.append([varstr, valstr, "$\\left[%s\\right]$" % boxunit])
```
whoburg: I'm not a fan of putting brackets around units to be printed in a table -- again would prefer to let rogue users do that themselves  :P

whoburg:
likely resolution of this pr is to accept the first change (removal of `\\`), and then make some of the unit printing options kwargs (and hopefully match naming conventions of `unitstr`).

bqpd:
The square brackets do make sense in @pgkirsch's thesis, so maybe we should put a comparison up on this thread for discussion?

bqpd:
@whoburg what's the status of this

whoburg:
my takeaway from this thread is that users want (and deserve) more custom options for latex printing. I think a good start on this is to standardize the `into=` and `dimless=` kwargs across all nomials, solution arrays, Model objects, etc.

The need for a latex printing refactor also surfaced in introducing `small_scripts.nomial_latex_helper` in #841.

I prefer adding customization options over simply changing the default.

whoburg:
The crux here is figuring out how to get `into` to act at the right level -- e.g. maybe it should be a dict that allows into to be specified for variables, nomials, unit strings, etc? Correctly done, this would also make implementing parts of `sensitivity_map` almost trivial.

bqpd:
Yeah, I think we need a framework for modifying/filtering display...

bqpd:
Closing because ewww square brackets. (we should make it so users can add them themselves, but that's another issue)

pgkirsch:
Boooooo

On Sep 9, 2016, at 11:18, bqpd <notifications@github.com<mailto:notifications@github.com>> wrote:


Closing because ewww square brackets. (we should make it so users can add them themselves, but that's another issue)

PR #723, opened 2016-05-26
A very small plot_convergence improvement

pgkirsch:
The x axis starts from iteration 1 instead of iteration 0, for less ambiguity and more readability.

This is the code that was used in the final version of my thesis.

Ready for a cursory review @whoburg @bqpd


bqpd:
Looks good to me.

PR #727, opened 2016-05-29
Make it possible to pass arguments to cvxopt solver

whoburg:
Fixed a bug and added a test. See inline comments for more details.

### DIFF
--- a/gpkit/_cvxopt.py
+++ b/gpkit/_cvxopt.py
@@ -2,10 +2,8 @@
 from cvxopt import solvers, spmatrix, matrix, log
 
 
-def cvxoptimize_fn(options=None):
-    "Return a cvxopt solve function for a particular set of options."
-    if options:
-        solvers.options.update(options)
+def cvxoptimize_fn():
+    "Return a cvxopt solve function"
     gpsolver = solvers.gp
 
     # pylint: disable=unused-argument
--- a/gpkit/geometric_program.py
+++ b/gpkit/geometric_program.py
@@ -152,7 +152,7 @@ class GeometricProgram(NomialData):
 
             if solver == "cvxopt":
                 from ._cvxopt import cvxoptimize_fn
-                solverfn = cvxoptimize_fn(*args, **kwargs)
+                solverfn = cvxoptimize_fn()
             elif solver == "mosek_cli":
                 from ._mosek import cli_expopt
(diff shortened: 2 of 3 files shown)


gpkit/_cvxopt.py
```diff
@@ -2,10 +2,8 @@
 from cvxopt import solvers, spmatrix, matrix, log
 
 
-def cvxoptimize_fn(options=None):
-    "Return a cvxopt solve function for a particular set of options."
-    if options:
-        solvers.options.update(options)
```
whoburg: this update of `solvers.options` has non-intuitive side effects, so I've removed it. For example, if one runs

`m.solve(solver="cvxopt", options={"feastol": 1e-6})`

and then runs

`m.solve(solver="cvxopt")`,

the second call will still have the feastol option from the earlier call.

gpkit/geometric_program.py
```diff
@@ -152,7 +152,7 @@ def _get_solver(solver):
 
             if solver == "cvxopt":
                 from ._cvxopt import cvxoptimize_fn
-                solverfn = cvxoptimize_fn(*args, **kwargs)
+                solverfn = cvxoptimize_fn()
```
whoburg: `cvxoptimize_fn` did not accept args or kwargs, so this line fails if a user tries to specify args or kwargs.

gpkit/tests/t_model.py
```diff
@@ -390,10 +390,22 @@ def test_partial_sub_signomial(self):
...
+class TestModelSolverSpecific(unittest.TestCase):
+    """test cases run only for specific solvers"""
+    def test_cvxopt_kwargs(self):
+        if "cvxopt" not in settings["installed_solvers"]:
+            return
+        x = Variable("x")
+        m = Model(x, [x >= 12])
+        # make sure it's possible to pass the kktsolver option to cvxopt
+        sol = m.solve(solver="cvxopt", verbosity=0, kktsolver="ldl")
+        self.assertAlmostEqual(sol["cost"], 12., NDIGS["cvxopt"])
```
whoburg: here's a place to put tests in `t_model` that shouldn't get run for every solver.

whoburg:
I'm going to merge this. It passes tests and is needed on a project.

@bqpd, can we please do a minor pip release to capture the recent bug fixes?

bqpd:
For sure!

PR #728, opened 2016-05-29
don't forget to pass args kwargs to solver

whoburg:


whoburg:
@galbramc any ideas on this error we're seeing? Is it possible a very old version of cvxopt is somehow installed on the Windows 7 machine? It's odd that the `kktsolver` argument works on all machines except that one.

whoburg:
Ok, I know what's going on, and it's gross.

Python x,y, which we currently recommend for Windows 7 users, includes cvxopt ﻿﻿1.1.7, instead of the latest 1.1.8. And 1.1.7 did not have a `*args, **kwargs` pass-thru from `gp` to `cp`.

@galbramc, have we tried the cvxopt build instructions (http://cvxopt.org/install/index.html#building-cvxopt-for-windows) instead of Python x, y on the Windows 7 CI machine?

I wonder if @martinandersen has a way to ping the Python x,y folks to update their cvxopt version.

And meanwhile, I'll see if I can implement a local patch for our windows users.

whoburg:
@martinandersen, the above is probably out of context since you can't see our continuous integration messages. The issue here is that we're passing the keyword arg `kktsolver="ldl"` to cvxopt's `solver.gp` method, and this didn't work for cvxopt 1.1.7, which our Windows 7 python x,y users have.

galbramc:
I recall trying unsuccessfully to compile cvxopt, but I don't know if those were the exact instructions I was following. If it's cumbersome, then it will be tough on your users to have to compile cvxopt.

martinandersen:
Unfortunately I do not have a way to ping the Python x,y folks.

The `solvers.gp` routine is a wrapper for `solver.cp` which did not include the `kktsolver` option until 1.1.8.

whoburg:
Planning to merge this in about an hour unless anyone disagrees.

PR #735, opened 2016-06-15
fix bug with hashing of Variables with units; add test

whoburg:
found a bug: identical Variables with units have different hashes.

@bqpd, could this explain the problem you were having with getting @pgkirsch's models to run with your refactor in progress?

ready for review after passing CI.


bqpd:
Hmm! I did see that bug while tracking down why nomialmaps wasn't working with Airplane, I'll have to check if this was the issue!

(ready to merge)

bqpd:
Unfortunately this was not the issue. :-/

PR #787, opened 2016-08-04
Fixedvarkeys

### LINKED ISSUE #784: models should be able to have variables that are _not_ in constraints
created during init by `self.varkeys.add(W)` etc?

bqpd:
Fix varkeys on model creation, and make it easy to add to that list. Closes #784 and makes hoburg/gpkit-models#17 possible.


whoburg:
double-check the ticket numbers this closes? 17 must be a typo.

bqpd:
Oops, yep, meant in gpkit-models

gpkit/keydict.py
```diff
@@ -149,6 +149,11 @@ class KeySet(KeyDict):
     "KeySets are KeyDicts without values, serving only to filter and map keys"
     collapse_arrays = False
 
+    def add(self, *args):
+        "Adds all args to the keyset"
+        for arg in args:
+            self[arg] = None
```
whoburg: python's `set.add` takes exactly one argument. I think we should match that interface here. https://docs.python.org/2/library/sets.html#sets.Set

gpkit/constraints/set.py
```diff
@@ -127,19 +128,17 @@ def subinplace(self, subs, value=None):
...
+        varkeys = KeySet()
+        if not init_dict is None:
+            varkeys.update(init_dict)
         for constraint in self:
             if hasattr(constraint, "varkeys"):
-                out.update(constraint.varkeys)
-        return out
+                varkeys.update(constraint.varkeys)
+            # else:
+            #     print type(constraint) # getting numpy bools...
```
whoburg: was the debugging printout above resolved?

gpkit/constraints/costed.py
```diff
@@ -24,10 +24,12 @@ def subinplace(self, subs, value=None):
         self.cost = self.cost.sub(subs, value)
         ConstraintSet.subinplace(self, subs, value)
 
-    @property
-    def varkeys(self):
+    def reset_varkeys(self, init_dict=None):
         "return all Varkeys present in this ConstraintSet"
```
whoburg: the method changed but the docstring did not -- is the docstring still accurate?

docs/source/examples/beam.py
```diff
@@ -47,6 +47,8 @@ def __init__(self, N=4, **kwargs):
         theta_eq[0] = (th[0] >= th_base)  # base boundary condition
         displ_eq = (w >= w.left + 0.5*dx*(th + th.left))
         displ_eq[0] = (w[0] >= w_base)
+        for constraint in [shear_eq, moment_eq, theta_eq, displ_eq]:
+            constraint.reset_varkeys()
```
whoburg: hmm, for backwards compatibility reasons I don't like that the example had to change. Conceptually, why is this reset needed here?

bqpd: Because we're adding new varkeys (e.g. w_base) after the array constraints are created. I don't like it a ton either.

gpkit/constraints/costed.py
bqpd: nerp

gpkit/constraints/set.py
bqpd: It's still true, seems to have been true for a while.

gpkit/keydict.py
bqpd: Then how should we add multiple?

docs/source/examples/beam.py
whoburg: I'm not convinced the beam example is best practice (we might want to implement it), but I don't think users should have to `reset_varkeys` for simple examples.

gpkit/keydict.py
whoburg: same way people add multiple things to a set -- I assume they write a for loop? Or also implement union, which takes an iterable?

docs/source/examples/beam.py
bqpd: I agree, but don't want setitem to reset varkeys. Alternatives?

gpkit/keydict.py
bqpd: Will do.

gpkit/constraints/set.py
bqpd: How about leaving it in & raising a separate issue?

docs/source/examples/beam.py
bqpd: Honestly, the right thing to do here is to use the new .union method

bqpd: Ended up using add, how's that look?

gpkit/constraints/set.py
whoburg: sounds good

docs/source/examples/beam.py
whoburg: better maybe, but I still have the backwards compatibility hesitation.

What about having array constraint's `__setitem__` call reset_varkeys? Doesn't that get directly at the issue here?

gpkit/keydict.py
```diff
@@ -149,10 +149,14 @@ class KeySet(KeyDict):
...
-        for arg in args:
-            self[arg] = None
+    def add(self, item):
+        "Adds an item to the keyset"
+        self[item] = None
+
+    def union(self, iterable):
+        "Adds an iterable to the keyset"
+        for item in iterable:
+            self.add(item)
```
whoburg: 👍

docs/source/examples/beam_output.txt
```diff
@@ -1,33 +0,0 @@
...
- \vec{\theta} : [ 0          -         -         -       ... ]           Slope                         
-      \vec{q} : [ 110       110       110       110      ... ]  [N/m]    Distributed load at each point
-      \vec{w} : [ 0          -         -         -       ... ]  [m]      Displacement                  
-
-Sensitivities
--------------
-      L : 4                                              Overall beam length           
-\vec{q} : [ 0.0072    0.0416    0.118     0.234    ... ] Distributed load at each point
-     EI : -1                                                                           
-
```
whoburg: we should rebase this PR so that this deletion can go away.

gpkit/constraints/set.py
```diff
@@ -127,19 +128,17 @@ def subinplace(self, subs, value=None):
...
+        self.varkeys = varkeys
```
whoburg: I have an external model that now fails on this branch -- the failure occurs on this line by raising an `AttributeError: can't set attribute`. Here's a relevant portion of the traceback (note `lc` here is a LinkedConstraintSet):

```
--> 375         Model.__init__(self, None, lc, **kwargs)
    376
    377     @property

/Users/whoburg/MIT/dev/gpkit/gpkit/constraints/model.pyc in __init__(self, cost, constraints, substitutions, name)
     56         cost = cost if cost else Monomial(1)
     57         constraints = constraints if constraints else []
---> 58         CostedConstraintSet.__init__(self, cost, constraints, substitutions)
     59         if self.__class__.__name__ != "Model" and not name:
     60             name = self.__class__.__name__

/Users/whoburg/MIT/dev/gpkit/gpkit/constraints/costed.pyc in __init__(self, cost, constraints, substitutions)
     18         if substitutions:
     19             subs.update(substitutions)
---> 20         ConstraintSet.__init__(self, constraints, subs)
     21
     22     def subinplace(self, subs, value=None):

/Users/whoburg/MIT/dev/gpkit/gpkit/constraints/set.pyc in __init__(self, constraints, substitutions, recursesubs)
     29             subs.update(constraints.substitutions)  # pylint: disable=no-member
     30         if recursesubs:
---> 31             self.reset_varkeys()
     32             self.substitutions = KeyDict.with_keys(self.varkeys,
     33                                                    self._iter_subs(subs))

/Users/whoburg/MIT/dev/gpkit/gpkit/constraints/costed.pyc in reset_varkeys(self, init_dict)
     30         if init_dict is not None:
     31             costkeys.update(init_dict)
---> 32         ConstraintSet.reset_varkeys(self, costkeys)
     33
     34     def rootconstr_str(self, excluded=None):

/Users/whoburg/MIT/dev/gpkit/gpkit/constraints/set.pyc in reset_varkeys(self, init_dict)
    139             # else:
    140             #     print type(constraint) # getting numpy bools...
--> 141         self.varkeys = varkeys
    142
    143     def as_posyslt1(self):

AttributeError: can't set attribute
```

bqpd: Can you email me the code? Something that inherits from ConstraintSet seems to still have varkeys as a property, but I'm certain that's not the case on this branch...maybe you have two different gpkit instances colliding?

docs/source/examples/beam.py
bqpd: I mentioned a couple hours ago I didn't like that. :p Mostly because it makes setitem slow, and if users have updated their constraintset's varkeys by hand they won't expect those varkeys to disappear just because they used setitem.

whoburg: fair -- could `__setitem__` do something more subtle, like adding the new variables introduced by whatever is passed in, but not calling `reset_varkeys`?

Another option is to try to rewrite Beam in a better way that doesn't mutate array constraints

whoburg: What's a case where `ArrayConstraint.__setitem__` does NOT need to involve a varkey update?

gpkit/constraints/set.py
whoburg: Sadly cannot e-mail this. Yes, turns out this was caused by the following hack:

```python
    @property
    def varkeys(self):
        # local speedy varkey access -- this is temporary until we
        # implement this more generally in gpkit
        if self._fastvks:
            return self._fastvks
        # pylint:disable=no-member
        self._fastvks = Model.varkeys.fget(self)
        return self._fastvks
```

whoburg: Next issue on this branch is with SummingConstraintSet, implemented as follows:

```python
class SummingConstraintSet(ConstraintSet):
    def __init__(self, lhs, varname, models=[], variables=[], **kwargs):
        summedvars = set([v.key for v in variables])
        alreadysummed = set()
        for model in models:
            mvars = model[varname]
            if not hasattr(mvars, "__len__"):
                mvars = [mvars]
            # next line makes the recursion stop at depth one
            # for safety to avoid double counting
            mvars = [v for v in mvars if v.key.models[0] == model.name]
            assert len(mvars) == 1
            summedvars = summedvars.union([v.key for v in mvars])
            for constraint in model.flat():
                if hasattr(constraint, "summedvars"):
                    alreadysummed = alreadysummed.union(constraint.summedvars)
        summedvars = summedvars.difference(alreadysummed)
        ConstraintSet.__init__(self, [lhs >= sum(Variable(**vk.descr)
                                                 for vk in summedvars)],
                               **kwargs)
```

When trying to initialize a SCS, get the following traceback:
```
     12             # next line makes the recursion stop at depth one
     13             # for safety to avoid double counting
---> 14             mvars = [v for v in mvars if v.key.models[0] == model.name]
     15             assert len(mvars) == 1
     16             summedvars = summedvars.union([v.key for v in mvars])

TypeError: 'NoneType' object has no attribute '__getitem__'
```

gpkit/constraints/set.py
whoburg: yep, lotsa bools on master, problem does not appear to be introduced by this PR

gpkit/constraints/set.py
bqpd: This error occurred because there was a variable without a model, perhaps fixed by my latest commit.

docs/source/examples/beam.py
bqpd: There's several, but the biggest issue is now fixed with a static unused_variables list, so I'm implementing this.

whoburg: Implementing for `ArrayConstraint.__setitem__` or `ConstraintSet.__setitem__`?

whoburg: n/m, I see the commit (ConstraintSet)

gpkit/constraints/set.py
whoburg: yeah, looks like model names are not getting appended to all varkeys on this branch. That's at least partially responsible for the error I'm seeing in the external model.

whoburg: and that appears to be true even as of the latest merge commit

bqpd: Try the new d8, just pushed.

whoburg:
This branch works for d8 but not for another external model. Issue appears to be with how model names are appended to variables.

gpkit/keydict.py
whoburg: actually I think this should be called `update`. `set.union` returns the union but does not mutate self. Update causes a mutation, like the implementation here.

gpkit/constraints/set.py
```diff
@@ -127,19 +132,17 @@ def subinplace(self, subs):
...
-        return self._varkeys()
-
-    def _varkeys(self, init_dict=None):
-        "return all Varkeys present in this ConstraintSet"
-        init_dict = {} if init_dict is None else init_dict
-        out = KeySet(init_dict)
+    def reset_varkeys(self, init_dict=None):
+        "Goes through constraints and collects their varkeys."
+        varkeys = KeySet()
+        if not init_dict is None:
```
whoburg: I think preferred syntax is `if init_dict is not None`

gpkit/keydict.py
bqpd: done!

gpkit/constraints/set.py
bqpd: done!

bqpd:
retest this please

bqpd:
@whoburg, model name appending to variables hasn't changed, so I don't think that's it?

whoburg:
nope, commenting out that line did not fix it. I'll debug a bit

whoburg:
model that's not getting model names appended is a LinkedConstraintSet, if that matters

whoburg:
Here's a MWE for not adding modelname to varkeys:

```python
from gpkit import Variable, VectorVariable, Model

class Thing(Model):
    "mass per unit length of nodes"
    def __init__(self, n, **kwargs):
        a = VectorVariable(n, "a", "g/m")
        b = VectorVariable(n, "b", "m")
        c = Variable("c", 17/4., "g")
        Model.__init__(self, None, [a >= c/b], **kwargs)


t = Thing(3)
print t.varkeys
```

output is 
```
{c: None, b_(1,): None, a_(1,): None, a_(0,): None, a_(2,): None, b_(2,): None, b_(0,): None}
```

whoburg:
added that MWE as a test to fix

bqpd:
Passes the test locally, now.

whoburg:
ok excellent, that was the issue. I'm going to try rebasing this onto master.

whoburg:
@mayork, would you mind switching to this branch briefly and confirming your model(s) run fine?

whoburg:
ok, the one ubuntu16 test failure was not caused by this; I'm ready to merge this.

mayork:
@whoburg I'm having trouble checking this out...I'm not sure if I'm doing something stupid but I can checkout branches like master equality and breakdown but get the following error when I try to check this out...its called `fixedvarkeys` right?

```
dhcp-18-111-87-78:gpkit mayork$ git checkout fixedvarkeys
error: pathspec 'fixedvarkeys' did not match any file(s) known to git.
dhcp-18-111-87-78:gpkit mayork$ git checkout equality
Already on 'equality'
Your branch is up-to-date with 'origin/equality'.
```

whoburg:
do `git fetch` first, and then you should be able to check out `fixedvarkeys`.

mayork:
okay got the branch installing and testing right now

mayork:
sorry slight delay I think I just broke the atmosphere model

PR #792, opened 2016-08-04
don't print table automatically, print timing

### LINKED ISSUE #779: don't print table by default
make the user do `print sol.table()`

bqpd:
Closes #779

### DIFF
--- a/docs/source/examples/external_sp.py
+++ b/docs/source/examples/external_sp.py
@@ -15,4 +15,5 @@ constraints = [ExternalConstraint(x, y),
               ]
 
 m = Model(objective, constraints)
-sol = m.localsolve(verbosity=1)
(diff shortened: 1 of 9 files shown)


docs/source/examples/external_sp.py
```diff
@@ -12,7 +12,8 @@
 constraints = [ExternalConstraint(x, y),
                x <= np.pi/2.,
                x >= np.pi/4.,
-              ]
+               ]
```
whoburg: our pylint doesn't like this -- I'll fix

whoburg:
I'm going to rebase this onto master to hopefuly fix the pylint issues

whoburg:
looks like windows machine might be staled (but it previously passed). Merging.

Issue #724, opened 2016-05-27
vector variable indices scrambled


whoburg:
High priority. Have found a bug where VectorVariable indices get scrambled in constraints involving separate models

MWE:

```python
from gpkit import VectorVariable, Variable, Model

N = 15

class XgtRange(Model):
    def __init__(self, **kwargs):
        X = VectorVariable(N, "X", "m")
        X_min = VectorVariable(N, "X_{min}", range(1, N + 1), "m")
        constraints = [X >= X_min]

        Model.__init__(self, sum(X), constraints, **kwargs)

class Y(Model):
    def __init__(self, **kwargs):
        Y = VectorVariable(N, "Y", "m")
        xgtr = XgtRange()
        constraints = [xgtr,
                       Y >= xgtr["X"]]
        Model.__init__(self, sum(Y), constraints, **kwargs)


if __name__ == "__main__":
    print "-- just XgtRange --"
    xm = XgtRange()
    sol = xm.solve(verbosity=0)
    print "X is ", sol("X")
    
    print "-- now Y --"
    ym = Y()
    sol = ym.solve(verbosity=0)
    print "X is ", sol("X")
    print "Y is ", sol("Y")
```

Output from the above (note that Y is scrambled unexpectedly):

```
-- just XgtRange --
X is  [  1.00000003   2.00000624   3.00000734   4.00000675   5.00000571
   6.00000461   7.00000361   8.00000274   9.00000201  10.00000142
  11.00000095  12.0000006   13.00000033  14.00000015  15.00000005] meter
-- now Y --
X is  [  1.           2.00000031   3.00000035   4.0000003    5.00000023
   6.00000017   7.00000012   8.00000008   9.00000005  10.00000003
  11.00000002  12.00000001  13.00000001  14.          15.        ] meter
Y is  [  7.00000024  14.          10.00000007   3.0000007    6.00000034
   8.00000016   2.00000062   4.0000006   11.00000004  13.00000001
   9.00000011  15.           5.00000046  12.00000002   1.        ] meter
```

The underlying issue appears to be `__getitem__` involving a VectorVariable. `print xm["X"]` returns:
```
[...]

PR #793, opened 2016-08-04
getitem only returns top variable

### LINKED ISSUE #786: Model.__getitem__ should return highest-level varkey
list of varkeys all with the same name should be in `.byname`. Also there could be a fuzzy search, `.search("W wing")` which would get all of `W_wing`, `W with model=['wing']`, `W descr="wing weight"` and return a list sorted by relevance (or just the most relevant)

bqpd:
closes #786 

does not work for vectorvariables.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -61,7 +61,7 @@ confidence=
 # --disable=W"
 #disable=import-star-module-level,old-octal-literal,oct-method,print-statement,unpacking-in-except,parameter-unpacking,backtick,old-raise-syntax,old-ne-operator,long-suffix,dict-view-method,dict-iter-method,metaclass-assignment,next-method-called,raising-string,indexing-exception,raw_input-builtin,long-builtin,file-builtin,execfile-builtin,coerce-builtin,cmp-builtin,buffer-builtin,basestring-builtin,apply-builtin,filter-builtin-not-iterating,using-cmp-argument,useless-suppression,range-builtin-not-iterating,suppressed-message,no-absolute-import,old-division,cmp-method,reload-builtin,zip-builtin-not-iterating,intern-builtin,unichr-builtin,reduce-builtin,standarderror-builtin,unicode-builtin,xrange-builtin,coerce-method,delslice-method,getslice-method,setslice-method,input-builtin,round-builtin,hex-method,nonzero-method,map-builtin-not-iterating
 
-disable=superfluous-parens,locally-disabled,misplaced-comparison-constant,too-few-public-methods,no-self-use
+disable=superfluous-parens,locally-disabled,misplaced-comparison-constant,too-few-public-methods,no-self-use,invalid-sequence-index
 
 [REPORTS]
 
--- a/gpkit/constraints/set.py
+++ b/gpkit/constraints/set.py
@@ -5,6 +5,16 @@ from ..small_scripts import try_str_without
 from ..repr_conventions import _str, _repr, _repr_latex_
 
 
+def _sort_by_num_models(var):
+    "return integer for Variable sorting"
+    return len(v.key.models) if v.key.models else 0
+
+
+def _sort_by_name_and_idx(var):
+    "return tuplef for Variable sorting"
+    return (var.key.str_without(["units", "idx"]), var.key.idx)
+
+
 class ConstraintSet(list):
     "Recursive container for ConstraintSets and Inequalities"
     def __init__(self, constraints, substitutions=None, recursesubs=True):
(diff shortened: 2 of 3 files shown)


gpkit/constraints/set.py
```diff
@@ -5,12 +5,6 @@
 from ..repr_conventions import _str, _repr, _repr_latex_
 
 
-def _var_sort_key(var):
-    "return thing to compare for Variable sorting"
-    k = var.key
-    return (k.str_without(["units", "idx"]), k.idx)
-
-
```
whoburg: this fixed #724; I want to retain some functionality (not necessarily `__getitem__` for getting a correctly-sorted VectorVariable from a model.

gpkit/constraints/set.py
```diff
@@ -44,8 +38,11 @@ def __getitem__(self, key):
             if len(variables) == 1:
                 return variables[0]
             else:
-                variables.sort(key=_var_sort_key)
-                return variables
+                sortfn = lambda v: len(v.key.models) if v.key.models else 0
```
whoburg: http://stackoverflow.com/questions/25010167/e731-do-not-assign-a-lambda-expression-use-a-def

gpkit/constraints/set.py
```diff
@@ -44,8 +38,11 @@ def __getitem__(self, key):
...
+                variables.sort(key=sortfn)
+                variable = variables[0]
+                # note: doesn't work for vector variables
```
whoburg: what is the plan for VectorVariables?

gpkit/constraints/set.py
```diff
@@ -44,8 +38,11 @@ def __getitem__(self, key):
...
+                return variable
```
whoburg: shouldn't we also retain the old `__getitem__` functionality, that returned multiple search hits, in a new method?

bqpd: Yes, can't do it today.

gpkit/constraints/set.py
bqpd: Probably they should be stored as arrays, not by their individual varkeys; but that's a issue which predates this and should have a new issue.

gpkit/constraints/set.py
bqpd: Indeed.

gpkit/constraints/set.py
bqpd: Great, probably can't do it until this evening.

gpkit/constraints/set.py
bqpd: Done.

gpkit/constraints/set.py
bqpd: Done.

gpkit/constraints/set.py
bqpd: Done in `variables_byname`

gpkit/constraints/set.py
bqpd: issue #799

gpkit/constraints/set.py
```diff
@@ -5,6 +5,16 @@
 from ..repr_conventions import _str, _repr, _repr_latex_
 
 
+def _sort_by_num_models(var):
+    "return integer for Variable sorting"
+    return len(v.key.models) if v.key.models else 0
```
whoburg: possible premature optimization, but would it be faster to not lookup `v.key.models` twice?

```python
def _sort_by_num_models(var):
    "return integer for Variable sorting"
    mods = v.key.models
    return len(mods) if mods else 0
```

.pylintrc
```diff
@@ -61,7 +61,7 @@ confidence=
 # --disable=W"
 #disable=import-star-module-level,old-octal-literal,oct-method,print-statement,unpacking-in-except,parameter-unpacking,backtick,old-raise-syntax,old-ne-operator,long-suffix,dict-view-method,dict-iter-method,metaclass-assignment,next-method-called,raising-string,indexing-exception,raw_input-builtin,long-builtin,file-builtin,execfile-builtin,coerce-builtin,cmp-builtin,buffer-builtin,basestring-builtin,apply-builtin,filter-builtin-not-iterating,using-cmp-argument,useless-suppression,range-builtin-not-iterating,suppressed-message,no-absolute-import,old-division,cmp-method,reload-builtin,zip-builtin-not-iterating,intern-builtin,unichr-builtin,reduce-builtin,standarderror-builtin,unicode-builtin,xrange-builtin,coerce-method,delslice-method,getslice-method,setslice-method,input-builtin,round-builtin,hex-method,nonzero-method,map-builtin-not-iterating
 
-disable=superfluous-parens,locally-disabled,misplaced-comparison-constant,too-few-public-methods,no-self-use
+disable=superfluous-parens,locally-disabled,misplaced-comparison-constant,too-few-public-methods,no-self-use,invalid-sequence-index
```
whoburg: hmm, but a model is a sequence, because it acts like a list, right? Where does this occur? Shouldn't we just disable locally?

whoburg:
coming along; 2nd round review complete.

gpkit/constraints/set.py
bqpd: Heh, sure thing.

.pylintrc
bqpd: You can use strings to get variables out of models, and I don't want a pylint error every single time that happens...

Issue #789, opened 2016-08-04
another simple cvxopt unknown example


whoburg:
For future analysis, here is a relatively simple but perfectly valid Model that cvxopt cannot solve with default settings.

```python
"""Commercial Aircraft Model loosely inspired by TASOPT"""
import numpy as np
from gpkit import Variable, Model


rho = Variable("\\rho", 0.4, "kg/m^3", "air density")
a = Variable("a", 300, "m/s", "speed of sound")
Wtemp = Variable("W_deleteme", 1, "lbf")
e = Variable("e", 0.85, "-", "Oswald efficiency")
S = Variable("S", 1300, "ft^2", "wing area")
A = Variable("A", 9.45, "-", "Aspect ratio")
Wwing = Variable("W_{wing}", "lbf", "wing weight")
CD = Variable("C_D", "-", "drag coefficient")
CL = Variable("C_L", "-", "lift coefficient")
Wfuel = Variable("W_{fuel}", 36864.9, "lbf", "fuel weight")
W = Variable("W", "lbf", "total weight")
V = Variable("V", "knots", "true airspeed")
T = Variable("T", "lbf", "thrust")
M = Variable("M", "-", "Mach number")
We = Variable("W_e", 84749.8, "lbf", "empty weight")
Wpay = Variable("W_{pay}", 38700, "lbf", "payload weight")
Wdry = Variable("W_{dry}", "lbf", "dry weight")


constraints = [W <= 0.5*rho*V**2*CL*S,
               T >= 0.5*rho*V**2*CD*S,
               Wwing >= Wtemp,
               CD >= 0.02 + CL**2/np.pi/e/A,
               W >= Wdry + Wfuel,
               M == V/a,
               Wdry >= We + Wpay + Wwing,
              ]

m = Model(T/V, constraints)
sol = m.solve()
```

As far as I can tell there is nothing wrong with this model. Here are two solver changes that cause it to solve:

 - solve with Mosek (`sol = m.solve(solver="mosek")`)
 - solve with cvxopt's ldl kktsolver (`sol = m.solve(kktsolver="ldl")`)

And for analyzing the cvxopt failure, here are some model changes, each of which causes cvxopt to solve the model with default settings:

 - change the first inequality to an equality (`W == 0.5*rho*V**2*CL*S,`)
 - remove the isolated constraint `M == V/a`
 - Increase the constant value of Wtemp to 10000
[...]

PR #795, opened 2016-08-04
use ldl kkt solver by default for cvxopt

whoburg:
The ldl solver is much better at handling rank issues. See #789.


whoburg:
@bqpd, ready for review and merging.

bqpd:
Looks great to me, go ahead and merge.

whoburg:
@mayork, note this means that with the latest gpkit you should not need to set the kktsolver for cvxopt (it's now done by default)

PR #813, opened 2016-08-11
added error raise

aechevarria1:
I added a check in the substitutions file so any nomial using a substitution method will go through the check. I added the check to the sublinplace function in the costed constraint file so that the substitutions a model tries to use is checked when using subinplace. There are also other changes to the file that I didn't do, I may have pulled an out of date version of the files on accident so they are showing up in the commit as well.


whoburg:
I'm closing this without merging; feel free to reopen if there is a path planned to making it pass unit tests.

PR #804, opened 2016-08-06
change tolerance for mosek on Ubuntu16

whoburg:
e-mail chain leading to this (newest to oldest):

```
Units don’t affect that at all; looks like most of the tolerances for mosek default to 1e-8 or smaller, although MSK_DPAR_INTPNT_NL_TOL_REL_GAP  defaults to 1e-6.
http://docs.mosek.com/7.0/capi/The_optimizers_for_continuous_problems.html#ch-lin-opt-tab-nlo-term-cri

I’ll begrudgingly try turning ndigits on that test to 6.

On Aug 6, 2016, at 9:31 AM, Marshall C. Galbraith wrote:

I presume the solver is iterative and has a convergence tolerance. How do unit impact that tolerance?
From: Warren W Hoburg
Sent: Saturday, August 06, 2016 9:27 AM
To: Marshall C. Galbraith
Cc: Edward (Ned) Burnell
Subject: Re: Ubuntu 16

How can we possibly be passing test_call (which has a 10 digit tolerance!) and failing test_call_units?


    def test_call(self):
        A = Variable('A', '-', 'Test Variable')
        prob = Model(A, [A >= 1])
        sol = prob.solve(verbosity=0)
        self.assertAlmostEqual(sol(A), 1.0, 10)

    def test_call_units(self):
        # test from issue541
        x = Variable("x", 10, "ft")
        y = Variable("y", "m")
        m = Model(y, [y >= x])
        sol = m.solve(verbosity=0)
        self.assertAlmostEqual(sol("y")/sol("x"), 1.0)
        self.assertAlmostEqual(sol(x)/sol(y), 1.0)


On Aug 6, 2016, at 9:26 AM, Marshall C. Galbraith wrote:

The pint version is printed as part of the script right before the tests are run.

Yeah looks like your 7 places is right on the threshold. Either relaxing the tolerance on the test or tightening the tolerance on the solver should fix this.

Thanks,
Marshall
From: Warren W Hoburg
Sent: Saturday, August 06, 2016 9:22 AM
To: Marshall C. Galbraith
Cc: Edward (Ned) Burnell
Subject: Re: Ubuntu 16

Hmm, I ran the test below on my machine and it also fails. I don’t understand the ndig cutoff, but at least it’s consistent.

Not sure why this answer comes out further from 1.0 on this particular platform. Probably the best solution is for us to relax the cutoff slightly.


##### test I just ran on my machine

import unittest


class TestUnittestFramework(unittest.TestCase):

[...]

Issue #541, opened 2016-03-03
solution array __call__ should return units


whoburg:
When a user does `sol("W")`, the output should have units. Otherwise, incorrect computations involving mismatched units could be done by an unsuspecting user.

MWE:
```python
from gpkit import Model, Variable

x = Variable("x", 10, "ft")
y = Variable("y", "m")
m = Model(y, [y >= x])
sol = m.solve(verbosity=0)
print sol("y")/sol("x")
```
(should print 1.0; prints `0.304799996606`)

PR #815, opened 2016-08-16
Add colorbar to sensitivitymap

aechevarria1:
Color bar only for paintby = "constants", didn't implement for paintby = "monomials" or "posynomials"


aechevarria1:
The fail seems to be due to the same issue as referenced in issue #541

whoburg:
I believe the failing test is actually due to #804. Rebasing this should therefore fix that issue. I'll do a rebase now and force push.

gpkit/interactive/sensitivity_map.py
```diff
@@ -7,7 +7,7 @@
 BLUE = np.array([16, 131, 246])/255.0
 RED = np.array([211, 24, 16])/255.0
 GRAY = 0.7*np.ones(3)
-
+#color_list = []
```
whoburg: remove this commented out line of code?

gpkit/interactive/sensitivity_map.py
```diff
@@ -29,6 +29,7 @@ def colorfn(senss):
 def signomial_print(sig, sol, colorfn, paintby="constants", idx=None):
     "For pretty printing with Sympy"
     mstrs = []
+    color_list = []
```
whoburg: wouldn't it make more sense for this to be a set instead of list?  

http://www.python-course.eu/sets_frozensets.php

`colorstrs = set()`

gpkit/interactive/sensitivity_map.py
```diff
@@ -38,6 +39,7 @@ def signomial_print(sig, sol, colorfn, paintby="constants", idx=None):
                          if var in sol["sensitivities"]["constants"]
                          else 0.0)
                 colorstr = colorfn(senss)
+                color_list.append(colorstr)
```
whoburg: `colorstrs.add(colorstr)`

gpkit/interactive/sensitivity_map.py
```diff
@@ -78,9 +80,9 @@ def signomial_print(sig, sol, colorfn, paintby="constants", idx=None):
             idx += 1
             colorstr = colorfn(senss)
             mstrs_.append("\\textcolor%s{%s}" % (colorstr, mstr))
-        return " + ".join(sorted(mstrs_)), idx
+        return (" + ".join(sorted(mstrs_)), idx)
     else:
-        return " + ".join(sorted(mstrs))
+        return (" + ".join(sorted(mstrs)), color_list)
```
whoburg: would be good to document return types for this method in the docstring

gpkit/interactive/sensitivity_map.py
```diff
@@ -119,8 +122,8 @@ def solution(self):
             return self.model.solution
 
     def _repr_latex_(self):
-        "iPython LaTeX representation."
-        return "$$\\require{color}\n"+self.latex+"\n$$"
+        "iPython LaTeX representation"
+        return "$$\\require{color}\n" +self.latex+"\n$$"+ self.color_list
```
whoburg: This change doesn't seem right to me. This adds a python list to a (latex) string? Has this been tested?

gpkit/interactive/sensitivity_map.py
```diff
@@ -147,13 +150,19 @@ def constraint_latex_list(self, paintby):
...
                 right = signomial_print(constr.right, sol, colorfn, paintby)
                 constr_tex = ("    & %s \\\\" %
-                              (left + constr.latex_opers[constr.oper] + right))
+                              (left[0] + constr.latex_opers[constr.oper]
+                               + right[0]))
+                for j in right[1]+left[1]:
+                    num_tup = (j[6:10], j[11:15], j[16:-1])
+                    if num_tup != ('0.70', '0.70', '0.70') \
+                                  and num_tup not in tup_list:
+                        tup_list.append(num_tup)
```
whoburg: I'm not sure I follow this. Is there a simpler / cleaner way to accomplish this?

whoburg:
Initial code review complete. @aechevarria1 you can respond to comments if discussion is needed and/or address them by making changes and pushing additional commits to this branch.

Also, I suggest you post a screenshot of what the output looks like with the new colorbar!

gpkit/interactive/sensitivity_map.py
aechevarria1: So the general process I used was to use a list to get the colors from signomial print, then add them to a color list in the sensitivity map code, then sort that list so the color bar is sorted. This is why I used a list. However, I could make this first color list a set, and then use something like this to add those components to a list that can be sorted later if you prefer. However, the colors need to be in a list eventually so they can be sorted.
``` python
for i in range(len(p)):
    print p.pop()
```

gpkit/interactive/sensitivity_map.py
aechevarria1: I'll make sure to add that

gpkit/interactive/sensitivity_map.py
aechevarria1: Sorry if it seems confusing, I change the color list into a string at a point in the code, so that is actually a string being appended to the return

gpkit/interactive/sensitivity_map.py
aechevarria1: I'll look into a better way to do this

aechevarria1:
This is currently what the color bar looks like. Let me know if you want anything to be changed
color bar

bqpd:
Sorry to butt in, but what about a smooth vertical gradient in the left column?

image

Issue #775, opened 2016-07-29
Installation instructions > installing cvxopt


tonystao:
CVXOPT package installation instructions on gpkit.readthedocs is incomplete. Need to install BLAS and LAPACK via: 

`sudo apt-get install libblas-dev liblapack-dev`

before running `sudo pip install cvxopt`. 
(on Ubuntu)

whoburg:
Does @galbramc agree?

galbramc:
Yes that is correct. We have those packages installed on all our testing machines for other codes, so I didn't realize they were needed for cvxopt as well.

whoburg:
@aechevarria1, this should be fairly easy to add to the docs.

whoburg:
this was fixed by #934

PR #817, opened 2016-08-16
added extra cvxopt installation step

aechevarria1:
Did I put this in the right place? Also is it fine to have Mac and Linux be separated?

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -65,9 +65,13 @@ For Windows 10, cvxopt does not appear to be an option.
 Installing cvxopt
 -----------------
 
-Mac OSX and Linux
+Mac OSX
     Run ``pip install cvxopt``
-    
+
+Linux
+    Run ``sudo apt-get install libblas-dev liblapack-dev``
+    then ``sudo pip install cvxopt``
+
     If experiencing issues with wheel in Ubuntu 16.04, try the `official installer. <http://cvxopt.org/install/index.html>`_
 
 Windows


aechevarria1:
Same unit tests keep failing, with reference to #541

galbramc:
test this please

galbramc:
If you want to install cvxopt in your system python, you need sudo for the mac as well. The problem is if you install gpkit with sudo, then the ~/.gpkit belongs to root and you don't have permissions to use it. This means someone really can't install gpkit in the system python...

docs/source/installation.rst
```diff
@@ -65,9 +65,13 @@ For Windows 10, cvxopt does not appear to be an option.
...
 -----------------
 
-Mac OSX and Linux
+Mac OSX
     Run ``pip install cvxopt``
-    
+
+Linux
+    Run ``sudo apt-get install libblas-dev liblapack-dev``
+    then ``sudo pip install cvxopt``
```
whoburg: @aechevarria1, I recommend addressing @galbramc's point by removing `sudo ` from both lines above.

whoburg:
Review complete. Once ready to merge, this PR will close #775.

galbramc:
No you have to use sudo with apt-get, but the pip install is the question. I don't know how to take a stance on if you should recommend to users to install things in the system, or push them towards installing things in virtualenvs. That is something you have to decide. Regardless, installing gpkit in the system, as things currently are, does not actually work correctly because of the permission problems with ~/.gpkit.

galbramc:
If you don't have a virtualenv, then we actually need a sudo command for OSX as well. I just don't know how to formulate recommending to users if they should install everything in their system, or if we should urge them to use virtualenvs.

aechevarria1:
I looked through a lot of popular python packages and their installation processes, and it seems that the majority of them use `pip install`. I didn't notice any of the packages having instructions for use of virtualenvs

bqpd:
Yeah, I think they generally assume that if you're installing in a virtualenv you'll know how to set it up yourself.

The permissions issue could be fixed just by chmod'ing `.gpkit`...

aechevarria1:
So are we going with `pip install` for both mac and linux and `sudo apt-get ...` for linux only?

whoburg:
This can be closed without merging once #934 is merged. Resolution to this was 9a5bb4d833e22a5ba775fba1b6487b25b488b098.

PR #831, opened 2016-09-02
convert united solutions back to original units

### LINKED ISSUE #809: incorrect units in results table
units in results table are incorrect despite being correct in the SolutionArray. MWE:

```python
from gpkit import Variable, Model, units

T = Variable("T", "N", "thrust")
Tmin = Variable("T_{min}", "N", "minimum thrust")

M = Model(T, [T >= Tmin])
M.substitutions.update({Tmin: 1000*units.lbf})
SOL = M.solve()
print SOL.table()
print 'SOL("T_{min}") is %s' % (SOL("T_{min}"),)
```
printout from the above is:
```
Using solver 'cvxopt'
Solving for 1 variables.
Solving took 0.00301 seconds.

Cost
----
 4448 [N] 

Free Variables
--------------
T : 4448  [N] thrust

Constants
---------
T_{min} : 1000  [N] minimum thrust

Sensitivities
-------------
T_{min} : 1 minimum thrust

SOL("T_{min}") is 1000 force_pound
```

bqpd:
closes #809

### DIFF
--- a/gpkit/small_classes.py
+++ b/gpkit/small_classes.py
@@ -142,8 +142,7 @@ def _enray_and_unit_dict(d_in, d_out, unitless_keys=(), united=False):
             if hasattr(v[0], "units"):
                 # if the first element of the list alrady has units,
                 # ensure unit consistency across the entire list
-                qty = Quantity(1, v[0].units)
-                v = [e.to(v[0].units).magnitude for e in v]*qty
+                v = [e.to(k.units).magnitude for e in v]*k.units
             else:
                 v = np.array(v)
                 if (united and hasattr(k, "units")
--- a/gpkit/small_classes.py
+++ b/gpkit/small_classes.py
@@ -140,7 +140,7 @@ def _enray_and_unit_dict(d_in, d_out, unitless_keys=(), united=False):
                                             unitless_keys, united)
         else:
             if hasattr(v[0], "units"):
-                # if the first element of the list alrady has units,
+                # if the first element of the list already has units,
                 # ensure unit consistency across the entire list
                 v = [e.to(k.units).magnitude for e in v]*k.units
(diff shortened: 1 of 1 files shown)


bqpd:
@whoburg, ready for review

whoburg:
Can we back up a moment prior to review:

I'm trying to understand how #809 occurred. I would have thought that the values (and units) printed in the solution table would be pulled from the SolutionArray. How did these get out of sync?

whoburg:
ok, I dug a little deeper and see that the table printing code prints the units of the varkey without regard to the units of the value.

This fix seems like a good fix for now. But perhaps in a future table printing refactor we should use the value for units printing, or at least double-check that the units agree?

Planning to merge this when tests pass.

whoburg:
this is failing pylint because of warnings fixed in another PR (which is frustrating, but currently a known issue). I think we can safely merge because Jenkin's gpkit_Push job doesn't know about those fixes yet.

@galbramc

galbramc:
Yeah, should be safe to merge if that is the case. You will still get pylint warnings until the other PR is merged unless we reset the counts. Zero warnings would of course solve this problem :P.

whoburg:
I know it would, and I don't mean to keep beating a dead horse, but in my opinion continuous integration should enable and promote continuous improvement. Our goal is to have zero pylint warnings, so we don't want to disable any warnings, but it's simply not practical for us to focus all our effort on fixing all the warnings at once. And the current setup discourages fixing pylint warnings in PRs that will take a long time to merge, to the point where I've sometimes undone those fixes.

Certainly not @galbramc's fault or problem, comment above is directed at Jenkins, they should allow comparing the warning count across different jobs, so that we could compare each PR to master.

PR #833, opened 2016-09-02
fix #832

### LINKED ISSUE #832: substitutions strangely persisting
```python
import gpkit
from gpkit.constraints.tight import TightConstraintSet as TCS

gpkit.SIGNOMIALS_ENABLED = True

x = gpkit.Variable("x")
y = gpkit.Variable("y")
ymax = gpkit.Variable("y_{max}", 0.1)


m = gpkit.Model(x, [x >= 0.5, TCS([x >= 1-y, y <= ymax])])
m.substitutions[ymax] = 0.2
m.localsolve()

m = gpkit.Model(x, [x >= 0.5, TCS([x >= 1-y, y <= ymax])])
m.localsolve() # should end with a y of 0.1, instead has 0.2
```

bqpd:
@whoburg ready for review

### DIFF
--- a/gpkit/tests/t_sub.py
+++ b/gpkit/tests/t_sub.py
@@ -138,6 +138,18 @@ class TestNomialSubs(unittest.TestCase):
 class TestGPSubs(unittest.TestCase):
     """Test substitution for Model and GP objects"""
 
+    def test_persistence(self):
+        x = gpkit.Variable("x")
+        y = gpkit.Variable("y")
+        ymax = gpkit.Variable("y_{max}", 0.1)
+
+        with gpkit.SignomialsEnabled():
+            m = gpkit.Model(x, [x >= 1-y, y <= ymax])
+            m.substitutions[ymax] = 0.2
+            self.assertAlmostEqual(m.localsolve(verbosity=0)["cost"], 0.8, 3)
+            m = gpkit.Model(x, [x >= 1-y, y <= ymax])
+            self.assertAlmostEqual(m.localsolve(verbosity=0)["cost"], 0.9, 3)
+
     def test_united_sub_sweep(self):
         A = Variable("A", "USD")
         h = Variable("h", "USD/count")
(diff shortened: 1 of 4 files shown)


bqpd:
Pylint errors aren't new and are in an unchanged file.

bqpd:
@whoburg, waiting for review.

gpkit/tests/t_sub.py
```diff
@@ -138,6 +138,18 @@ def test_signomial(self):
...
+        x = gpkit.Variable("x")
+        y = gpkit.Variable("y")
+        ymax = gpkit.Variable("y_{max}", 0.1)
+
+        with gpkit.SignomialsEnabled():
+            m = gpkit.Model(x, [x >= 1-y, y <= ymax])
+            m.substitutions[ymax] = 0.2
+            self.assertAlmostEqual(m.localsolve(verbosity=0)["cost"], 0.8, 3)
+            m = gpkit.Model(x, [x >= 1-y, y <= ymax])
+            self.assertAlmostEqual(m.localsolve(verbosity=0)["cost"], 0.9, 3)
```
whoburg: I'm confused/worried about something.

I tried simplifying the test to a simple GP:
```python
x = Variable("x")
y = Variable("y")
ymin = Variable("y_{min}", 0.1)
m = Model(x, ConstraintSet([x >= y, y >= ymin]))
```
But this does not reproduce the issue (on master). I'm confused why this issue occurs for `SignomialProgram` objects but not plain `Model` objects, since both inherit from `CostedConstraintSet`. Is it possible that, although the change to CostedConstraintSet fixes this issue, there is an additional problem somewhere in SP that we should address?

whoburg:
review complete, see one concern re: why this only happens for SPs.

gpkit/tests/t_sub.py
bqpd: Possibly! Yeah I also noticed that it wasn't working with the GP.

bqpd:
Created issue #836

PR #835, opened 2016-09-05
common code for x0 defaults in SPs

whoburg:
This should make things cleaner on the `equality` branch.

ready for review.

### DIFF
--- a/gpkit/nomials/nomial_math.py
+++ b/gpkit/nomials/nomial_math.py
@@ -693,12 +693,20 @@ class SignomialInequality(ScalarSingleEquationConstraint):
 
     def _fill_default_x0(self, x0, varkeys):
         """For all keys in varkeys, updates x0 with default values and
-        substitutions. Returns x0."""
+        substitutions. Returns x0.
+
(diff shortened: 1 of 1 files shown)


gpkit/nomials/nomial_math.py
```diff
@@ -693,12 +693,20 @@ def as_posyslt1(self):
...
+
+        Order of precedence for x0 default:
+            - substitution value
+            - x0 value
+            - sp_init value
+            - 1.0
+        """
         if x0 is None:
-            x0 = {vk: vk.descr["sp_init"] for vk in varkeys
-                  if "sp_init" in vk.descr}
```
whoburg: Note that `sp_init` was not getting used for partial subs - I think this was probably a bug

bqpd: Good point.

bqpd:
Looks great to me, but it's bringing up pylint bugs fixed in another branch. Should we merge?

bqpd:
retest this please

PR #845, opened 2016-09-06
exclude variables by name as well as by string

bqpd:
closes https://github.com/hoburg/gpkit-models/issues/79


bqpd:
@whoburg, ready for review

bqpd:
TODO: add `include_only` to this, and remove their effect later; have them only occur with names. This should be able to happen with set logic

bqpd:
@whoburg, waiting for review

PR #855, opened 2016-09-07
Update from gpkit.GP --> gpkit.Model in ipynb example

pgkirsch:
A really trivial change, just updating the syntax in the `simpleaircraft.ipynb` example. Ready for a 2-secon review @bqpd @whoburg.


bqpd:
It no longer has the plots, though?

pgkirsch:
Ah my bad... here it is with executed cells, but I have kind of annoying warnings.. any idea how to switch those off?

bqpd:
If you run the cell twice they'll disappear. :-/

pgkirsch:
I made a mess of this..

pgkirsch:
Wow that was an embarrassing amount of effort for such a dumb fix. Should be good now?

bqpd:
Looks great! Man, I wrote that thing so long ago... @whoburg, remember `monify_up`?

whoburg:
classic, `monify_up` belongs in a gpkit museum.

whoburg:
ps thanks for fixing this @pgkirsch.

simpleaircraft.ipynb should perhaps become an inline readthedocs example instead of an ipynb (could be copy/pasted to an ipynb trivially). @pgkirsch, from your experience showing GPkit to people, what's your take on ipynb examples vs inline examples in readthedocs?

pgkirsch:
Hmmm I don't really have a strong preference either way and they both serve a purpose. Inline ones can look really slick, ipynb is more dynamic so people can see it in action and maybe relate to it in that way.

Issue #666, opened 2016-04-17
no upper bound warning triggered unexpectedly


whoburg:
This is a MWE derived from one of the 'call an external solver' models. The result is `C_D has no upper bound` and a solver RankError, which is surprising since `CD` is the objective.

``` python
from gpkit import Variable, Model
from gpkit.constraints.set import ConstraintSet

CD = Variable("C_D", "-", "Drag Coefficient")
cdmin = Variable("cdmin", 0.5)

class SimpleCDModel(ConstraintSet):

    def as_posyslt1(self):
        raise TypeError("SimpleCDModel is not allowed to solve as a GP.")

    def as_gpconstr(self, x0):
        return (CD >= cdmin)

constraints = [SimpleCDModel([])]

m = Model(CD, constraints)
sol=m.localsolve()
```

bqpd:
Works with `constraints = [SimpleCDModel([]), cdmin >= 0.1]`. The problem is introducing a new constant in the middle of a solve; it doesn't look for a value.

bqpd:
Still should print out a _lower_ bounds warning instead of the upper bound warning.

bqpd:
The _feasibility programs_ are what's missing an upper bound for CD, and correctly printing this out. So this isn't a bug, just a potentially unwanted feature...

bqpd:
@whoburg, @pgkirsch, @mayork, @mjburton11, I'm tempted to make feasibility models not print bounds warnings. What sayest thou?

pgkirsch:
I agree with this because SPs often give misleading bound warnings in the feasibility solves.

On Sep 9, 2016, at 20:00, bqpd <notifications@github.com<mailto:notifications@github.com>> wrote:


@whoburg<https://github.com/whoburg>, @pgkirsch<https://github.com/pgkirsch>, @mayork<https://github.com/mayork>, @mjburton11<https://github.com/mjburton11>, I'm tempted to make feasibility models not print bounds warnings. What sayest thou?

whoburg:
Hm, I'm less convinced. Bound warnings are often super useful for debugging Linking errors (including for SPs, which may go straight to feasibility solve mode)

I haven't seen the misleading bound warnings @pgkirsch mentions. Are we sure they were not the result of one of the many bugs we recently fixed?

I'm still not clear on why the example at the top of this ticket triggers a feasibility solve of any kind. The `as_gpconstr` approximate version seems primal feasible.

pgkirsch:
[...]

Issue #563, opened 2016-03-09
freeing fixed variables


whoburg:
Is it possible to "free" a variable in an existing model, such that its `value` is removed, and the corresponding model, when solved, treats that variable as a free variable?

In an existing model `m`, I tried `m["C_L"].key.descr.pop("value")`, but it doesn't actually remove the value from the descr. I also don't know if other changes would be needed to `Model` to implement this robustly.

... or maybe there's a much smarter way to accomplish converting fixed variables in an existing model to free variables?

bqpd:
In the refactor, you can just remove it from the substitutions dict. :) All values have already been incorporated into it.

bqpd:
This should be a test!

bqpd:
This works, just needs to be tested and documented.

bqpd:
It's used in the new primal feasibility models, so just needs to be documented.

whoburg:
this is nitpicky, but I think this also deserves a unit test beyond the indirect test as part of primal feasibility models (whose implementation may change in the future)

bqpd:
Also, my bad, it's actually not used in the new feasibility models but should be.

bqpd:
actually, my bad, it is.

whoburg:
is this unit tested somewhere? I don't consider the issue closed until it is.

bqpd:
It's implicitly tested everywhere we use RelaxConstants...

whoburg:
I still feel that this functionality would benefit from a focused unit test. This would document exactly what we expect to work (as a user, I don't currently know how to use this feature).

bqpd:
We test this in t_model/test_cost_freeing

bqpd:
It's also in the docs at https://gpkit.readthedocs.io/en/latest/advancedcommands.html#freeing-fixed-variables

Issue #521, opened 2016-02-24
feasibility does not support monomial equality constraints


whoburg:
Just made a naive attempt at using our feasibility tools on an infeasible model containing monomial equality constraints:

``` python
from gpkit import Variable, Model
x = Variable("x")
m = Model(x, [x == 3, x == 4])
# m.solve("mosek")  # raises a RuntimeWarning!
feas = m.feasibility()
```

The current output is:

``` python
AttributeError                            Traceback (most recent call last)
<ipython-input-14-c0ff50aa3d6a> in <module>()
      3 m = Model(x, [x == 3, x == 4])
      4 # m.solve("mosek")  # raises a RuntimeWarning!
----> 5 feas = m.feasibility()

/Users/whoburg/MIT/dev/gpkit/gpkit/model.pyc in feasibility(self, search, constvars, verbosity)
    603             raise
    604         else:
--> 605             return self._feasibility("gp", search, constvars, verbosity)
    606 
    607     def localfeasibility(self,

/Users/whoburg/MIT/dev/gpkit/gpkit/model.pyc in _feasibility(self, programtype, search, constvars, verbosity)
    628 
    629         if "overall" in search:
--> 630             m = feasibility_model(self, "max")
    631             m.substitutions = allsubs
    632             infeasibility = m._solve(programtype, None, verbosity, False)["cost"]

/Users/whoburg/MIT/dev/gpkit/gpkit/feasibility.pyc in feasibility_model(program, flavour, varname, constants, signomials, programType)
     49         constraints = ([1/slackvar] +  # slackvar > 1
     50                        [constraint/slackvar  # constraint <= sv
---> 51                         for constraint in constraints])
     52         prog = programType(cost, constraints)
     53 

/Users/whoburg/MIT/dev/gpkit/gpkit/nomials.pyc in __div__(self, other)
    399             return Signomial(self.exps, self.cs/other)
    400         elif isinstance(other, Monomial):
--> 401             return other.__rdiv__(self)
    402         elif isinstance(other, PosyArray):
    403             return np.array(self)/other

/Users/whoburg/MIT/dev/gpkit/gpkit/nomials.pyc in __rdiv__(self, other)
    500         "Divide other by this Monomial"
    501         if isinstance(other, Numbers + (Signomial,)):
--> 502             return other * self**-1
    503         else:
    504             return NotImplemented

/Users/whoburg/MIT/dev/gpkit/gpkit/nomials.pyc in __mul__(self, other)
[...]

Issue #520, opened 2016-02-24
feasibility example from docs does not run


whoburg:
The feasibility example at http://gpkit.readthedocs.org/en/latest/advancedcommands.html#feasibility-analysis does not currently run. The error below comes from the uneven division `PosyArray(m.signomials)/feas["constraints"]`

current output:

``` python
ValueError                                Traceback (most recent call last)
<ipython-input-7-edeafb6f1253> in <module>()
     13 # USING CONSTRAINTS
     14 m = Model(x, [x <= x_max, x >= x_min])
---> 15 m.constraints = PosyArray(m.signomials)/feas["constraints"]
     16 m.solve()
     17 

ValueError: operands could not be broadcast together with shapes (3,) (2,) 

```

whoburg:
- [ ] Let's also make sure the `docs/` feasibility example(s), like this one, get run as part of unit tests, and
 - [ ] unit test feasibility_model with flavour="constants", as suggested by @galbramc due to a recent coverage report.

galbramc:
So I can easily run the files docs/source/examples/*.py, but how do I execute the examples in docs/source/advancedcommands.rst?

bqpd:
superseded / closed by the new feasibility models

whoburg:
The feasibility docs are still out of date and not run as part of the unit tests.

bqpd:
They're run as part of the unit tests in the relevant PR...

whoburg:
👍

PR #856, opened 2016-09-07
Feasibility models

bqpd:
at long last!

depends on #854
- [x] write a `Model.feasibility()` method that reports on both primal and dual feasibility
  - @galbramc I'm afraid this will require another circular import...we can meet in person sometime and go over why I think this is an OK (well, python-essential) design pattern. :p
- [x] update the include/exclude logic in `RelaxConstants` to share with `LinkedConstraints`: see #845


bqpd:
@mjburton11, @mayork, @pgkirsch, try out the `Model.debug` method on this branch when your models don't solve, try to break it.

bqpd:
we should test the new debug tools with SPs, as well.

bqpd:
@galbramc macys has been hanging for 15m

galbramc:
Was it hanging or just backlogged with lots of builds in the queue? Did it hang while trying to clone the repo?

galbramc:
Sure you want to disable cyclic imports? What about other people that are not as experienced or aware of the problems of cyclic imports?

bqpd:
oh, I didn't check when it was hanging, sorry.

I think that the bad kind of cyclic imports are fairly obvious, because they happen when imports are either not working, at the end of a file, or in a function called during `gpkit.__init__`, so they'll be an easy thing to spot in code review.

bqpd:
closes #666

.pylintrc
```diff
@@ -61,7 +61,7 @@ confidence=
 # --disable=W"
 #disable=import-star-module-level,old-octal-literal,oct-method,print-statement,unpacking-in-except,parameter-unpacking,backtick,old-raise-syntax,old-ne-operator,long-suffix,dict-view-method,dict-iter-method,metaclass-assignment,next-method-called,raising-string,indexing-exception,raw_input-builtin,long-builtin,file-builtin,execfile-builtin,coerce-builtin,cmp-builtin,buffer-builtin,basestring-builtin,apply-builtin,filter-builtin-not-iterating,using-cmp-argument,useless-suppression,range-builtin-not-iterating,suppressed-message,no-absolute-import,old-division,cmp-method,reload-builtin,zip-builtin-not-iterating,intern-builtin,unichr-builtin,reduce-builtin,standarderror-builtin,unicode-builtin,xrange-builtin,coerce-method,delslice-method,getslice-method,setslice-method,input-builtin,round-builtin,hex-method,nonzero-method,map-builtin-not-iterating
 
-disable=superfluous-parens,locally-disabled,misplaced-comparison-constant,too-few-public-methods,no-self-use,invalid-sequence-index
+disable=superfluous-parens,locally-disabled,misplaced-comparison-constant,too-few-public-methods,no-self-use,invalid-sequence-index,cyclic-import
```
whoburg: Really hoping to avoid doing this. Everything I read about cyclic imports is that they often indicate it's time to refactor. For a long time I've been complaining about the confusing inheritance among `Model`, `ConstraintSet`, and `CostedConstraintSet`. I think fixing the circular imports issue is a chance to clean that up, and disabling this will make us ignore it. GPkit's use of inheritance is fairly well structured, let's not give up on making it nice at the top level.

docs/source/advancedcommands.rst
```diff
@@ -1,34 +1,12 @@
 Advanced Commands
 *****************
 
-Feasibility Analysis
+Relaxation Analysis
```
whoburg: I think we should leave it as "Feasibility Analysis", not necessarily because it's better, but because it's standard in the field (see e.g. Boyd's GP tutorial paper section 3.1).

I'd also be supportive of "Primal feasibility analysis".

docs/source/advancedcommands.rst
```diff
@@ -1,34 +1,12 @@
...
 ====================
 
 If your Model doesn't solve, you can automatically find the nearest feasible version of it with the ``Model.feasibility()`` command, as shown below. The feasible version can either involve relaxing all constraints by the smallest number possible (that is, dividing the less-than side of every constraint by the same number), relaxing each constraint by its own number and minimizing the product of those numbers, or changing each constant by the smallest total percentage possible.
```
whoburg: text above still references `Model.feasibility()` and may be out of date.

Also, a sidenote: for real applications, I think we should promote putting the top-level specs in a named constraint set (e.g. "Specifications" or "Requirements"). This way, it will be super easy to only relax the top-level requirements (range, payload, takeoff length, etc) while not relaxing physical relationships that aren't appropriate to relax. This could make a nice example! It also has the benefit of removing the "choice" of what variable should be the objective; instead, set targets for each of the requirements/specifications, and always solve a Feasibility formulation (which will do better than the targets if they were feasible to begin with).

Sorry for the long comment, just brain dumping

docs/source/examples/relaxation.py
```diff
@@ -0,0 +1,38 @@
...
+print "Original model"
+print "=============="
+print m
+print
+# m.solve()  # raises a RuntimeWarning!
+
+print "With constraints relaxed equally"
+print "================================"
+from gpkit.constraints.relax import RelaxAll
+relaxall = RelaxAll(m)
```
whoburg: I'm not convinced that RelaxAll will be appropriate to use for any real Model. There are going to be some constraints (e.g. equalities) that are simply not appropriate to relax.

I'm advocating this because I'm thinking it might make more sense for the relaxation to happen at the ConstraintSet level, instead of the Model level. And I think this might help with the circular imports.

docs/source/examples/relaxation.py
```diff
@@ -0,0 +1,38 @@
...
+print
+
+print "With constants relaxed individually"
+print "==================================="
+from gpkit.constraints.relax import RelaxConstants
+relaxedconstants = RelaxConstants(m)
+relaxedconstants.cost *= m.cost**0.01  # add a bit of the original cost in
+print relaxedconstants
+print relaxedconstants.solve(verbosity=0).table()  # brings x_min down to 1.0
+print
```
whoburg: I bet `RelaxConstants` will be super useful! But I again think it might be more appropriate at the ConstraintSet (i.e. submodel) level. Of could this can include making a ConstraintSet of the original model.

docs/source/examples/simpleflight_output.txt
```diff
@@ -43,8 +43,6 @@ Sensitivities
      W_{W_{coeff1}} : 0.2903  Wing Weight Coefficent 1        
             N_{ult} : 0.2903  ultimate load factor            
      W_{W_{coeff2}} : 0.1303  Wing Weight Coefficent 2        
-             (CDA0) : 0.09156 fuselage drag area              
-                \mu : 0.08599 viscosity of air
```
whoburg: I thought we dropped this change? (it's showing up in #856). should we rebase that PR?

gpkit/constraints/model.py
```diff
@@ -108,3 +108,51 @@ def subconstr_latex(self, excluded=None):
         "The collapsed appearance of a ConstraintBase"
         if self.name:
             return "%s_{%s}" % (self.name, self.num)
+
+    def debug(self):
+        "Attempts to diagnose infeasible models."
+        from .relax import RelaxConstants, RelaxConstraints
+        from .bounded import Bounded
+
```
whoburg: check solver status here? Mosek, at least, will give a `PRIM_INFEAS_CER` for many primal infeasible models, in which case there's no sense in doing the Bounded formulation...

gpkit/constraints/model.py
```diff
@@ -108,3 +108,51 @@ def subconstr_latex(self, excluded=None):
...
+            iterator = enumerate(zip(relaxvals, feas[0][0]))
+            for i, (relaxval, constraint) in iterator:
+                if relaxval >= 1.01:
+                    relax_percent = "%i%%" % ((relaxval-1)*100)
+                    print ("  %i: %4s relaxed  (as posynomial: %s <= 1)"
+                           % (i, relax_percent, constraint.right))
+        except (ValueError, RuntimeWarning):
+            print ("Model does not solve with relaxed constraints.")
+        # NOTE: If the cost has a very strong relationship to feasibility,
+        #       the self.cost**0.01 component above could be the problem.
```
whoburg: great inline

gpkit/constraints/relax.py
```diff
@@ -0,0 +1,140 @@
+"""Models for assessing primal feasibility"""
+from .model import Model
+from .set import ConstraintSet
+from ..nomials import Variable, VectorVariable, parse_subs, NomialArray
+
+
+class RelaxAll(Model):
```
whoburg: maybe make it a method of `Model`, and implement it by forming a `RelaxConstraints` and changing the cost? Or cut entirely?

gpkit/constraints/relax.py
```diff
@@ -0,0 +1,140 @@
...
+        substitutions = dict(constraints.substitutions)
+        posynomials = constraints.as_posyslt1()
+        self.relaxvar = Variable("C", models=["Relax"],
+                                 modelnums=[Model._nums["Relax"]])
+        Model.__init__(self, self.relaxvar,
+                       [[posy <= self.relaxvar
+                         for posy in posynomials],
+                        self.relaxvar >= 1],
+                       substitutions,
+                       name=False)
```
whoburg: why doesn't this use `RelaxConstraints`?

gpkit/constraints/relax.py
```diff
@@ -0,0 +1,140 @@
...
+    relaxvars : Variable
+        The variables controlling the relaxation. A solved value of 1 means no
+        relaxation was necessary or optimal for a particular constraint.
+        Higher values indicate the amount by which that constraint has been
+        made easier: e.g., a value of 1.5 means it was made 50 percent easier
+        in the final solution than in the original problem.
+
+    [Boyd2007] : "A tutorial on geometric programming", Optim Eng 8:67-122
+
+    """
```
whoburg: great docstring!!

gpkit/constraints/relax.py
```diff
@@ -0,0 +1,140 @@
...
+        if not isinstance(constraints, ConstraintSet):
+            constraints = ConstraintSet(constraints)
+        substitutions = dict(constraints.substitutions)
+        posynomials = constraints.as_posyslt1()
+        N = len(posynomials)
+        self.relaxvars = VectorVariable(N, "C", models=["Relax"],
+                                        modelnums=[Model._nums["Relax"]])
+        Model.__init__(self, self.relaxvars.prod(),
+                       [[posynomials <= self.relaxvars],
+                        self.relaxvars >= 1],
```
whoburg: I think there are some cases where we won't want the `>= 1` constraint. E.g. if I set targets for my requirements, it's nice to find out how much I can exceed them by.

gpkit/constraints/relax.py
```diff
@@ -0,0 +1,140 @@
...
+    Attributes
+    ----------
+    relaxvars : Variable
+        The variables controlling the relaxation. A solved value of 1 means no
+        relaxation was necessary or optimal for a particular constant.
+        Higher values indicate the amount by which that constant has been
+        made easier: e.g., a value of 1.5 means it was made 50 percent easier
+        in the final solution than in the original problem. Of course, this
+        can also be determined by looking at the constant's new value directly.
+    """
```
whoburg: great docstring!

gpkit/constraints/relax.py
```diff
@@ -0,0 +1,140 @@
...
+    def __init__(self, constraints, include_only=None, exclude=None):
+        exclude = frozenset(exclude) if exclude else frozenset()
+        include_only = frozenset(include_only) if include_only else frozenset()
+        if not isinstance(constraints, ConstraintSet):
+            constraints = ConstraintSet(constraints)
+        substitutions = dict(constraints.substitutions)
+        constants, _, _ = parse_subs(constraints.varkeys,
+                                     constraints.substitutions)
```
whoburg: smells like duplicate code w/ LinkedConstraintSet ?

gpkit/constraints/relax.py
```diff
@@ -0,0 +1,140 @@
...
+        self.relaxvars, relaxation_constraints = [], []
+        self.num = Model._nums["Relax"]
```
whoburg: I think we need `NamedConstraintSet`...

gpkit/constraints/signomial_program.py
```diff
@@ -107,7 +109,11 @@ def localsolve(self, solver=None, verbosity=1, x0=None, rel_tol=1e-4,
             try:
                 result = gp.solve(solver, verbosity-1, **kwargs)
             except (RuntimeWarning, ValueError):
-                nearest_feasible = feasibility_model(gp, "max")
+                feas_constrs = ([slackvar >= 1] +
+                                [posy <= slackvar
+                                 for posy in gp.posynomials[1:]])
```
whoburg: odd that this doesn't use the `constraints.relax` formulations?

whoburg:
This is great! Initial review complete.

gpkit/constraints/relax.py
bqpd: Ahaha.

docs/source/examples/relaxation.py
bqpd: That really won't help with the circular imports, but I like the idea.

gpkit/constraints/model.py
bqpd: Yup, that's definitely an option if the model has solved before.

gpkit/constraints/relax.py
bqpd: And then just substitute out all the relaxvars?

gpkit/constraints/relax.py
bqpd: But what if you can exceed them only by letting sometime else relax? And a lot of times I think that'll be dual-infeasible. I think that's a different usecase and requires a different framing.

gpkit/constraints/relax.py
bqpd: the include_only/exclude code? Sure, it's a design pattern that I want to standardize.

gpkit/constraints/signomial_program.py
bqpd: Slower, circularer :p

gpkit/constraints/relax.py
whoburg: or perhaps RelaxConstraints should be capable of producing this formulation as a special case, e.g. by a boolean keyword arg?

The gp_tutorial lists two different versions of the RelaxConstraints formulation (equations 10 and 11).

gpkit/constraints/relax.py
whoburg: yep

gpkit/constraints/relax.py
whoburg: I disagree but time will tell. @mjburton11 is going to try the feasibility formulation (with a single shared slack variable) for the 1682 model. We will see how it goes. He will not be able to use this with the >= 1 constraint, but it's so simple he's planning to just hand code for now.

The plan is a SystemRequirements class that looks something like the following (constant target variable definitions omitted)
```python
class SystemPerformanceTargets(ConstraintSet):
s = Variable("s", "-", "shared requirement margin")
constraints = [R*s >= R_target
               t_loiter*s >= t_loiter_target,
               W_pay*s >= W_pay_target,
               P_pay *s >= P_pay_target,
               MTOW <= s*MTOW_target]
```

gpkit/constraints/relax.py
bqpd: But for now it's only four lines in common...

gpkit/constraints/relax.py
bqpd: Oh, I think that'll work fine. But in general I think relaxing _every_ constant in a model without the `<= 1` constraint would cause dual infeasibility, and could confuse users.

gpkit/constraints/signomial_program.py
bqpd: (mostly the former. Reforming all those posynomials is kinda slow)

bqpd:
closes #520

whoburg:
@bqpd speaking: self.cost**0.001 is assuming a monomial cost.

whoburg:
input should be able to be a ConstraintSet

.pylintrc
bqpd: Ack, we forgot to go over this yesterday. I'm really pretty sure we'll want to do this.

galbramc: You should keep in mind that while @bqpd is very familiar with which cyclic imports are dangerous and which ones are not, new students do not necessarily have that experience. Keeping this in will help educate them and likely lead toward clearer and cleaner code. That's my 2 cents.

bqpd: I don't want them either, there's just no real alternative beside sticking all of our code in one file (eww) or using non-local imports (ewwwwww). Python 3 thankfully does better compiling and bypasses this issue entirely.

Y'all are welcome to propose alternatives, or to live with the pylint warnings, but I really don't see another way to do it.

docs/source/examples/relaxation.py
bqpd: After talking yesterday, I think we're agreed that the use case you're thinking of is for now better done manually?

docs/source/examples/relaxation.py
bqpd: After talking yesterday, I think we're agreed that the use case you're thinking of is for now better done manually?

gpkit/constraints/relax.py
bqpd: After talking yesterday, I think we're agreed that the use case you're thinking of is for now better done manually?

gpkit/constraints/relax.py
bqpd: After talking yesterday, I think we're agreed that the use case you're thinking of is for now better done manually?

docs/source/advancedcommands.rst
bqpd: I really like this design pattern!

bqpd:
@whoburg, ready for re-review

bqpd:
closes #563

bqpd:
@whoburg, ready for re-review

bqpd:
closes #521

bqpd:
@whoburg, I added a test for this, but it doesn't work with cvxopt during testing...even though it works outside it. A little weird.

galbramc:
Looking at the output, the problem is that you are printing the execution time in debug_output.txt.

```
--- a/docs/source/examples/debug_output.txt
+++ b/docs/source/examples/debug_output.txt
@@ -1,7 +1,7 @@
 Debugging...
 Using solver 'cvxopt'
 Solving for 6 variables.
-Solving took 0.222 seconds.
+Solving took 0.0255 seconds.
 Model does not solve with bounded variables or relaxed constants.
```

This guarantees that the output will change every time you run the tests, and the trick of using ```git status``` to make sure the examples are still producing the same output will always fail. I only check for this with cvxopt, which is why mosek does not show this as a failure (even though all output files are changed with mosek).

bqpd:
That's part of the problem, but the bigger one is that the feasibility
solves should work; the rest of the printout is different from what happens
when I run the script by hand (with any solver)

galbramc:
What do you mean by 'run by hand'. Jenkins actually runs the example explicitly with
```python /home1/jenkins/workspace/gpkit_PullRequest_Unit/buildnode/reynolds/optimizer/cvxopt/docs/source/examples/debug.py```

.pylintrc
whoburg: proposed alternative: why not make the feasibility models inherit from `CostedConstraintSet`?

docs/source/examples/relaxation.py
whoburg: there was a miscommunication above; I was confused by the names. My point was that usually it doesn't make sense to relax all the constraints.

I think the current naming does not make clear the available distinctions, namely:

 - one might want to relax only a subset of the constraints rather than all the constraints
 - one might want the relaxation variables to be shared or independent (currently named `RelaxAll` and `RelaxConstraints`, respectively.

docs/source/examples/relaxation.py
whoburg: Still feel these are most natural as ConstraintSets or CostedConstraintSets rather than Models.

gpkit/constraints/relax.py
whoburg: I thought my use case would be supported under one of the available feasibility models. But it can certainly be hand-coded easily.

gpkit/constraints/relax.py
whoburg: no problem

gpkit/constraints/relax.py
whoburg: I think this would allow inheritance from `CostedConstraintSet`?

galbramc:
test this please

bqpd:
@whoburg, CostedConstraintSet does not have a solve method; inheriting from it really doesn't make sense.

bqpd:
re: names, any alternatives?

whoburg:
Re: `CostedConstraintSet`, why do the feasibility models need a solve method?

Re names, I had proposed earlier using the terms in the literature (see section 3.1 of Boyd's GP tutorial) -- this would suggest the name `FeasibilityProblem` or similar -- something with "Feasibility" in the name.

docs/source/advancedcommands.rst
bqpd: Done.

docs/source/advancedcommands.rst
bqpd: Fixed stale reference, saved your comment above to figure out what to do with later.

bqpd:
@whoburg, I think this is ready to merge, if it passes tests...I'll let those sit overnight.

whoburg:
I can do a more thorough review, but first:

I didn't see a response to the suggestion that these inherit from CostedConstraintSet. This would eliminate the need for circular imports, and matches our design pattern used elsewhere of requiring users to wrap a `Model` around a constraint set in order to call solve.

We could have a helper function that does the full feasibility analysis and solving -- it would be a method of Model -- but I don't see why these implementations need to inherit from Model.

bqpd:
These can't inherit from `CostedConstraintSet` because only `Model` has the `.solve` method. These implementations need to inherit from `Model` because the user will be treating them as models...

whoburg:
maybe we can clear up my confusion with a quick discussion soon

bqpd:
Sounds good!

bqpd:
retest this please

bqpd:
retest this please

bqpd:
@whoburg, this is ready for review.

While Relaxed now inherit from ConstraintSet, they still have the circular dependency temporarily; the reason for this is that `master` stores the model/modelnum lookup in `Model`, while newvectorized stores it at the global level, and so changing this branch to do that would have caused annoyances when merging newvectorized. If we can merge this now, I'll make the necessary changes on `newvectorized`, release v0.4.3, and then we can pull that to master and remove the dependency.

bqpd:
Or I could merge newvectorized first, and push out v0.4.3 with both sets of changes.

bqpd:
superseded by #953

PR #861, opened 2016-09-08
reduce number of keys to make keydicts faster

bqpd:
I tried doing something fancy, but this worked better. :-/


bqpd:
@whoburg, ready for review

PR #858, opened 2016-09-07
fast subinplace

bqpd:
after profiling @mjburton11's code.

depends on #854

PR #854, opened 2016-09-07
don't link substitutions, require them for as_posyslt1()

bqpd:
I decided that substitutions actually shouldn't be linked, and this branch implements that in the proper fashion. Should close #827, removes some ugly lines of code...

Oh, and it fixes an unrelated bug @mjburton11 found today as well.

@galbramc, there's some pylint errors in this that I couldn't get to disappear. The no-member one is silly, since it's doing isinstance on `ConstraintSet`, not `list`, but I couldn't get it to go away with a comment. The circular import one is less silly (it's a mild hack) but I can't get it to go away either.

PR #873, opened 2016-09-09
substitutions keys must be in varkeys

### LINKED ISSUE #848: Check substitutions dict for phantom varkeys
It's easy to forget to update the substitutions dict when changing a variable's varkey, or even to just type it in wrong. It might therefore be nice to have some kind of check to ensure that every entry in the substitutions dict is a varkey in the model that it's being used for. Sorry, I know there are concerns about model building speed and this won't help but it shouldn't be a particularly hard fix.

e.g.
```python
rho = Variable('\\rho')
...

substitutions = {
    ...
    'rho': 1.225,
    ...
}

m = Model(objective, constraints, substitutions)
# Should return an error, something like "The varkey 'rho' does not appear in model"
```
This will at least prompt a user to check the varkey for rho and notice it is wrong.

bqpd:
closes #848


bqpd:
@pgkirsch, see if this meets your needs!

whoburg:
@bqpd, this seems to be getting commits from another PR, should it be rebased?

bqpd:
Naw, squashing will fix all that. (but yeah I probably shoulda rebased)

pgkirsch:
Works perfectly, thanks!

bqpd:
@whoburg, ready for review

gpkit/constraints/set.py
```diff
@@ -55,8 +55,19 @@ def __init__(self, constraints, substitutions=None):
                 if hasattr(self[i], "substitutions"):
                     self.substitutions.update(self[i].substitutions)
         self.reset_varkeys()
-        self.substitutions.update({self[k].key: v for k, v in subs.items()
-                                   if k in self.varkeys})
```
whoburg: let's keep the deleted lines above, commented out, with a note about timing? (I didn't realize the alternative implementation would be so many more lines. Totally open to moving back to @bqpd's original short and sweet version as long as it's not slow.)

gpkit/keydict.py
```diff
@@ -48,7 +49,23 @@ def update(self, *args, **kwargs):
...
+                    if key.veckey:
+                        key = key.veckey
+                    elif len(keys) > 1:
+                        raise ValueError("substitution key '%s' was ambiguous;"
+                                         " .variables_byname('%s') will show"
+                                         " which variables it may refer to."
+                                         % (key, key))
+                else:
+                    raise KeyError("key '%s' does not refer to any varkey in"
+                                   " this ConstraintSet" % key)
```
whoburg: what an appropriate use of `KeyError`

gpkit/constraints/set.py
bqpd: It goes back to being short and sweet in #873 :)

whoburg:
ok, I reviewed commit cac69dce0495145b07521ed2a0f50265f77ccaf3 and it looks good. Now realizing I probably reviewed out of order; this is not ready to merge because it includes the commits from #858, which depends on #854.

Getting there...

bqpd:
Naw, this doesn't include #858, and I already merged #854...

whoburg:
The commits tab (which has 6 commits) is pretty confusing then...

PR #884, opened 2016-09-10
faster GPs, among other things

### LINKED ISSUE #819: have m.program.gps[x].result be able to print a table
this could make debugging a lot more user friendly

### LINKED ISSUE #837: units in solution table printouts
A couple proposed improvements to make result table printouts even safer:
1. (carry-over from #831): when looping through key, value pairs in `SolutionArray.results_table`, use the value instead of key for units printing, or at least double-check that the units agree.
2. When gpkit's units are disabled (`if not gpkit.units`), don't print units in results tables. It's misleading.

If item 2 above is implemented, we can remove the `if not gpkit.units` conditional from `test_units_sub` in `t_solution_array`.

bqpd:
closes #837 
closes #819 

Speed up @pgkirsch's airplane model by a factor of about 2.


bqpd:
@whoburg, ready for review

gpkit/tests/t_sub.py
```diff
@@ -175,7 +175,7 @@ def test_united_sub_sweep(self):
                                 h: 35*gpkit.units("USD"),
                                 Q: ("sweep", [50, 100, 500])})
         firstcost = m.solve(verbosity=0)["cost"][0]
-        self.assertAlmostEqual(firstcost, 1760, 3)
+        self.assertAlmostEqual(1760*gpkit.units("USD")/firstcost, 1, 5)
```
whoburg: 👍

gpkit/small_scripts.py
```diff
@@ -37,7 +42,8 @@ def unitstr(units, into="%s", options="~", dimless='-'):
         except ValueError:
             rawstr = "1.0 " + str(units.units)
         units = "".join(rawstr.replace("dimensionless", dimless).split()[1:])
-    return into % units if units else ""
+        return into % units if units else ""
+    return ""
```
whoburg: A suggestion for cleaning this up is on the `unitstrcleanup` branch. Please treat it like a PR and merge it into this branch if you like it.

gpkit/small_classes.py
```diff
@@ -93,6 +93,19 @@ def to_united_array(self, unitless_keys=(), united=False):
...
+
+
+def _classify(d_in, cls):
+    "Converts all contained dictionaries into the given class."
+    for k, v in d_in.items():
+        if isinstance(v, dict) and v:
+            if isinstance(v.keys()[0], Strings):
+                _classify(v, cls)
+            else:
+                d_in[k] = cls(v)
```
whoburg: I don't understand this implementation. It doesn't seem to do what the docstring says. What is the significance of Strings? Here is an example where I expected the subdict to get converted to an A but it didn't:

```python
class A(dict):
    def __repr__(self):
        return "A"

d = {5: {6: 7, "hi": 8}, "hi": 9}
_classify(d, A)
print d   # {'hi': 9, 5: {'hi': 8, 6: 7}}
```

gpkit/keydict.py
```diff
@@ -44,6 +45,8 @@ def __init__(self, *args, **kwargs):
     def update(self, *args, **kwargs):
         "Iterates through the dictionary created by args and kwargs"
         for k, v in dict(*args, **kwargs).items():
+            if hasattr(v, "copy"):
+                v = v.copy()
```
whoburg: this seems odd and non-dict-like. Why not let whoever is calling `update` copy their own values if they want/need to?

gpkit/geometric_program.py
```diff
@@ -201,14 +202,7 @@ def _get_solver(solver):
             print ("solution checking took %.2g%% of solve time" %
                    ((time() - tic) / soltime * 100))
 
-        ## Let constraints process the results
-        if hasattr(self.constraints, "process_result"):
-            self.constraints.process_result(self.result)
-        else:
-            for constraint in self.constraints:
-                if hasattr(constraint, "process_result"):
-                    constraint.process_result(self.result)
```
whoburg: (note to self: moved to `prog_factories`)

gpkit/constraints/signomial_program.py
```diff
@@ -121,13 +123,28 @@ def localsolve(self, solver=None, verbosity=1, x0=None, rel_tol=1e-4,
...
+        x0.update(self.substitutions)
+        return x0
+
     def gp(self, x0=None, verbosity=1):
         """Get a GP approximation of this SP at x0"""
+        x0 = self._default_x0(x0)
         gp = GeometricProgram(self.cost, self.as_gpconstr(x0),
                               self.substitutions, verbosity=verbosity)
-        gp.x0 = x0  # NOTE: SIDE EFFECTS
+        gp.x0 = KeyDict(x0)  # NOTE: SIDE EFFECTS
```
whoburg: should this be a `FastKeyDict` since it's just for storage?

gpkit/constraints/signomial_program.py
```diff
@@ -121,13 +123,28 @@ def localsolve(self, solver=None, verbosity=1, x0=None, rel_tol=1e-4,
...
+        x0 = KeyDict({vk: vk.sp_init for vk in self.varkeys if vk.sp_init})
+        x0.varkeys = self.varkeys  # check that keys are actually in the SP
+        if x0_in is not None:
+            x0.update(x0_in)
+        x0.update(self.substitutions)
+        return x0
+
     def gp(self, x0=None, verbosity=1):
         """Get a GP approximation of this SP at x0"""
+        x0 = self._default_x0(x0)
```
whoburg: maybe only do this if `x0 is None`? I believe our default in nomials is to not support a partially-filled x0; we could adopt that here too to avoid repeatedly creating the default each iteration.

gpkit/constraints/prog_factories.py
```diff
@@ -82,7 +82,13 @@ def solvefn(self, solver=None, verbosity=1, skipsweepfailures=False,
         solution.program = self.program
         solution.to_united_array(unitless_keys=["sensitivities"], united=True)
         if self.cost.units:
-            solution["cost"] *= self.cost.units
```
whoburg: `assert not hasattr(solution["cost"], "units")` just to be sure?

docs/source/examples/beam_output.txt
```diff
@@ -8,8 +8,8 @@ Free Variables
           dx : 1.2                                             [m]   Length of an element
      \vec{M} : [ 1.98e+03  1.27e+03  713       317      ... ]  [N*m] Internal moment     
      \vec{V} : [ 660       528       396       264      ... ]  [N]   Internal shear      
-\vec{\theta} : [ 0         0.177     0.285     0.341    ... ]        Slope               
-     \vec{w} : [ 0         0.106     0.384     0.759    ... ]  [m]   Displacement        
+\vec{\theta} : [  -        0.177     0.285     0.341    ... ]        Slope               
+     \vec{w} : [  -        0.106     0.384     0.759    ... ]  [m]   Displacement
```
whoburg: why did this change?

whoburg:
review complete.

gpkit/keydict.py
bqpd: Because I like `KeyDict(kd)` to act like a copy of `kd`, that's the only reason.

gpkit/constraints/signomial_program.py
bqpd: It's not, though; users may well access the `x0` values of the GP.

docs/source/examples/beam_output.txt
bqpd: Because of the new copies in KeyDict

gpkit/small_classes.py
bqpd:
```python
class A(dict):
    def __repr__(self):
        return "A"

d = {"a": {6: 7, "hi": 8}, "hi": 9}
_classify(d, A)
print d   # {"a": A, "hi": 9}
```

whoburg: ... and makes the assumption that all keys have the same type (due to the use of `[0]`) - docstring should probably state that assumption too.

Any risk for mixed key types in user substitution dicts?

docs/source/examples/beam_output.txt
whoburg: hmm, it is non obvious why copies in KeyDicts led to dashes in beam output.

I'm just trying to push for super clear APIs because things like this feel like fragile unexpected side effects.

gpkit/small_classes.py
bqpd: Nope, not anymore. Unless they replace instead of updating a substitution dict...

gpkit/keydict.py
bqpd: And I think most users will expect numpy arrays to copy...we could do this only for numpy arrays?

gpkit/keydict.py
```diff
@@ -170,3 +173,8 @@ def __getitem__(self, key):
     def __setitem__(self, key, value):
         "Assigns the key itself every time."
         KeyDict.__setitem__(self, key, None)
+
+
+class FastKeyDict(KeyDict):
+    "FsatKeyDicts are KeyDicts that don't map keys, only collate arrays"
```
whoburg: `"Fsat"`

docs/source/examples/beam_output.txt
bqpd: I can explain this one tomorrow in person! It changed at first because of the unexpected side effects caused by not copying...

gpkit/keydict.py
whoburg: As a developer I'm trying to understand when I can get away with using a `FastKeyDict` vs a `KeyDict`, but I'm not completely clear on that from this docstring.

bqpd: you're right. should be `"Fst"`

bqpd: `"FstKyDcts r KyDcts tht dn't mp kys, nly cllt rrys"`

bqpd: much faster that way.

bqpd: Re: when to use, yeah I should clarify.

Basically, KeyDicts have two features:
  1. access values with alternate keys
  2. collapse array-keyed values into the appropriate arrays

KeySets have only the first of those, FastKeyDicts have only the second.

gpkit/keydict.py
whoburg: given no other information I would expect `update` to work like this (no copy):

```python
In [1]: from numpy import array

In [2]: a = array([7, 3, 4])

In [3]: d = {"key": 7}

In [4]: d.update({"array": a})

In [5]: d["array"] is a
Out[5]: True
```

docs/source/examples/beam_output.txt
whoburg: ok

gpkit/keydict.py
whoburg: ok see that's really helpful, make that the docstring.

whoburg: and lol, now you're ready to read some aviation forecasts. 

https://www.aviationweather.gov/products/fa/boston

gpkit/small_classes.py
bqpd: Clarified docstrings

gpkit/keydict.py
bqpd: Resolved in `fastsubs`

gpkit/constraints/prog_factories.py
bqpd: thankfully posynomials don't copy on reference, so this should be fine.

gpkit/constraints/signomial_program.py
bqpd: done.

gpkit/keydict.py
bqpd: Edited the docstring

bqpd:
@whoburg, ready for re-review

bqpd:
retest this please

bqpd:
retest this please

bqpd:
@galbramc, it didn't start automatically testing the last commit.

galbramc:
Don't know why it didn't start...

galbramc:
Sorry, hit the wrong button

galbramc:
I'm updating Windows. You will have to wait for that before the test can run.

gpkit/small_scripts.py
whoburg: it passed all the tests and didn't change any of the examples...

whoburg: to make logistics simple I'll open a separate PR from that branch for further discussion

gpkit/keydict.py
```diff
@@ -5,24 +5,24 @@
 
 
 class KeyDict(dict):
-    """KeyDicts allow storing and accessing the same value with multiple keys
```
whoburg: For what it's worth I thought this first sentence was a great intro; I'd vote to keep it.

whoburg:
Does a rebase make sense here to fix the merge conflict? Curious how that will play out in the review system.

bqpd:
Given the bug that was there, a rebase wouldn't have fixed it.

How do I indicate I've responded to your review, @whoburg?

Also, normally I would be able to merge this as soon as it passed tests and then merge it into any other branches that needed it.

PR #887, opened 2016-09-10
remove stale patches and warnings

whoburg:
Removes an old pint patch (edit: actually, the patch is still needed)
Removes setup method warning
Removes out of date Modular ipynb

@bqpd, ready for review.


whoburg:
hmph, looks like https://github.com/hgrecco/pint/issues/366 is not actually fixed on 0.7.2.

bqpd:
So should we remove than unpatch?

whoburg:
yeah I'll rebase and drop the unpatch.

whoburg:
(force pushed)

whoburg:
ready for re-review

PR #893, opened 2016-09-12
KeyDict: failed __getitem__ causes state change

whoburg:
Found an issue where attempting to access a missing key from a KeyDict causes a state change where future updates and calls to `in` act incorrectly.

MWE is the attached test (also added as a commit to this PR):

```python
    def test_failed_getattr(self):
        kd = KeyDict()
        with self.assertRaises(KeyError):
            _ = kd["waldo"]
            # a failed __getitem__ causes potential state change (issue 893)
        # TODO uncomment next assertNotIn after getting rest of test working
        # self.assertNotIn("waldo", kd)
        kd.update({"waldo": 5})
        res = kd["waldo"]
        self.assertEqual(res, 5)
        self.assertIn("waldo", kd)
```

Issue can arise in interactive sessions when someone checks whether something is in a Models'
 substitutions dict, gets a KeyError, and then cannot update that key from then onwards.

### DIFF
--- a/gpkit/keydict.py
+++ b/gpkit/keydict.py
@@ -89,6 +89,7 @@ class KeyDict(dict):
         key, idx = self.parse_and_index(key)
         keys = self.keymap[key]
         if not keys:
+            del self.keymap[key]
             raise KeyError("%s was not found." % key)
         values = []
         for key in keys:
--- a/gpkit/keydict.py
+++ b/gpkit/keydict.py
@@ -89,7 +89,7 @@ class KeyDict(dict):
         key, idx = self.parse_and_index(key)
         keys = self.keymap[key]
         if not keys:
-            del self.keymap[key]
+            del self.keymap[key] # remove blank entry added due to defaultdict
             raise KeyError("%s was not found." % key)
         values = []
         for key in keys:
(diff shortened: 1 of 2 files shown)


whoburg:
@bqpd, priority issue.

whoburg:
Just checked, this issue exists back at 780750bdc7fd8070120c6312926dfec43253e5af (a month ago), so not a result of any of the recent activity.

bqpd:
@whoburg, fix ready for review

gpkit/keydict.py
```diff
@@ -89,6 +89,7 @@ def __getitem__(self, key):
         key, idx = self.parse_and_index(key)
         keys = self.keymap[key]
```
bqpd: because `keymap` is a defaultdict, this added a blank entry to `keymap[key]`

whoburg:
Looks great. Added one commit. And just confirmed that this does resolve the higher-level problem I observed with @mjburton11 today that led to finding this bug.

Planning to merge when all tests pass.

whoburg:
actually going to push one more cosmetic commit. Will squash all these.

whoburg:
@bqpd following good protocol, how about you merge this since I added commits.

bqpd:
Will do!

PR #906, opened 2016-09-20
simplify unitstr

whoburg:
minor simplifications. passes all tests, so if these changes are unacceptable let's add test(s) to catch the issue(s).

### DIFF
--- a/gpkit/small_scripts.py
+++ b/gpkit/small_scripts.py
@@ -24,21 +24,18 @@ def mag(c):
     return c.magnitude if isinstance(c, Quantity) else c
 
 
-def unitstr(units, into="%s", options="~", dimless='-'):
+def unitstr(units, into="%s", options="~", dimless="-"):
     "Returns the unitstr of a given object."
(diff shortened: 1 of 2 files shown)


whoburg:
ready for review.

galbramc:
test this please

galbramc:
So this breaks pint 0.6. Do you still want to support pint 0.6?

bqpd:
@whoburg

whoburg:
Ok, the pint 0.6 issue was that pint changed the way `Unit` objects are formatted from 0.6 to 0.7.2. The following illustrates the issue (independent of `unitstr`):

```python
from gpkit import units
print "{:~}".format(units("foot").units)
print "{:~}".format(units("nautical_mile").units)
```
On pint 0.6, the output is
```
foot
nautical_mile
```
On pint 0.7.2, the output is 
```
ft
nmi
```
Because of this, I think it's ok for our users to have different unit formatting depending on which version of pint they have installed. I updated the unit test accordingly.

Ready for re-review.

PR #908, opened 2016-09-21
Set minimum numpy version to 1.8.1

### LINKED ISSUE #907: macOS Sierra and numpy 1.8.0rc1
Well, looks like macOS Sierra comes with numpy 1.8.0rc1, and gpkit isn't happy with that. See https://acdl.mit.edu/csi/job/gpkit_PullRequest_Unit/buildnode=macys_Sierra,optimizer=cvxopt/1595/console.

@bqpd and @whoburg , not sure what you want to do about this. You could tell all your users they have to upgrade to a newer version of numpy. Either by upgrading their system numpy or using virtualenv's to install a newer version. Ideally you can fix gpkit so it works with 1.8.0rc1.

galbramc:
This closes #907

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -10,7 +10,7 @@ Install dependencies
 GPkit's dependencies are the python packages
 
 - ``pip``
-- ``numpy`` version 1.8 or newer
+- ``numpy`` version 1.8.1 or newer
 - ``scipy``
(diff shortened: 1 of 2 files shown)


galbramc:
Just noticed the documentation say 1.8, I'll update it to 1.8.1.

PR #915, opened 2016-09-28
Linked Variable Diagram

mayork:
I made removed all the hard coded things from @pgkirsch linking diagram script so now it should be easy to use.

@bqpd I think this is ready for review


mayork:
revoke ready for review statement..I'll fix pylint but the other problem is that the color list needs to be dynamically sized...do we want to keep repeating the same 9 colors or do something with RGB so each line is a different color?

bqpd:
For now let's rotate, but could you general an example graphic for a complex model (168, Engine, Commercial Aircraft) so we could see what that color rotation looks like?

mayork:
screen shot 2016-09-28 at 4 08 40 pm

it looks like this

bqpd:
Sounds good. Rotating colors looks fine to me; @whoburg, any preference?

pgkirsch:
Damn..that's a gnarly looking model. Still amazed you got that working in SP @mayork!

On Sep 28, 2016, at 13:38, bqpd <notifications@github.com<mailto:notifications@github.com>> wrote:


Sounds good. Rotating colors looks fine to me; @whoburg<https://github.com/whoburg>, any preference?

whoburg:
gnarly indeed - thanks @pgkirsch for the inspiration and proof of concept, and @mayork for generalizing it.

What about all black? If colors are going to collide anyway this might be the best option?

bqpd:
If you look at the dense left cluster of linked variables, I think all-black would be much more confusing than rotating colors, even though two of those end up as blue.

mayork:
@1ozturkbe

whoburg:
@bqpd, let's make sure this makes it into the release notes for the next version.

PR #917, opened 2016-10-01
Initial commit of plot sweep 1D

pgkirsch:
This is a function I wrote for aircraft sweeps and thought it might be a useful addition to the code base. Feel free to cannibalize as you see fit. Example usage:

``` python
In [1]: from gpkitmodels.aircraft.vtail import VerticalTail

In [2]: from gpkit.interactive.plotting import plot_sweep_1d

In [3]: import numpy as np

In [4]: vt = VerticalTail.standalone_737()

In [5]: plot_sweep_1d(vt, {'l_{fuse}': np.linspace(32, 40, 10)}, 'W_{vt}')
```
1. runs a 1D sweep
2. plots the result
3. replaces the substitution


pgkirsch:
Note that #918 needs to be fixed before this will work.

bqpd:
@pgkirsch, I believe #918 is not a problem with the current master branch?

bqpd:
Also, could you post an example plot made with this function / do you think it would be worth adding to the advanced commands (or a new 'Plotting'/'Interactive') section in the documentation?

whoburg:
retest this please

bqpd:
@pgkirsch, @whoburg  I believe this just has to be confirmed working on master / documented in this thread and on readthedocs to be merged.

gpkit/interactive/plotting.py
```diff
@@ -295,3 +295,36 @@ def plot_convergence(model):
...
+    # Check that dependent var exists
+    if dependentvar not in model.varkeys:
+        raise KeyError("{0} is not in model".format(dependentvar))
+
+    sweepvar = sweep.keys()[0]
+    if sweepvar in model.substitutions:
+        originalsubvalue = model.substitutions[sweepvar]
+    x = sweep.values()[0]
+    model.substitutions.update({sweepvar: ('sweep', x)})
+    sol = model.localsolve()
```
whoburg: This only works for SPs and fails for GPs. @bqpd, is there a recommended way for @pgkirsch to deal with this?

bqpd: Try to solve, and catch the InvalidGP error if it isn't a GP?

Or make this a ConstraintSet ;D

bqpd:
@pgkirsch, if you could write the docs, I'll get it working with solve/localsolve!

pgkirsch:
I can do that

On Dec 6, 2016, at 00:50, bqpd <notifications@github.com<mailto:notifications@github.com>> wrote:


@pgkirsch<https://github.com/pgkirsch>, if you could write the docs, I'll get it working with solve/localsolve!

pgkirsch:
@bqpd I added a brief section to the docs, with an example plot. The example is totally contrived but hopefully it is ok. Let me know if you would like anything else in the docs. (By the way I agree that docs might need to be reorganized a bit)

I also made a quick and dirty fix to the solve vs. local solve thing. Feel free to make better if you see fit.

pgkirsch:
@whoburg Added a solve/localsolve fix and added an example plot to the docs (but here it is as well)
plot_sweep_1d_example

docs/source/advancedcommands.rst
```diff
@@ -150,6 +150,22 @@ Example Usage
...
+-------------------
+
+A function called ``plot_sweep_1d`` has been created to facilitate creating, solving, and plotting the results of a single-variable sweep. Example usage is as follows:
+
+.. code-block:: python
+
+    import numpy as np
+    from gpkit.interactive.plotting import plot_sweep_1d
+    # m is an existing model, with variables x and y
+    f, ax = plot_sweep_1d(m, {'x': np.linspace(1, 2, 20)}, 'y')
```
whoburg: If `m` is simple enough it would be good to include it here so that the example is reproducible.

gpkit/interactive/plotting.py
```diff
@@ -295,3 +296,47 @@ def plot_convergence(model):
...
+
+    sweepvar = sweep.keys()[0]
+    if sweepvar in model.substitutions:
+        originalsubvalue = model.substitutions[sweepvar]
+    x = sweep.values()[0]
+    model.substitutions.update({sweepvar: ('sweep', x)})
+    try:
+        sol = model.solve()
+    except InvalidGPConstraint:
+        sol = model.localsolve()
```
whoburg: longer term I wonder if this should eventually be a method of `SolutionArray` (or accept one as an arg) instead of acting on a `Model`.

docs/source/advancedcommands.rst
bqpd: Yes pls!

gpkit/interactive/plotting.py
```diff
@@ -295,3 +296,47 @@ def plot_convergence(model):
...
     return fig, ax
+
+def plot_sweep_1d(model, sweep, dependentvar):
+    """Creates and plots a sweep from an existing model
+
+    Example usage:
+    f, _ = plot_sweep_1d(m, {'x': np.linspace(1, 2, 5)}, 'y')
+    f.savefig('mysweep.png')
+    """
+    # Check that sweep var exists
```
bqpd: How about `(key, val), = sweep.items()` to also assert that there's only one?

gpkit/interactive/plotting.py
bqpd: I think of this as just a convenience function, for when you don't want to be bothered to do the other three lines of code that would be necessary :p

gpkit/interactive/plotting.py
```diff
@@ -295,3 +296,47 @@ def plot_convergence(model):
...
+        xvarname = model[sweepvar].descr['name']
+    try:
+        yvarname = model[dependentvar].descr['label']
+    except KeyError:
+        yvarname = model[dependentvar].descr['name']
+    x_units = unitstr(model[sweepvar], into=" [%s] ", dimless="")
+    y_units = unitstr(model[dependentvar], into=" [%s] ", dimless="")
+    ax.set_xlabel(xvarname + x_units)
+    ax.set_ylabel(yvarname + y_units)
+    plt.show()
```
bqpd: In case it gets modified by the user after returning, I don't think you should show it here...

gpkit/interactive/plotting.py
```diff
@@ -295,3 +296,47 @@ def plot_convergence(model):
...
+        originalsubvalue = model.substitutions[sweepvar]
+    x = sweep.values()[0]
+    model.substitutions.update({sweepvar: ('sweep', x)})
+    try:
+        sol = model.solve()
+    except InvalidGPConstraint:
+        sol = model.localsolve()
+    y = mag(sol(dependentvar))
+    f, ax = plt.subplots()
+    ax.plot(x, y)
```
bqpd: Should the plot have some nice default stylings?

bqpd:
test models please

bqpd:
test models please

bqpd:
test models please

Issue #929, opened 2016-10-20
Post Solve ndarray error


bjarni177:
Optimal solution found.
Solving took 0.173 seconds.
result packing took 8.2% of solve time
solution checking took 1.6% of solve time
Solving took 4 GP solves and 1.59 seconds.
Traceback (most recent call last):
  File "/Users/Bjarni/Documents/GitHub/gpkit-models/gpkitmodels/aircraft/fuselage.py", line 664, in <module>
    sol = Fuselage.test(fuselage_type = 'narrowbody')
  File "/Users/Bjarni/Documents/GitHub/gpkit-models/gpkitmodels/aircraft/fuselage.py", line 546, in test
    return fu.localsolve(solver = None, verbosity=4)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/gpkit/constraints/prog_factories.py", line 92, in solvefn
    self.process_solution(solution)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/gpkit/constraints/set.py", line 287, in process_solution
    constraint.process_solution(sol)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/gpkit/constraints/set.py", line 287, in process_solution
    constraint.process_solution(sol)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/gpkit/constraints/tight.py", line 31, in process_solution
    negy = negy.sub(variables).value
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/gpkit/nomials/nomial_math.py", line 215, in sub
    _, exps, cs, _ = substitution(self, substitutions)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/gpkit/nomials/substitution.py", line 197, in substitution
    " of type '%s'" % (var, type(sub)))
TypeError: could not substitute w_{floor}_Fuselage with value of type '<type 'numpy.ndarray'>'
[Finished in 5.1s with exit code 1]

mayork:
this was on `feasmod` branch right?

bjarni177:
on the feassweep branch

mayork:
[...]

PR #932, opened 2016-10-21
Ndarray error

mayork:
fix to #929 

@bqpd


bqpd:
Since it's just one line of change I'll make it directly on the branch; thanks for pushing this!

PR #937, opened 2016-11-03
cleanup operator overloading and fix arrayconstraint subs/latex

bqpd:
@whoburg, ready for review

### DIFF
--- a/gpkit/constraints/single_equation.py
+++ b/gpkit/constraints/single_equation.py
@@ -41,16 +41,25 @@ class SingleEquationConstraint(object):
         if not excluded:
             excluded = ["units"]  # previously bool(self.left.units)
         latex_oper = self.latex_opers[self.oper]
-        return ("%s %s %s" % (self.left.latex(excluded), latex_oper,
-                              self.right.latex(excluded)))
+        latexleft = trycall(self.left, "latex", excluded, str(self.left))
+        latexright = trycall(self.right, "latex", excluded, str(self.right))
+        return ("%s %s %s" % (latexleft, latex_oper, latexright))
 
     def sub(self, subs):
         "Returns a substituted version of this constraint."
-        subbed = self.func_opers[self.oper](self.left.sub(subs),
-                                            self.right.sub(subs))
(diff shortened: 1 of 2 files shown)


whoburg:
@galbramc, we keep failing windows during `Receiving objects:` -- any ideas?

whoburg:
retest this please

galbramc:
The VM is having networking problems connecting to github. I can try rebooting the VM to see if that helps.

bqpd:
@whoburg, ready for review

gpkit/constraints/single_equation.py
```diff
@@ -41,16 +41,25 @@ def latex(self, excluded=None):
...
+        latexright = trycall(self.right, "latex", excluded, str(self.right))
+        return ("%s %s %s" % (latexleft, latex_oper, latexright))
 
     def sub(self, subs):
         "Returns a substituted version of this constraint."
-        subbed = self.func_opers[self.oper](self.left.sub(subs),
-                                            self.right.sub(subs))
+        subbedleft = trycall(self.left, "sub", subs, self.left)
+        subbedright = trycall(self.right, "sub", subs, self.right)
+        subbed = self.func_opers[self.oper](subbedleft, subbedright)
```
whoburg: I'm ok with this solution to the possibility that `self.left` or `self.right` does not have a `latex` attribute, but would it be cleaner to make sure that it does? Is it arrayconstraint objects that might not have `latex`? Should they have `latex` or inherit from something that does, defaulting as appropriate?

I'm thinking that making sure gpkit objects have `latex` might be more robust -- that way users who want to write scripts that access `latex` won't need to import `trycall` or write their own...

bqpd: Okay! Between this and another issue, I'm thinking the right solution might be to cast the left and right sides of a constraint to gpkit objects? (being signomials / nomial arrays of signomials) instead of staying 'raw'. Doing so would also fix this problem...

whoburg: related to #956

bqpd:
merging this PR for now, but I think #956 is the right way to handle this.

PR #938, opened 2016-11-03
drop support for operator overloading with pint Quantity objects

bqpd:
This isn't an easy choice to make, but the following didn't work on master with pint 0.7 and something had to be done:
```python
    def test_units(self):
        "make sure multiplication with units works (issue 492)"
        # have had issues where Quantity.__mul__ causes wrong return type
        m = 1.2 * gpkit.units.ft * Variable("x", "m")**2
        self.assertEqual(type(m), Monomial)
        if m.units:
            self.assertEqual(m.units, 1*gpkit.units.ft*gpkit.units.m**2)
        # also multiply at the end, though this has not been a problem
        m = 0.5 * Variable("x", "m")**2 * gpkit.units.kg
        self.assertEqual(type(m), Monomial)
        if m.units:
            self.assertEqual(m.units, 1*gpkit.units.kg*gpkit.units.m**2)
        # and with vectors...
        v = 0.5 * VectorVariable(3, "x", "m")**2 * gpkit.units.kg
        self.assertEqual(type(v), NomialArray)  # is actually a Quantity...
        if v.units:
            self.assertEqual(v.units, 1*gpkit.units.kg*gpkit.units.m**2) # ...of units 'kg'!
        v = 0.5 * gpkit.units.kg * VectorVariable(3, "x", "m")**2
        self.assertEqual(type(v), NomialArray)  # same here
        if v.units:
            self.assertEqual(v.units, 1*gpkit.units.kg*gpkit.units.m**2 # and here
```

I tried enhancing our pint subversion, but whatever two-step process pint does in its operator overloading to turn iterables into quantities is buried deep, and I didn't want to extend the rickety cantilever of our subversions any further into pint.

So what this PR does is make `gpkit.units`, by default, return `Monomials`. This means we don't have to subvert pint anymore to get operator overloading working, everything's already a gpkit object during Model creation.

However, this means that: 

  1. substitutions of united quantities will now be of monomials / nomialarrays instead
    - so I decided to catch that in keydict and just store them as quantities
  2. solutions will be Quantities that can't interact directly with gpkit objects
    - this isn't a problem for substitution, just when a signomial is multiplied by a solution / solution vector. I don't see a great way to handle this, but an error is better than the weird silent failures we had before.

### DIFF
--- a/gpkit/__init__.py
+++ b/gpkit/__init__.py
@@ -33,6 +33,7 @@ class GPkitUnits(object):
     def __init__(self):
         self.Quantity = ureg.Quantity  # pylint: disable=invalid-name
         if hasattr(ureg, "__nonzero__"):
+            # that is, if it's a DummyUnits object
             self.__nonzero__ = ureg.__nonzero__
             self.__bool__ = ureg.__bool__
 
--- a/gpkit/__init__.py
+++ b/gpkit/__init__.py
@@ -24,7 +24,23 @@ SIGNOMIALS_ENABLED = False
 
 # global variable initializations
 DimensionalityError = ValueError
-units = None
+ureg, units = None, None  # pylint: disable=invalid-name
+
+
+class GPkitUnits(object):
+    "Return monomials instead of Quantitites"
+
+    def __init__(self):
+        self.Quantity = ureg.Quantity  # pylint: disable=invalid-name
+        if hasattr(ureg, "__nonzero__"):
+            self.__nonzero__ = ureg.__nonzero__
+            self.__bool__ = ureg.__bool__
+
+    def __getattr__(self, attr):
+        return Monomial(self.Quantity(1, getattr(ureg, attr)))
+
+    def __call__(self, arg):
+        return Monomial(self.Quantity(1, ureg(arg)))
 
 
 def enable_units(path=None):
@@ -35,7 +51,7 @@ def enable_units(path=None):
 
     If gpkit is imported multiple times, this needs to be run each time."""
     # pylint: disable=invalid-name,global-statement
-    global units, DimensionalityError, UNIT_REGISTRY
+    global DimensionalityError, UNIT_REGISTRY, ureg, units
     try:
         import pint
         if path:
@@ -47,8 +63,10 @@ def enable_units(path=None):
             UNIT_REGISTRY.load_definitions(os_sep.join([path, "usd_cpi.txt"]))
             # next line patches https://github.com/hgrecco/pint/issues/366
             UNIT_REGISTRY.define("nautical_mile = 1852 m = nmi")
-        units = UNIT_REGISTRY
+
+        ureg = UNIT_REGISTRY
         DimensionalityError = pint.DimensionalityError
+        units = GPkitUnits()
     except ImportError:
         print("Optional Python units library (Pint) not installed;"
               " unit support disabled.")
@@ -71,7 +89,7 @@ def disable_units():
         from gpkit import disable_units
         disable_units()
     """
-    global units  # pylint: disable=global-statement
+    global ureg, units  # pylint: disable=invalid-name,global-statement
 
     class DummyUnits(object):
         "Dummy class to replace missing pint"
@@ -91,7 +109,8 @@ def disable_units():
         def __call__(self, arg):
             return 1
 
-    units = DummyUnits()
+    ureg = DummyUnits()
+    units = ureg
 
 enable_units()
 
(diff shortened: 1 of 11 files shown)


bqpd:
@whoburg, ready for review

whoburg:
We're having to do this due to https://github.com/hgrecco/pint/issues/256 -- `pint.Quantity` is not currently set up to play nicely with operator overloading in other objects / libraries.

whoburg:
retest this please

gpkit/__init__.py
```diff
@@ -24,7 +24,23 @@
...
 DimensionalityError = ValueError
-units = None
+ureg, units = None, None  # pylint: disable=invalid-name
+
+
+class GPkitUnits(object):
+    "Return monomials instead of Quantitites"
+
+    def __init__(self):
+        self.Quantity = ureg.Quantity  # pylint: disable=invalid-name
```
whoburg: I'm a bit confused about why this works -- isn't ureg `None`? (set 7 lines ago)

gpkit/__init__.py
```diff
@@ -24,7 +24,23 @@
...
+        if hasattr(ureg, "__nonzero__"):
```
whoburg: is the `if hasattr` due to changes in pint, or because `ureg` is sometimes not a pint object
? Inline comment explaining please.

gpkit/__init__.py
```diff
@@ -24,7 +24,23 @@
 
 # global variable initializations
 DimensionalityError = ValueError
-units = None
+ureg, units = None, None  # pylint: disable=invalid-name
+
+
+class GPkitUnits(object):
```
whoburg: should it inherit from `pint.unit.UnitRegistry`? This way we get any auxiliary methods/functionality that we aren't currently using but that pint super users might expect of a `ureg`

gpkit/keydict.py
```diff
@@ -116,6 +116,12 @@ def __setitem__(self, key, value):
                 emptyvec = np.full(key.shape, np.nan, **kwargs)
                 dict.__setitem__(self, key, emptyvec)
         for key in self.keymap[key]:
+            if getattr(value, "exp", None) is not None and not value.exp:
```
whoburg: why not `if hasattr(value, "exp") and not value.exp:`? I realize that's not quite equivalent -- is `value.exp` sometimes `None`?

gpkit/tests/t_nomials.py
```diff
@@ -182,13 +183,24 @@ def test_numerical_precision(self):
     def test_units(self):
         "make sure multiplication with units works (issue 492)"
         # have had issues where Quantity.__mul__ causes wrong return type
-        m = 1.2 * units.ft * Variable('x')**2
+        m = 1.2 * gpkit.units.ft * Variable("x", "m")**2
         self.assertTrue(isinstance(m, Monomial))
-        self.assertEqual(m.units, 1*units.ft)
+        if m.units:
```
whoburg: why does this need an if?

gpkit/tests/t_nomials.py
```diff
@@ -182,13 +183,24 @@ def test_numerical_precision(self):
...
+            self.assertEqual(m.units, 1*gpkit.ureg.kg*gpkit.ureg.m**2)
+        # and with vectors...
+        v = 0.5 * VectorVariable(3, "x", "m")**2 * gpkit.units.kg
+        self.assertTrue(isinstance(v, NomialArray))
+        if v.units:
+            self.assertEqual(v.units, 1*gpkit.ureg.kg*gpkit.ureg.m**2)
+        v = 0.5 * gpkit.units.kg * VectorVariable(3, "x", "m")**2
+        self.assertTrue(isinstance(v, NomialArray))
+        if v.units:
+            self.assertEqual(v.units, 1*gpkit.ureg.kg*gpkit.ureg.m**2)
```
whoburg: instead of all the if's how about an early `if not gpkit.units:  return`?

gpkit/tests/t_sub.py
```diff
@@ -171,8 +171,8 @@ def test_united_sub_sweep(self):
         Q = Variable("Q", "count")
         Y = Variable("Y", "USD")
         m = Model(Y, [Y >= h*Q + A/Q])
-        m.substitutions.update({A: 500*gpkit.units("USD"),
-                                h: 35*gpkit.units("USD"),
```
whoburg: I thought we said that subbing in `gpkit.units` would work?

gpkit/tests/t_vars.py
```diff
@@ -154,13 +154,13 @@ def test_init(self):
     def test_constraint_creation_units(self):
         v = VectorVariable(2, "v", "m/s")
         c = (v >= 40*gpkit.units("ft/s"))
-        c2 = (v >= [20, 30]*gpkit.units("ft/s"))
+        c2 = (v >= np.array([20, 30])*gpkit.units("ft/s"))
         if gpkit.units:
             self.assertTrue(c.right.units)
-            self.assertTrue(c2.right.units)
+            self.assertTrue(NomialArray(c2.right).units)
```
whoburg: should this casting perhaps happen automatically internally? (so that the RHS of an ArrayConstraint is never a raw array but is instead a NomialArray)

whoburg:
fixes #946

gpkit/__init__.py
bqpd: I don't think we want to encourage pint super users to think of this as a ureg...

gpkit/__init__.py
bqpd: This isn't run until `__init__` is called on line 97/115

gpkit/keydict.py
bqpd: Yup. For varkeys in particular. :-/ they still have `None` attributes for everything...

gpkit/tests/t_nomials.py
bqpd: in case units is disabled.

gpkit/tests/t_sub.py
bqpd: Oh, it does. I should revert this.

gpkit/tests/t_vars.py
bqpd: Quite possibly! I don't feel strongly about it, but would close the issue if you made one. :)

bqpd:
@galbramc, I've been waiting >15m for the dependency tests to run...it passed with the three previous commits, so I'm merging. :p

gpkit/__init__.py
whoburg: gotcha.

gpkit/tests/t_vars.py
whoburg: #956

PR #944, opened 2016-11-08
re-enable units after tests

whoburg:
one-liner change

on the current master, if you run `python -c "from gpkit.tests import run; run(); from gpkit import Variable; assert Variable('x', 'ft').units is not None"`

you get:
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
AssertionError
```

this fixes that.

@bqpd, ready for review


bqpd:
lol! ready to merge once tests pass.

whoburg:
oops, I rebase-merged instead of squash-merging, and as a result the PR number doesn't show up in `master`'s commit history. My bad, will get it right next time.

PR #954, opened 2016-11-24
preparation for version 0.5

bqpd:
- [x] add signup for mailing list and link to gitter on documentation site


whoburg:
timing question -- when should this get reviewed / merged?

bqpd:
This will be the branch that carries the pypi-uploaded version into master, so it'll be the last to be reviewed and merged!

whoburg:
@bqpd, can you try to clean up the figures in http://gpkit.readthedocs.io/en/latest/gpkit101.html prior to release? They're out of date.

Since we're linking to the docs in our release e-mail, I'm expecting a number of users might get inspired to read the docs top-to-bottom; it would be good if this section were in better shape.

bqpd:
@whoburg, the only error I can see in the diagrams is the use of `step()` instead of `gp()` in the last one. The whole section needs to be rewritten to emphasize our glorious constraintsets, but I don't think that'll happen before v0.5...I'll make the issue.

whoburg:
loving 1347b3c164705a69339545e182aa6c1884b35d40.

I'll review this tomorrow morning (or when ready, whichever is later)

bqpd:
@whoburg, ready for review. Passes install tests except for a MOSEK bug on Windows 10x64, which is in process of being debugged...

MANIFEST
```diff
@@ -1,37 +1,33 @@
-setup.py
 gpkit/__init__.py
 gpkit/_cvxopt.py
-gpkit/build.py
```
bqpd: Sorted alphabetically, to match github.

docs/source/installation.rst
```diff
@@ -113,7 +113,7 @@ Install GPkit
   - Run ``pip install gpkit`` at the command line (for system python installs, use ``sudo pip``)
   - Run ``pip install ipywidgets`` for interactive control of models (recommended)
   - Run ``python -c "import gpkit.tests; gpkit.tests.run()"`` to run the tests; if any tests do not pass, please email ``gpkit@mit.edu`` or `raise a GitHub issue <http://github.com/hoburg/gpkit/issues/new>`_.
-  - *Optional:* to install gpkit into an isolated python environment, install ``virtualenv``, run ``virtualenv $DESTINATION_DIR`` then activate it with ``source $DESTINATION_DIR/bin/activate``
```
bqpd: This didn't seem like the kind of advice we should give here. Open to it elsewhere in the docs.

galbramc:
Looks like the new version of gpkit has problems with the pip cache again. It's somewhat sporadic though so I don't understand it...

bqpd:
hmm, I'll test something.

bqpd:
just uploaded v0.5.01 to pypi, removing change to build.py that @whoburg
raised doubts about.

galbramc:
You know you really should use https://testpypi.python.org/pypi. That way you don't end up pushing multiple versions to the official site just to sort out these bugs.

galbramc:
We can setup jenkins to work with the testing site as well.

galbramc:
I fired off a couple of extra builds just to make sure that fixed things. Does this patch now only exit on your computer or is it in this PR as well?

bqpd:
Looks like @whoburg was right about that causing potential build issues! Funny it didn't affect the tests... Pushing up another change.

galbramc:
The tests don't go through the same kind of install/unistall/install/uninstall/install process that the gpkit_install does. It does that to make sure things are working with the pip cache, but that seems to have found this other bug as well.

bqpd:
It's in the PR as well now.

bqpd:
@whoburg, ready for review

docs/source/releasenotes.rst
```diff
@@ -3,8 +3,29 @@ Release Notes
 
 This page lists the changes made in each point version of gpkit.
 
+Version 0.5.0
+=============
+ * No longer recommend the use of linked variables and subinplace
```
whoburg: should this say "(see below)"?

docs/source/releasenotes.rst
```diff
@@ -3,8 +3,29 @@ Release Notes
...
+ * Switched default solver to MOSEK
+ * Added Linked Variable diagram (PR #915)
+ * Dropped operator overloading with pint (PR #938)
```
whoburg: More like "Changed how overloaded operators interact with pint"? (we didn't drop our operator overloading to form constraints, which is how I might read the above)

docs/source/releasenotes.rst
```diff
@@ -3,8 +3,29 @@ Release Notes
...
+
+We're moving to an explicitly-irregular ``setup`` method, which (if declared for a class inheriting from Model) is automatically called during ``Model.__init__`` inside a ``NamedVariables(self.__class__.__name__)`` environment. This 1) handles the naming of variables more explicitly and efficiently, and 2) allows us to capture variables created within ``setup``, so that constants that are not a part of any constraint can be used directly (several examples of such template models are in the new `Building Complex Models` documentation).
+
+``Model.__init__`` calls ``setup`` with the arguments given to the constructor,  with the exception of ``substitutions``, which is a reserved keyword. This allows for the easy creation of a named model with custom parameter values (as in the documentation's Beam example). ``setup`` methods should return an iterable (list, tuple, ConstraintSet, ...) of constraints or nothing if the model contains no constraints. To declare a submodel cost, set ``self.cost`` during ``setup``. However, we often find declaring a model's cost explicitly just before solving to be a more legible practice.
```
whoburg: 1st sentence: "with the exception of the reserved keyword ``substitutions``"?

gpkit/constraints/relax.py
```diff
@@ -28,8 +28,6 @@ class ConstraintsRelaxedEqually(ConstraintSet):
     """
 
     def __init__(self, constraints):
-        if not isinstance(constraints, ConstraintSet):
-            constraints = ConstraintSet(constraints)
```
whoburg: I think the docstrings throughout this module be updated to say `constraints : ConstraintSet` instead of `constraints : iterable`? (any old iterable won't have `as_posyslt1`, which is called below.

gpkit/nomials/data.py
```diff
@@ -54,13 +50,6 @@ def __hash__(self):
                                          [str(self.units)]))
         return self._hashvalue
 
-    @classmethod
-    def fromnomials(cls, nomials):
-        """Construct a NomialData from an iterable of Signomial objects"""
-        nd = cls()  # use pass-through version of __init__
-        nd.init_from_nomials(nomials)
```
whoburg: should we also drop the `init_from_nomials` method? I think this is the only place it was used

gpkit/nomials/data.py
```diff
@@ -54,13 +50,6 @@ def __hash__(self):
...
-        return nd
```
whoburg: I can't recall why we wrote `fromnomials`. Did we ever need it, or did we just think it might be nice to have someday?

gpkit/nomials/nomial_core.py
```diff
@@ -6,24 +6,6 @@
 from ..repr_conventions import _str, _repr, _repr_latex_
 
 
-def fast_monomial_str(exp, c):
```
whoburg: If we're dropping this method, then we should also drop the commented-out code in `gpkit/nomials/nomial_math.py` that calls it.

gpkit/nomials/nomial_math.py
```diff
@@ -219,26 +215,18 @@ def subinplace(self, substitutions):
...
-                keys.update(exp)
-            raise ValueError("could not substitute for %s" % keys)
-        return mag(cs).sum()
-
     def __le__(self, other):
         if isinstance(other, (Numbers, Signomial)):
             return SignomialInequality(self, "<=", other)
-        return NotImplemented
+        else:
+            return NotImplemented
```
whoburg: just curious, why is this better?

docs/source/index.rst
```diff
@@ -16,6 +16,8 @@ Supported solvers are
 `MOSEK <http://mosek.com>`_
 and `CVXOPT <http://cvxopt.org>`_.
 
+Join our `mailing list <https://mailman.mit.edu/mailman/listinfo/gpkit-users/>`_ and/or `chatroom <https://gitter.im/gpkit-users/Lobby>`_ for support and examples.
+
```
whoburg: great idea!

setup.py
```diff
@@ -53,7 +53,7 @@
     author_email="gpkit@mit.edu",
     url="https://www.github.com/hoburg/gpkit",
     install_requires=["numpy >= 1.8.1", "pint", "scipy"],
-    version="0.5.0.0",
+    version="0.5.0.2",
```
whoburg: I agree with @galbramc that we should use the PyPI testing site to avoid all these minor versions getting pushed to the official site.

whoburg:
another issue: when I run the unit tests from an ipython session (`%run gpkit/tests/run_tests.py`), I get a modified `docs/source/examples/relaxation_output.txt`:

```
diff --git a/docs/source/examples/relaxation_output.txt b/docs/source/examples/r
index 7cace9f..63e2d97 100644
--- a/docs/source/examples/relaxation_output.txt
+++ b/docs/source/examples/relaxation_output.txt
@@ -25,7 +25,7 @@ Free Variables
 --------------
 x : 1.414

-  | Relax.2
+  | Relax.4
 C : 1.414

 Constants
@@ -56,7 +56,7 @@ Free Variables
 --------------
 x : 1

-  | Relax.3
+  | Relax.5
 C : [ 1         2        ]

 Constants
@@ -95,7 +95,7 @@ Free Variables
 x_max : 1
 x_min : 1

-      | Relax.4
+      | Relax.6
 x_max : 1
 x_min : 2
```

galbramc:
Strange that the modified examples isn't showing up in jenkins. Are you running with cvxopt or mosek?

whoburg:
I have both installed but mosek is default, listed first in `gpkit/env/settings`.

whoburg:
the changed output also occurs when I run `python -c "from gpkit.tests import run; run()"`

bqpd:
That's normal and good! Now that we print the modelnums, we can see them accumulate in an interactive environment. Any non-interactive use (such as the tests) will remain deterministic.

whoburg:
but I'm saying that on my computer, running the tests in a non-interactive environment (via `python -c "from gpkit.tests import run; run()"`) results in changing the output files, which seems not normal and not good.

bqpd:
Hmm! That might be because you have a different number of solvers installed? Do you also get changed output with `fulltests.sh` but not with `runtests.sh`?

PR #953, opened 2016-11-22
Vectorization and feasibility

bqpd:
@whoburg, I think we should just merge this, which has the feasibility and the newvectorized stuff. It was easier to fix pylint bugs on the two branches together.

This implements only what we discussed today.

Closes #428
Closes #520
Closes #521
Closes #563
Closes #666
Closes #765
Closes #926
Closes #950 
Closes #946 
Closes #931 
Closes #778

PR #957, opened 2016-11-28
Solarraybugs

### LINKED ISSUE #837: units in solution table printouts

bqpd:
To be rebased after #953 is merged.

Closes #837


whoburg:
holding on review until this is rebased.

Issue #959, opened 2016-11-29
Windows 10 (64bit) installation error, "no solver was given"


norheim:
Perhaps there is an already resolved issue for this, but couldn't find anything either here or in the readthedocs documentation. Did a fresh install of Anaconda (64 bit), created a virtual environment(should not have any effects), and installed all dependencies (see below for pip freeze). Mosek is properly installed as running mosek -f works fine, and the licence is properly working as running msktestlic proves. Gcc is also available in the command line through installation of mingw (64bit). Most gpkit tests pass, except a couple that seem to complain: 

`ValueError: No solver was given; perhaps gpkit was not properly installed, or found no solvers during the installation process`

Going into gpkit\env\settings confirms the lack of solver as installed_solvers is blank. Any idea of what could be wrong? 

pip freeze
backports-abc==0.5
backports.shutil-get-terminal-size==1.0.0
certifi==2016.9.26
colorama==0.3.7
configparser==3.5.0
**ctypesgen==0.post125**
decorator==4.0.10
entrypoints==0.2.2
enum34==1.1.6
functools32==3.2.3.post2
**gpkit==0.4.2.1**
ipykernel==4.5.1
**ipython==5.1.0**
ipython-genutils==0.1.0
ipywidgets==5.2.2
Jinja2==2.8
jsonschema==2.5.1
jupyter-client==4.4.0
jupyter-core==4.2.0
MarkupSafe==0.23
mistune==0.7.3
nbconvert==4.2.0
nbformat==4.1.0
notebook==4.2.3
**numpy==1.11.2**
pathlib2==2.1.0
pickleshare==0.7.4
Pint==0.7.2
prompt-toolkit==1.0.9
Pygments==2.1.3
pyzmq==16.0.2
**scipy==0.18.1**
simplegeneric==0.8.1
singledispatch==3.4.0.3
six==1.10.0
tornado==4.4.2
traitlets==4.3.1
wcwidth==0.1.7
widgetsnbextension==1.2.6
win-unicode-console==0.5

bqpd:
If you open a python console, what is the output of

```python
from gpkit.build import build_gpkit
build_gpkit()
```

norheim:
Thanks for getting back this quickly @bqpd . Here is the output from the console:

```
from gpkit.build import build_gpkit
build_gpkit()
#     Could not find __init__.py
Started building gpkit...
Attempting to find and build solvers:
# Looking for cvxopt
#   Trying to import cvxopt...
[...]

PR #961, opened 2016-11-29
don't use python arch to assume system arch

### LINKED ISSUE #773: Windows mosek_cli erroneously not supported
It gives a `"mosek_cli" not supported for this architecture`, which is a bizarre error message that doesn't really make sense for a command line solver interface.

bqpd:
Closes #773.

@galbramc, I think this will fix that Windows10 bug as well. 

@norheim, I think this will allow for the use of `mosek_cli` in #959, and will probably be on master today.

### DIFF
--- a/gpkit/build.py
+++ b/gpkit/build.py
@@ -164,25 +164,10 @@ class Mosek(SolverBackend):
     def look(self):
         "Looks in default install locations for latest mosek version."
         if sys.platform == "win32":
-            if platform.architecture()[0] == '64bit':
-                rootdir = "C:\\Program Files\\Mosek"
-                mosek_platform = "win64x86"
-                libpattern = "mosek64_?_?.dll"
-                self.flags = "-Wl,--export-all-symbols,-R"
-            ## below is for 32-bit windows ##
-            ## TODO: for unknown reasons neither command line tools or
-            ##       mosek expopt.so works for 32-bit windows
-            ##       someone should look into this if 32-bit windows
-            ##       is really really really needed
-            #elif platform.architecture()[0] == '32bit':
-            #    rootdir = "C:\\Program Files (x86)\\Mosek"
-            #    mosek_platform = "win32x86"
-            #    libpattern = "mosek?_?.dll"
-            #    self.flags = "-Wl,--export-all-symbols,-R"
-            else:
-                log("# Build script does not support mosek"
-                    " your architecture (%s)" % platform.architecture()[0])
-                return
+            rootdir = "C:\\Program Files\\Mosek"
+            mosek_platform = "win64x86"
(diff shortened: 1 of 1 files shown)


whoburg:
basically, if we find a mosek/mosek_cli, we'll try to build/call it, no matter what architecture of Windows the user has.

@galbramc, do you have any comments on this?

galbramc:
test this please

galbramc:
Yeah this is not the answer. I am quite confident that GPkit does not work with 32bit mosek. I don't know why, but I know that's why I put in that code so that any 32bit mosek would be ignored.

bqpd:
Okay, but this solves bugs we've seen in practice of mosek_cli not being used for a particular platform, including for 32bit Python on a 64bit Windows machine. What would you suggest?

galbramc:
So this looks correct now

```
# Looking for mosek_cli
# Build script does not support mosek_cli your architecture (32bit)
# Did not find mosek_cli
```
https://acdl.mit.edu/csi/job/gpkit_PullRequest_Unit/buildnode=windows7x64,optimizer=cvxopt/1731/console

but I still see this message

```
# Looking for mosek_cli
# Build script does not support mosek_cli your architecture (64bit)
# Did not find mosek_cli
```
https://acdl.mit.edu/csi/job/gpkit_PullRequest_Unit/1731/buildnode=windows10x64,optimizer=mosek/console

galbramc:
I don't think that the mosek_cli works for 32bit python on a 64bit system. At least that's what I recall. So you want the build script to only look for mosek_cli for 64bit python on 64bit system.

bqpd:
I'm of the opinion that we should let MOSEK raise that error. If we find
MOSEK, we'll try to use it; if MOSEK doesn't work, that's something they
can figure out.

galbramc:
That is for you and @whoburg to decide how you want the code to behave. You will be responding to users saying "I'm getting all these strange errors when I try to use mosek, why does this not work?". I assure you they will have no way of knowing that the problem is they have a 32bit python.

bqpd:
Yeah, you're probably right. I'm going to merge this for now and let those issues pop up so we can figure out the right way to tackle this.

bqpd:
@norheim, if you install from the current master you should get `mosek_cli` support

galbramc:
I don't think you fixed the problem though. I'm still seeing the message:

```
# Looking for mosek_cli
# Build script does not support mosek_cli your architecture (64bit)
# Did not find mosek_cli
```
https://acdl.mit.edu/csi/job/gpkit_Push/buildnode=windows10x64,optimizer=mosek/356/console

If you look further down below it says:

```
Replaced found solvers (['mosek', 'cvxopt']) with environment var GPKITSOLVERS (mosek, mosek_cli)
```

PR #965, opened 2016-11-30
Cleans up table printing

bqpd:
@whoburg, ready for review.


gpkit/solution_array.py
```diff
@@ -186,9 +186,17 @@ def results_table(data, title, minval=0, printunits=True, fixedcols=True,
...
-            model = ", ".join(k.descr.get("models", ""))
+            kmodels = k.descr.get("models", [])
+            kmodelnums = k.descr.get("modelnums", [])
+            model = ""
+            for i, kmodel in enumerate(kmodels):
+                if model:
+                    model += "/"
+                model += kmodel
+                if kmodelnums[i] != 0:
+                    model += ".%i" % kmodelnums[i]
```
whoburg: I suspect the above for loop reduces to something like

```python
model = "/".join([kstr + (".%i" % knum if knum != 0 else "")
                  for kstr, knum in zip(kmodels, kmodelnums)
                  if kstr])
```

gpkit/solution_array.py
```diff
@@ -215,7 +223,7 @@ def results_table(data, title, minval=0, printunits=True, fixedcols=True,
                 lines.append(["", "", "", ""])
             if model is not "":
                 if not latex:
-                    lines.append([model+" | ", "", "", ""])
+                    lines.append([("modelname",), model, "", ""])
```
whoburg: I think the "modelname" bit deserves an inline comment, including mentioning that it gets specifically looked-for down below in the file.

This is somewhat of a hack, right?

gpkit/solution_array.py
```diff
@@ -253,9 +261,14 @@ def results_table(data, title, minval=0, printunits=True, fixedcols=True,
...
-        lines = [[fmt.format(s) for fmt, s in zip(fmts, line)]
-                 for line in lines]
-        lines = [title] + ["-"*len(title)] + [''.join(l) for l in lines] + [""]
+        for i, line in enumerate(lines):
+            if line[0] == ("modelname",):
+                line = [fmts[0].format(" | "), line[1]]
+            else:
+                line = [fmt.format(s) for fmt, s in zip(fmts, line)]
+            line = "".join(line).rstrip()  # pylint:disable=redefined-variable-type
+            lines[i] = line
```
whoburg: why not `line = "".join(line).rstrip()`? This would kill the pylint disable too

gpkit/solution_array.py
bqpd: Yeah. We should write some elegant table printing code at some point.

gpkit/solution_array.py
bqpd: Yeah. Because it's a bit of a hack I decided to go for legibility here :o

whoburg:
hey @mjburton11, are you using gpkit's latex table printing, or do you have your own local version? With this merged I'd suggest switching to gpkit's if you haven't already -- it will help us to find bugs and other things to improve.

And if you have any local improvements, you might make those into a PR.

bqpd:
His has always been semi-custom; I wrote a new one with him this afternoon that is more custom but also more general, and we're in the process of getting it ready for `master`.

mjburton11:
The solution table is gpkit's.  The one @bqpd is referring to is submodel latex printing.

Issue #905, opened 2016-09-20
Substituted varkey not found in BoundedConstraintSet


pgkirsch:
``` python
In [1]: from gpkit import Variable, Model

In [2]: from gpkit.constraints.bounded import BoundedConstraintSet

In [3]: x = Variable('x')

In [4]: y = Variable('y')

In [5]: m = Model(x, [x >= y], {'y':1})

In [6]: bm = Model(m.cost, BoundedConstraintSet(m))

In [7]: bm.solve()
Using solver 'cvxopt'
Solving for 1 variables.
Solving took 0.022 seconds.
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-7-14295aee770d> in <module>()
----> 1 bm.solve()

C:\Users\philippe.kirschen\Documents\Optimization\gpkit\gpkit\constraints\prog_factories.pyc in solvefn(self, solver, verbosity, skipswee
pfailures, *args, **kwargs)
     89         if "sweepvariables" in solution:
     90             solution["variables"].update(solution["sweepvariables"])
---> 91         self.process_solution(solution)
     92         self.solution = solution  # NOTE: SIDE EFFECTS
     93         return solution

C:\Users\philippe.kirschen\Documents\Optimization\gpkit\gpkit\constraints\set.pyc in process_solution(self, sol)
    285         for constraint in self:
    286             if hasattr(constraint, "process_solution"):
--> 287                 constraint.process_solution(sol)

C:\Users\philippe.kirschen\Documents\Optimization\gpkit\gpkit\constraints\bounded.pyc in process_solution(self, sol)
     86             if abs(lam_lt) >= 1e-7:  # arbitrary threshold
     87                 out["sensitive to lower bound"].append(varkey)
---> 88             value = mag(sol["freevariables"][varkey])
     89             distance_below = np.log(value/self.lowerbound)
     90             distance_above = np.log(self.upperbound/value)

C:\Users\philippe.kirschen\Documents\Optimization\gpkit\gpkit\keydict.pyc in __getitem__(self, key)
     97         if not keys:
     98             del self.keymap[key] # remove blank entry added due to defaultdict
---> 99             raise KeyError("%s was not found." % key)
    100         values = []
    101         for key in keys:

KeyError: 'y was not found.'

```

bqpd:
Weird, it looks like you may have a different implementation of defaultdict? I can make a branch that fixes this, but for now could you try editing the indented block line 98 -100 to

```python
[...]

PR #976, opened 2016-12-03
minor improvements for next release

whoburg:
place to collect minor edits not worthy of an entire PR.


bqpd:
This is not the same bug as #905, though triggered by the same program.

bqpd:
This error is caused by bounding a constant, which creates tautological constraints which are not passed to the solver, and hence not present as `lams`

gpkit/tests/t_constraints.py
```diff
@@ -221,8 +222,21 @@ def test_sigconstr_in_sp(self):
...
+
+class TestBounded(unittest.TestCase):
+    """Test bounded constraint set"""
+
+    def test_substitution_issue905(self):
+        x = Variable("x")
+        y = Variable("y")
+        m = Model(x, [x >= y], {"y": 1})
+        bm = Model(m.cost, Bounded(m))
+        sol = bm.solve()
```
bqpd: `verbosity=0`

bqpd:
@whoburg ready for review. (not sure I see the point of using a `minor` branch that's not focused on particular files)

Issue #753, opened 2016-06-29
clarify infeasibility warnings


mayork:
I haven't seen these warnings before. These first two are thrown mid solve. Not sure if that indicates anything specific, or if the model is just bad. I was thinking it means a variable might be going unbounded to infinity so I ran with the bounded debugging tool. Seemed to stop the warning, but I wanted to double check this was reasonable.

```
Warning (from warnings module):
  File "c:\users\mayork\documents\github\gpkit\gpkit\nomials\substitution.py", line 147
    mag(cs_)[i] *= sub**x
RuntimeWarning: overflow encountered in double_scalars

Warning (from warnings module):
  File "C:\Python27\lib\site-packages\pint\quantity.py", line 715
    magnitude = magnitude_op(new_self._magnitude, other._magnitude)
RuntimeWarning: invalid value encountered in double_scalars
```

this warning is given when the solve terminates. I suspect it's similar to infeasible but I'm not 100% sure. 

```
Traceback (most recent call last):
  File "C:\Users\mayork\Documents\GitHub\gpkit-models\gpkitmodels\aircraft\engine.py", line 257, in <module>
    bounds = engineOffD.determine_unbounded_variables(engineOffD,verbosity=4)
  File "C:\Users\mayork\Documents\GitHub\gpkit-models\gpkitmodels\aircraft\engine.py", line 164, in determine_unbounded_variables
    sol = m.localsolve(solver, verbosity, **kwargs)
  File "c:\users\mayork\documents\github\gpkit\gpkit\constraints\prog_factories.py", line 80, in solvefn
    result = solvefn(solver, verbosity-1, *args, **kwargs)
  File "c:\users\mayork\documents\github\gpkit\gpkit\constraints\signomial_program.py", line 112, in localsolve
    result = nearest_feasible.solve(solver, verbosity=verbosity-1)
  File "c:\users\mayork\documents\github\gpkit\gpkit\geometric_program.py", line 211, in solve
    nu=solver_out["nu"], la=solver_out["la"])
  File "c:\users\mayork\documents\github\gpkit\gpkit\geometric_program.py", line 359, in check_solution
    primal_exp_vals[mi].sum())
RuntimeWarning: Primal solution violates constraint: 1.00004530548 is greater than 1.
```

bqpd:
Yeah, looks like a dual-infeasible model?

mayork:
[...]

PR #977, opened 2016-12-03
Mosek 8

### LINKED ISSUE #975: MOSEK 8 not working on the ubuntu CI server
see #974

bqpd:
@galbramc, super sorry for the hassle with MOSEK 8, but since I can't confirm the bug I've changed the constraint that's failing tests on reynolds-ubuntu16 to be a bit more dual-infeasible. I expect this to pass tests with MOSEK 8 on reynolds-ubuntu16 now?

Closes #975

### DIFF
--- a/gpkit/tests/t_model.py
+++ b/gpkit/tests/t_model.py
@@ -404,7 +404,7 @@ class TestSP(unittest.TestCase):
         from gpkit.constraints.bounded import Bounded
         x = Variable("x")
         y = Variable("y")
-        m = Model(x*y, [x*y**1.000001 >= 100])
+        m = Model(x*y, [x*y**1.01 >= 100])
         with self.assertRaises((RuntimeWarning, ValueError)):
             m.solve(self.solver, verbosity=0)
         m = Model(x*y, Bounded(m, verbosity=0))
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,7 @@ setup(
     author_email="gpkit@mit.edu",
     url="https://www.github.com/hoburg/gpkit",
     install_requires=["numpy >= 1.8.1", "pint", "scipy"],
-    version="0.5.0.3",
+    version="0.5.0.4",
     packages=["gpkit", "gpkit._mosek", "gpkit.tests", "gpkit.interactive",
               "gpkit.nomials", "gpkit.constraints", "gpkit.tools"],
     package_data={"gpkit": ["pint/*", "env/*"],
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,7 @@ setup(
     author_email="gpkit@mit.edu",
     url="https://www.github.com/hoburg/gpkit",
     install_requires=["numpy >= 1.8.1", "pint", "scipy"],
-    version="0.5.0.2",
+    version="0.5.0.3",
     packages=["gpkit", "gpkit._mosek", "gpkit.tests", "gpkit.interactive",
               "gpkit.nomials", "gpkit.constraints", "gpkit.tools"],
     package_data={"gpkit": ["pint/*", "env/*"],


galbramc:
Ok, Mosek 8 is back.

test this please

galbramc:
So I don't know at all what dual-infeasible actually means, but it sounds like something that is bad and should produce an error. From these error messages, I'm guessing Mosek 8 is actually solving something it should not solve? What kind of answer did you get with the old problem that was not quite as dual-infeasible?

I think this is something to be discussed with @whoburg, I can only ask the question without the means to contribute to the answer.

bqpd:
It's a problem were we've intentionally constructed a cost that is has a non-zero but small gradient that, if followed, will lead variables to values of infinity and 1/infinity.

With the problem as originally posed, changing the variable's values by 30 orders of magnitude would change the cost by 0.006% (improvement percentage is linear in the orders of magnitude). It's probably reasonable for a solver to not follow this gradient, if annoyingly clever of it in this case.

I've modified the formulation so that changing the variable's values by 30 orders of magnitude changes the cost by 100%. MOSEK does not ignore this potential improvement, which is also probably the right thing to do.

@whoburg: the question, I think, is what threshold to expect from solvers?

whoburg:
The threshold to expect is clearly documented for the solvers, so this is related to #753.

The current cvxopt thresholds appear to be (https://github.com/cvxopt/cvxopt/blob/master/src/python/cvxprog.py):

```
       options['show_progress'] True/False (default: True)
       options['maxiters'] positive integer (default: 100)
       options['refinement'] nonnegative integer (default: 1)
       options['abstol'] scalar (default: 1e-7)
       options['reltol'] scalar (default: 1e-6)
       options['feastol'] scalar (default: 1e-7).

        If the exit status is 'optimal', then the primal and dual
        infeasibilities are guaranteed to be less than 
        solvers.options['feastol'] (default 1e-7).  The gap is less than
        solvers.options['abstol'] (default 1e-7) or the relative gap is 
        less than solvers.options['reltol'] (defaults 1e-6).   
```

And for mosek, from here http://docs.mosek.com/8.0/capi/all-parameters.html#dparam.intpnt_nl_tol_pfeas. I just checked and none of these changed from mosek 7 to mosek 8, so it's a little surprising that we're seeing different behavior. If we could easily write out an expopt input file, might be worth sending mosek a MWE?

```

MSK_DPAR_INTPNT_NL_TOL_PFEAS

    Primal feasibility tolerance used when a nonlinear model is solved.

    Accepted Values: [0.0 ;1.0 ]

    Default Value: 1.0e-8

    Groups: Interior-point method, Termination criterion, Nonlinear convex method

MSK_DPAR_INTPNT_NL_TOL_REL_GAP

    Relative gap termination tolerance for nonlinear problems.

    Accepted Values: [1.0e-14 ;+inf ]

    Default Value: 1.0e-6

    Groups: Termination criterion, Interior-point method, Nonlinear convex method

MSK_DPAR_INTPNT_NL_TOL_REL_STEP

    Relative step size to the boundary for general nonlinear optimization problems.

    Accepted Values: [1.0e-4 ;0.9999999 ]

    Default Value: 0.995

    Groups: Interior-point method, Nonlinear convex method

MSK_DPAR_INTPNT_QO_TOL_DFEAS

    Dual feasibility tolerance used when the interior-point optimizer is applied to a quadratic optimization problem..

    Accepted Values: [0.0 ;1.0 ]

    Default Value: 1.0e-8

    Groups: Interior-point method, Termination criterion

MSK_DPAR_INTPNT_QO_TOL_INFEAS

    Controls when the conic interior-point optimizer declares the model primal or dual infeasible. A small number means the optimizer gets more conservative about declaring the model infeasible.

    Accepted Values: [0.0 ;1.0 ]

    Default Value: 1.0e-10

    Groups: Interior-point method, Termination criterion

```

galbramc:
So all your tests will fail until this branch gets merged... Not sure if you want to try to sort this out first or should I revert back Mosek to 7 on reynolds-ubuntu16?

bqpd:
@whoburg, please review

Issue #824, opened 2016-08-30
should all Nomial.__eq__ return constraints, not just between monomials?


bqpd:
@whoburg in #818 

> BTW, this is making me wonder whether we should implement `Nomial.__eq__`to return a Constraint (consistent with other operators we've overloaded), and then call for `__nonzero__` when we explicitly want it, e.g. in unit tests. This is clearly a separate design discussion though.

If this is done, as part of it we should implement https://github.com/convexengineering/gpkit/issues/956

whoburg:
Some further detail on how this proposal would work:

`Nomial.__eq__` would try to create a Constraint, which will usually lead to raising an Exception. There are two cases where an Exception would not be raised:

1. If both Nomials are Monomials, in which case a `MonomialEquality` constraint will be returned.
2. Inside a `SignomialsEnabled` environment, in which case a `SignomialEquality` constraint will be returned.

Testing whether two Nomial objects are the same would be accomplished by a method other than `__eq__`, e.g. `Nomial.equivalent_to(other)`.

`Constraint.__nonzero__` behavior would be up for discussion -- we could use the current 0 means not always-true and 1 means always-true, or switch to something more pythonic like 1 for "there's a non empty constraint here", and then implement `Constraint.always_true()` that returns a bool.

PR #978, opened 2016-12-04
Catch bad elements

### LINKED ISSUE #794: numpy bools showing up as constraints in tests
you can see them by uncommenting this line https://github.com/hoburg/gpkit/pull/787/files/0f425f6e3ba7a2bf56b596de7ca9452334bb1947#diff-239d6c98c4984f5dd74cfd8222ed2d3dR140

### LINKED ISSUE #871: numpy.bool_ constraints used in NomialArray's ==
We should catch this later, or make it more explicit with a `Constraint.Set.__nonzero__`

bqpd:
@mayork, this would print the following for https://github.com/hoburg/turbofan/issues/1:

```
ValueError                                Traceback (most recent call last)
/home/tao/projects/turbofan/CFM_56_performance_components_setup.py in <module>()
   1195 
   1196 if __name__ == "__main__":
-> 1197     engine = Engine(0, True, 4)
   1198     mission = TestMission(engine)
   1199 

/home/tao/projects/gpkit/gpkit/constraints/model.pyc in __init__(self, cost, constraints, *args, **kwargs)
     63                 start_args = [cost, constraints]
     64                 args = tuple(a for a in start_args if a is not None) + args
---> 65                 constraints = self.setup(*args, **kwargs)  # pylint: disable=no-member
     66                 from .. import NAMEDVARS, MODELS, MODELNUMS
     67                 setup_vars = NAMEDVARS[tuple(MODELS), tuple(MODELNUMS)]

/home/tao/projects/turbofan/CFM_56_performance_components_setup.py in setup(self, res7, cooling, N)
    131                 #only if mixing = false
    132                 #compute f without mixing, overestimation if there is cooling
--> 133                 TCS([self.combustor['eta_{B}'] * self.engineP['f'] * self.combustor['h_f'] + self.engineP['h_{t_3}'] >= 'h_{t_4}']),
    134 
    135                 self.engineP['P_{t_4}'] == self.combustor['\pi_{b}'] * self.engineP['P_{t_3}'],   #B.145

/home/tao/projects/gpkit/gpkit/constraints/tight.pyc in __init__(self, constraints, substitutions, reltol, raiseerror)
     11     def __init__(self, constraints, substitutions=None, reltol=1e-6,
     12                  raiseerror=False):
---> 13         super(Tight, self).__init__(constraints, substitutions)
     14         self.reltol = reltol
     15         self.raiseerror = raiseerror

/home/tao/projects/gpkit/gpkit/constraints/set.py in __init__(self, constraints, substitutions)
     48                                         % (repr(side), type(side)))
     49                     raise ValueError("%s\nFull constraint: %s"
---> 50                                      % (cause, constraint))
     51                 elif not isinstance(constraint, ConstraintSet):
     52                     if hasattr(constraint, "__iter__"):

ValueError: An ArrayConstraint was created with elements of numpy.bool_, because
NomialArray comparison with 'h_{t_4}' <type 'str'> does not return a valid cons
traint.
Full constraint: [1e-06*h_{t_3}_Engine, EnginePerformance, CompressorPerformance
_(0,) + eta_{B}_Engine, Combustor*f_Engine, EnginePerformance, CombustorPerforma
nce_(0,)*h_f_Engine, Combustor, 1e-06*h_{t_3}_Engine, EnginePerformance, Compres
sorPerformance_(1,) + eta_{B}_Engine, Combustor*f_Engine, EnginePerformance, Com
bustorPerformance_(1,)*h_f_Engine, Combustor, 1e-06*h_{t_3}_Engine, EnginePerfor
mance, CompressorPerformance_(2,) + eta_{B}_Engine, Combustor*f_Engine, EnginePe
rformance, CombustorPerformance_(2,)*h_f_Engine, Combustor, 1e-06*h_{t_3}_Engine
, EnginePerformance, CompressorPerformance_(3,) + eta_{B}_Engine, Combustor*f_En
gine, EnginePerformance, CombustorPerformance_(3,)*h_f_Engine, Combustor] >= h_{
t_4}
```

(I artificially trimmed it to 80 characters wide for the error message to show more of what it would look like in a terminal window)

Closes #871 
Closes #794

### DIFF
--- a/gpkit/constraints/set.py
+++ b/gpkit/constraints/set.py
@@ -1,7 +1,7 @@
 "Implements ConstraintSet"
 import numpy as np
 
-from ..small_classes import HashVector
+from ..small_classes import HashVector, Numbers
 from ..keydict import KeySet, KeyDict
 from ..small_scripts import try_str_without
 from ..repr_conventions import _str, _repr, _repr_latex_
@@ -23,6 +23,7 @@ class ConstraintSet(list):
         # initializations for attributes used elsewhere
         self.posymap = []
         self.unused_variables = None
+        self.numpy_bools = False
 
         # get substitutions and convert all members to ConstraintSets
         subs = dict(substitutions) if substitutions else {}
@@ -33,11 +34,32 @@ class ConstraintSet(list):
         else:
             # constraintsetify everything
             for i, constraint in enumerate(self):
-                if not isinstance(constraint, ConstraintSet):
+                guess = None
+                if getattr(constraint, "numpy_bools", None):
+                    cause = ("An ArrayConstraint was created with elements of"
+                             " numpy.bool_")
+                    for side in [constraint.left, constraint.right]:
+                        if not (isinstance(side, Numbers)
+                                or hasattr(side, "exps")
+                                or hasattr(side, "__iter__")):
+                                    cause += ( ", because "
+                                        "NomialArray comparison with %.10s %s"
+                                        " does not return a valid constraint."
+                                        % (repr(side), type(side)))
+                    raise ValueError("%s\nFull constraint: %s"
+                                     % (cause, constraint))
+                elif not isinstance(constraint, ConstraintSet):
                     if hasattr(constraint, "__iter__"):
                         list.__setitem__(self, i, ConstraintSet(constraint))
-                    elif isinstance(constraint, bool):
-                        # TODO refactor depending on the outcome of issue #824
+                    elif not hasattr(constraint, "varkeys"):
+                        if isinstance(constraint, np.bool_):
+                            # allow NomialArray equalities (arr == "a", etc.)
+                            self.numpy_bools = True  # but mark them
+                            # so we can catch them (see above) in ConstraintSets
+                            continue
+                        cause = "" if not isinstance(constraint, bool) else (
+                                            "Did the constraint list contain"
+                                            " an accidental equality?")
                         if len(self) == 1:
                             loc = "as the only constraint"
                         elif i == 0:
@@ -46,12 +68,8 @@ class ConstraintSet(list):
                             loc = "at the end, after %s" % self[i-1]
                         else:
                             loc = "between %s and %s" % (self[i-1], self[i+1])
-                        raise ValueError("%s found %s."
-                                         " Did the constraint list contain an"
-                                         " accidental equality?"
-                                         % (type(constraint), loc))
-                # conditional because numpy_bools are used for vector ==
-                # TODO: remove conditional or otherwise clean up
+                        raise ValueError("%s was found %s. %s"
+                                         % (type(constraint), loc, cause))
                 if hasattr(self[i], "substitutions"):
                     self.substitutions.update(self[i].substitutions)
         self.reset_varkeys()
--- a/gpkit/constraints/set.py
+++ b/gpkit/constraints/set.py
@@ -34,20 +34,8 @@ class ConstraintSet(list):
         else:
             # constraintsetify everything
             for i, constraint in enumerate(self):
-                guess = None
                 if getattr(constraint, "numpy_bools", None):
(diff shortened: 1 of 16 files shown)


bqpd:
@whoburg, ready for review

whoburg:
@bqpd can you give a quick overview so I understand the context?

What's your proposal for #824? Would we support SignomialEquality constraints through operator overloading?

What's the behavior when a vector is involved in an == constraint?

bqpd:
Hmm nothing to do with SignomialEqualityConstraints, but rather with NomialArray `==` overloading. As with Signomials, we want `==` to be useful both for constraints (when applicable) and for equality. As a result, expressions like those tested `x = VectorVariable(2, "x"); x == "ababa"` result in `[False, False]`. The proposal is to allow such boolsy ArrayConstraints to be created, but warn about them when they're put into another ConstraintSet.

gpkit/constraints/set.py
```diff
@@ -262,3 +256,36 @@ def process_result(self, result):
...
+    else:
+        loc = "between %s and %s" % (cns[i-1], cns[i+1])
+    raise ValueError("%s was found %s. %s"
+                     % (type(constraint), loc, cause))
+
+
+def raise_badsubconstraintarray(constraint):
+    "Identify the bad subconstraint array and raise a ValueError"
+    cause = ("An ArrayConstraint was created with elements of"
+             " numpy.bool_")
```
whoburg: this is nitpicky, but hardcoding numpy.bool_ in the above seems ugly to me -- this could cause unexpected behavior if this method is accidentally called in another context. I'd either

a) retrieve the object type numpy.bool_ somehow instead of hardcoding it, so that it will never be incorrect, or

b) change the method name to include numpy_bool so that the method will never be used inappropriately.

gpkit/constraints/set.py
```diff
@@ -262,3 +256,36 @@ def process_result(self, result):
...
+        " an accidental equality?")
+    if len(cns) == 1:
+        loc = "as the only constraint"
+    elif i == 0:
+        loc = "at the start, before %s" % cns[i+1]
+    elif i == len(cns) - 1:
+        loc = "at the end, after %s" % cns[i-1]
+    else:
+        loc = "between %s and %s" % (cns[i-1], cns[i+1])
+    raise ValueError("%s was found %s. %s"
```
whoburg: SUUUUPER nitpicky but why not put a space before "Did" in the `cause` string and then make this `raise ValueError("%s was found %s.%s"`

gpkit/tests/t_constraints.py
```diff
@@ -14,6 +14,24 @@
...
+            _ = Model(x, [x >= 1, x == "A", x >= 1, ])
+        with self.assertRaises(ValueError):
+            _ = Model(x, [x == "A", x >= 1])
+        v = VectorVariable(2, "v")
+        with self.assertRaises(ValueError):
+            _ = Model(x, [v == "A"])
+        with self.assertRaises(ValueError):
+            _ = Model(x, [v <= ["A", "B"]])
+        with self.assertRaises(ValueError):
+            _ = Model(x, [v >= ["A", "B"]])
```
whoburg: These tests all involve string comparison; I also want to see the much more interesting case of `VectorVariable(2, "v") == Variable("x")`

bqpd: This is for testing the comparisons that contained numpy.bools!

gpkit/constraints/set.py
bqpd: done!

gpkit/constraints/set.py
bqpd: did (b)

gpkit/tests/t_constraints.py
whoburg: got it, I was somehow under the incorrect impression that we don't always support for overloaded `==` involving vectors.

PR #980, opened 2016-12-04
change Tight tol to 1e-2

### LINKED ISSUE #968: Default constraint tightness threshold for Tight
This has come to my attention as the d8 model grows bigger and bigger. Tight is often implemented in models to see whether solutions makes sense (eg. for aircraft design, weight and drag constraints should always be tight since we are minimizing fuel burn or takeoff weight). 

But sometimes, a large number of constraints are shown to be not 'tight', although sometimes they have errors past the default threshold only due to numerical precision. The following example is from the solution of hoburg/d8/CFP_Fuselage_Performance.py, commit 3ab890. 

image

Although all constraints are tight, the numerical errors in the SP solution add up to show that the constraints are past the error threshold. So, instead of having Tight work to e-6 (0.0001%), we should perhaps reduce the default error to something on the order of 0.1-0.01%.

bqpd:
Closes #968

@whoburg, ready for review

### DIFF
--- a/gpkit/constraints/tight.py
+++ b/gpkit/constraints/tight.py
@@ -7,11 +7,13 @@ from .. import SignomialsEnabled
 
 class Tight(ConstraintSet):
     "ConstraintSet whose inequalities must result in an equality."
+    reltol = 1e-6
 
-    def __init__(self, constraints, substitutions=None, reltol=0.01,
+    def __init__(self, constraints, substitutions=None, reltol=None,
                  raiseerror=False):
         super(Tight, self).__init__(constraints, substitutions)
-        self.reltol = reltol
+        if reltol:
+            self.reltol = reltol
         self.raiseerror = raiseerror
 
     def process_result(self, result):
--- a/gpkit/constraints/tight.py
+++ b/gpkit/constraints/tight.py
@@ -8,7 +8,7 @@ from .. import SignomialsEnabled
 class Tight(ConstraintSet):
     "ConstraintSet whose inequalities must result in an equality."
 
-    def __init__(self, constraints, substitutions=None, reltol=1e-6,
+    def __init__(self, constraints, substitutions=None, reltol=0.01,
                  raiseerror=False):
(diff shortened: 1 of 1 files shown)


whoburg:
Guys I'm sorry but this is really not the right way to deal with this. I predict this would come back to bite us later.

I have several ideas for how to handle constraints that aren't tight within our reltol of 1e-6. But none of them involve a looser tolerance. Happy to discuss in detail.

In the meantime, if you really want a lower tolerance, I recommend one of the following solutions. But I don't think we can endorse such a loose tolerance at a GPkit-wide level.

1) You can pass the reltol argument for every `Tight` you create. E.g. define `TTOL=1e-3` at the top of your file, and then create `Tight(constraint, reltol=TTOL)`

2) You can redefine `Tight` locally, e.g.

```python
class NotSoTight(Tight):
    "A Tight with a looser tolerance"
    def __init__(self, constraints, reltol=1e-3, **kwargs):
        super(NotSoTight, self).__init__(constraints, reltol, **kwargs)
```

3) Sounds like Ned doesn't support this, but we could give Tight a class attribute reltol, and change the `__init__` kwarg default to `reltol=None`. If no reltol is specified, it would be pulled from the class attribute, defaulting to `reltol=1e-6`. Then you could simply

```python
from gpkit.constraints import Tight
Tight.reltol = 1e-3
```

This design pattern is used elsewhere e.g. in cvxopt and in matplotlib.

bqpd:
That last design pattern looks good! I think there was a game of telephone going on in #968, where I thought the idea was to specify within a particular submodel what the tightness would be by means of an attribute of that submodel.

bqpd:
I'm happy to keep `1e-6` as the default for now, but given that literally none of our users support that as a default, it seems strange to not raise it to a more reasonable interpretation of user intent.

bqpd:
@whoburg, ready for review

bqpd:
@mayork, @1ozturkbe, this is now merged with syntax (3) above

PR #986, opened 2016-12-04
Geometric Program inherits from ConstraintSet

### LINKED ISSUE #819: have m.program.gps[x].result be able to print a table

bqpd:
Closes #819


bqpd:
@whoburg, ready for review

gpkit/constraints/geometric_program.py
```diff
@@ -44,35 +46,28 @@ class GeometricProgram(NomialData):
...
-        # pylint: disable=too-many-locals,super-init-not-called
-        # note: any substitutions in constraints are ignored -- overwritten
-        # by substitutions kwarg. This will be fixed when GeometricProgram
-        # inherits from ConstraintSet.
-        self.cost = cost
-        self.constraints = constraints
-        self.substitutions = substitutions if substitutions else {}
-
-        ## Create list of substituted posynomials <= 1
+        # pylint:disable=super-init-not-called
```
whoburg: maybe this pylint disable should go down by the `CostedConstraintSet` init so it's more obvious which init is not being called

gpkit/constraints/geometric_program.py
```diff
@@ -44,35 +46,28 @@ class GeometricProgram(NomialData):
...
+        # initialize attributes modified by internal methods
+        self.result = None
+        self.solver_log = None
+        self.solver_out = None
+        CostedConstraintSet.__init__(self, cost, constraints)
+        # NOTE: unnecessary work done above in creating self.substitutions
+        self.substitutions = substitutions
```
whoburg: great inline comment, but I'm slightly confused: it doesn't make sense to pass `substitutions` to `CostedConstraintSet.__init__` and have that manage all the subs?

gpkit/constraints/geometric_program.py
```diff
@@ -186,30 +176,22 @@ def _get_solver(solver):
...
-        if hasattr(self.constraints, "process_result"):
-            self.constraints.process_result(self.result)
-        else:
-            for constraint in self.constraints:
-                if hasattr(constraint, "process_result"):
-                    constraint.process_result(self.result)
+            print("solution checking took %.2g%% of solve time" %
+                  ((time() - tic) / soltime * 100))
 
+        self.process_result(self.result)
```
whoburg: nice.

gpkit/constraints/geometric_program.py
```diff
@@ -271,28 +252,20 @@ def _compile_result(self, solver_out):
...
-            # TODO: enable this once there's a plan for how to use it
-            # result["sensitivities"]["constraints"][str(constr)] = constr_sens
+        var_senss = self.sens_from_dual(la[1:].tolist(),
+                                        [[nu[i] for i in m_idx]
+                                         for m_idx in self.m_idxs[1:]])
+        # add cost's sensitivity in
+        var_senss += {var: sum([self.cost.exps[i][var]*nu[i] for i in locs])
+                      for (var, locs) in self.cost.varlocs.items()
+                      if (var in self.cost.varlocs
+                          and var not in self.posynomials[0].varlocs)}
```
whoburg: I know this isn't new code but part of it is confusing me. This only happens when the cost is not `posynomials[0]`?

bqpd: The cost is always posynomials[0], so this is just adding those sensitiviites in.

gpkit/constraints/geometric_program.py
bqpd: It has to be the first line in the function, it seems.

whoburg: oh weird gotcha

gpkit/constraints/geometric_program.py
bqpd: Because the constraint set substitutions sometimes remove the cost's value, which will be pulled out of the cost again by CostedConstraintSet.

On the other hand, determined that we didn't really need to call ConstraintSet.init here...so skipped that.

gpkit/constraints/geometric_program.py
bqpd: self.cost is not substituted, self.posynomials[0] is

bqpd: We could alternately use `var not in self.substitutions`

whoburg: that seems clearer!

bqpd: but we use `posynomials[0]` above (with a comment explaining why), so this seemed more consistent?

PR #989, opened 2016-12-05
Full backwards compatibility with Vectorize and NamedVariables

bqpd:
Undoes `__init__` speedups to implement exactly what we had before, because some models using `LinkedConstraintSet` rely upon it.

@whoburg this is a priority to merge, please review.

### DIFF
--- a/gpkit/constraints/model.py
+++ b/gpkit/constraints/model.py
@@ -6,7 +6,9 @@ from ..geometric_program import GeometricProgram
 from .signomial_program import SignomialProgram
 from .linked import LinkedConstraintSet
 from ..small_scripts import mag
-from .. import end_variable_naming, begin_variable_naming, NamedVariables
+from ..keydict import KeyDict
+from ..varkey import VarKey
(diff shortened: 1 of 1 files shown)


gpkit/constraints/model.py
```diff
@@ -90,6 +75,15 @@ def __init__(self, cost=None, constraints=None, *args, **kwargs):
             # even if they aren't used in any constraints
             self.unused_variables = setup_vars
             self.reset_varkeys()
+        # for backwards compatibility keep add_modelnames
+        if not hasattr(self, "setup") and self.__class__.__name__ != "Model":
```
whoburg: would be nice to have an inline comment here explaining the plan for when this will be removed

gpkit/constraints/model.py
```diff
@@ -122,6 +116,20 @@ def subconstr_latex(self, excluded=None):
         if self.name:
             return "%s_{%s}" % (self.name, self.num)
 
+    def _add_modelname_tovars(self, name, num):
```
whoburg: ditto, inline comment - plan for when this will be removed

whoburg:
@mayork this is on master now

PR #991, opened 2016-12-06
Fixes direct GP creation + lint + speed

bqpd:


bqpd:
@whoburg, will merging directly because this patches master (which was broken in the previous commit for SPs whose initial guess was infeasible...can't believe we don't test any of those!)

gpkit/nomials/nomial_math.py
```diff
@@ -520,10 +516,12 @@ def as_posyslt1(self, substitutions=None):
...
             self.pmap = pmap  # pylint: disable=attribute-defined-outside-init
-            p = Posynomial(exps, cs, simplify=False)
-            if p.any_nonpositive_cs:
+            try:
+                p = Posynomial(exps, cs, simplify=False)
+            except ValueError:
                 raise RuntimeWarning("PosynomialInequality %s became Signomial"
-                                     " after substitution" % self)
+                                     " after substitution %s"
+                                     % (self, substitutions))
```
whoburg: nice.

PR #998, opened 2016-12-07
add initially-infeasible sp

### LINKED ISSUE #992: test initially-infeasible SP
per #991

bqpd:
Closes #992

@whoburg, ready for review

### DIFF
--- a/gpkit/tests/t_model.py
+++ b/gpkit/tests/t_model.py
@@ -272,6 +272,10 @@ class TestSP(unittest.TestCase):
         sol = m.localsolve(verbosity=0)
         self.assertAlmostEqual(sol["cost"], 2**0.5, self.ndig)
         self.assertAlmostEqual(sol(y), 0.5, self.ndig)
+        second_solve_key_names = [key.name[:5]
+                                  for key in m.program.gps[1].cost.exp.keys()
+                                  if key.name[:5] == "\\fbox"]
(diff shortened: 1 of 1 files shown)


whoburg:
Here's a quick illustration of what's going on with the SP in the test:
[infeasible_sp.pdf](https://github.com/hoburg/gpkit/files/636517/infeasible_sp.pdf)

The problem's feasible set is the shaded green area. Am I understanding correctly that you call this "initially infeasible" because our default initial guess of all ones is infeasible?

gpkit/tests/t_model.py
```diff
@@ -259,6 +259,20 @@ def test_values_vs_subs(self):
...
+
+        with SignomialsEnabled():
+            sigc = x >= y + y**2 - y**3
+            sigc2 = x <= y**0.5
+
+        m = Model(1/x, [sigc, sigc2, y <= 0.5])
+
+        sol = m.localsolve(verbosity=0)
+        self.assertAlmostEqual(sol["cost"], 2**0.5, self.ndig)
+        self.assertAlmostEqual(sol(y), 0.5, self.ndig)
```
whoburg: maybe test that one of the objectives is an fbox?

bqpd:
It's not an initial guess for parameter values, though, so it was harder to make than I expected! It's really only an initial guess for what monomial approximation to start with, which I think partly explains the robustness of the heuristic.

PR #1003, opened 2016-12-09
add spdata to tools

bqpd:
as used to create examples in https://github.com/hoburg/gpkit-models/issues/8

### DIFF
--- a/gpkit/tools/spdata.py
+++ b/gpkit/tools/spdata.py
@@ -45,7 +45,11 @@ class SPData(NomialData):
 
     def save(self, filename):
         "Save spdata to an h5 file."
-        import h5py
+        try:
+            import h5py  # pylint:disable=import-error
+        except ImportError as ie:
+            print "SPData.save requires the h5py library."
+            raise ie
         h5f = h5py.File(filename, 'w')
         try:
             h5f.create_dataset('cs', data=self.cs)
--- a/gpkit/tools/spdata.py
+++ b/gpkit/tools/spdata.py
@@ -17,7 +17,7 @@ class SPData(NomialData):
     def __init__(self, model):
         # pylint:disable=super-init-not-called
         if not hasattr(model, "solution"):
-            print "You nede to solve the model first."
+            print "You need to solve the model first."
 
         self.signomials = [model.cost]
         for constraint in model.flat(constraintsets=False):
--- a/gpkit/tools/spdata.py
+++ b/gpkit/tools/spdata.py
@@ -1,7 +1,8 @@
+"Implements SPData class"
 import h5py
 import numpy as np
-from ..nomials import *
-from ..geometric_program import genA
+from ..nomials import NomialData, SignomialInequality, PosynomialInequality
+from ..constraints.geometric_program import genA
 from ..small_scripts import mag
 
 
@@ -15,6 +16,7 @@ class SPData(NomialData):
     ``spdata.save('example_sp.h5')``
     """
     def __init__(self, model):
+        # pylint:disable=super-init-not-called
         if not hasattr(model, "solution"):
             print "You nede to solve the model first."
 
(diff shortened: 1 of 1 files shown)


bqpd:
@whoburg, ready for review

bqpd:
retest this please

galbramc:
This adds h5py as another dependency. Is that intentional? I need to update the scripts to install it if so.

bqpd:
I don't think we should add it as a dependency, no. This was a useful tool
that may be used by future tests, but until that point we're okay.

galbramc:
Hmm, well I need to add it to the jenkins script, or you need to suppress the pylint warning that h5py can't be imported. @whoburg, which solution would you prefer?

gpkit/tools/spdata.py
```diff
@@ -0,0 +1,61 @@
...
+    Usage
+    -----
+
+    ``spdata = SPData(m)``
+    ``spdata.save('example_sp.h5')``
+    """
+    def __init__(self, model):
+        # pylint:disable=super-init-not-called
+        if not hasattr(model, "solution"):
+            print "You need to solve the model first."
```
whoburg: raise a `RuntimeWarning` here?

gpkit/tools/spdata.py
```diff
@@ -0,0 +1,61 @@
...
+
+        self.signomials = [model.cost]
+        for constraint in model.flat(constraintsets=False):
+            if isinstance(constraint, (SignomialInequality,
+                                       PosynomialInequality)):
+                self.signomials.extend(constraint.unsubbed)
```
whoburg: else... print a warning / raise an Exception?

gpkit/tools/spdata.py
```diff
@@ -0,0 +1,61 @@
...
+
+        # k [j]: number of monomials (columns of F) present in each constraint
+        k = [len(p.cs) for p in self.signomials]
+        # p_idxs [i]: posynomial index of each monomial
+        p_idxs = []
+        for i, p_len in enumerate(k):
+            p_idxs += [i]*p_len
+        self.p_idxs = np.array(p_idxs)
+        # A [i, v]: sparse matrix of variable's powers in each monomial
+        self.A, _ = genA(self.exps, self.varlocs)
```
whoburg: inline TODO suggesting that we get these from self.A, etc once NomialData has those properties?

gpkit/tools/spdata.py
```diff
@@ -0,0 +1,61 @@
...
+    def __repr__(self):
+        return str(self.signomials)
+
+    def save(self, filename):
+        "Save spdata to an h5 file."
+        try:
+            import h5py  # pylint:disable=import-error
+        except ImportError as ie:
+            print "SPData.save requires the h5py library."
+            raise ie
```
whoburg: could assert that if `filename` has an extension, it isn't something other than `.h5`

gpkit/tools/spdata.py
```diff
@@ -0,0 +1,61 @@
...
+        self.A, _ = genA(self.exps, self.varlocs)
+
+        self.varsols = np.array([mag(model.solution(var))
+                                 for var in self.varlocs])
+        self.varnames = np.array([str(var) for var in self.varlocs])
+
+    def __repr__(self):
+        return str(self.signomials)
+
+    def save(self, filename):
```
whoburg: maybe this should be the `save` method of `NomialData`, to better set ourselves up for doing the same thing with text, csv, etc?

gpkit/tools/spdata.py
bqpd: Good point. Probably a `ValueError`

gpkit/tools/spdata.py
bqpd: Yup, should probably print a warning.

gpkit/tools/spdata.py
bqpd: I think filename is a reasonable thing to expect the user to handle...

PR #1013, opened 2016-12-20
Fix remote skipsolvers

bqpd:
@whoburg, ready for review. No longer high priority.

### DIFF
--- a/gpkit/tests/from_paths.py
+++ b/gpkit/tests/from_paths.py
@@ -9,7 +9,7 @@ from gpkit.tests.helpers import generate_example_tests, new_test
 
 
 class TestFiles(unittest.TestCase):
-    "Stub to be filled with files in pwd/TEST"
+    "Stub to be filled with files in $pwd/TESTS"
     pass
 
 
@@ -64,5 +64,8 @@ def run(filename="TESTS", xmloutput=False, skipsolvers=None):
                if not skipsolvers or s not in skipsolvers]
     tests = generate_example_tests("", [TestFiles], solvers,
                                    newtest_fn=newtest_fn)
+    if not solvers:
+        # Dummy test in case all installed solvers are skipped.
+        tests[0].test_dummy = lambda self: None
     from gpkit.tests.run_tests import run as run_
     run_(tests=tests, unitless=False, xmloutput=xmloutput)
--- a/gpkit/tests/from_paths.py
+++ b/gpkit/tests/from_paths.py
@@ -38,7 +38,7 @@ def add_filetest(testclass, path):
             dirname = os.path.dirname(path)
             if dirname:
                 os.chdir(os.path.dirname(path))
-            mod = importlib.import_module(os.path.basename(path)[:-3])
+            mod = __import__(os.path.basename(path)[:-3])
         finally:
             os.chdir(top_level)
         if not hasattr(mod, "test"):
--- a/gpkit/tests/from_paths.py
+++ b/gpkit/tests/from_paths.py
@@ -3,7 +3,6 @@
 import unittest
 import os
 import re
-import importlib
 from gpkit import settings
 from gpkit.tests.helpers import generate_example_tests, new_test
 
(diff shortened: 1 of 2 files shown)


bqpd:
test models please

galbramc:
test models please

galbramc:
Why is this trying to use mosek?

https://acdl.mit.edu/csi/job/gpkit_PullRequest_Models/buildnode=macys_Sierra,optimizer=cvxopt/23/console

This build should only have cvxopt as the installed solver.

bqpd:
Because mosek is explicitly declared as the solver, and gpkit doesn't check
whether it was found at install time before trying to use mosek...

galbramc:
So people need to add explicit statements to the test in regards to which solver they use? Is this a gpkit problem or a ResearchModel TESTS problem.

bqpd:
About to merge branches on the research repositories that fix this.

galbramc:
All, thats what those PRs were for, that sounds good.

bqpd:
test models please

galbramc:
Oh I forgot. You need to add --no-deps as well only to the local install command. So only with -e do you want --no-cache-dir --no-deps. Otherwise pip downloads all are recompiles all the dependencies as well. Which is why it is taking extra long for this model test.

bqpd:
whoops! wildo.

galbramc:
You should change it so --no-cache-dir also only shows up for the local install.

bqpd:
test models please

bqpd:
You can't use both --no-cache-dir and --no-deps

galbramc:
Oh yes you can. We are using both of those. If you don't use --no-deps with --no-cache, it will download and recompile all dependencies, because it can't find them in the cache. So you need to use them together.

bqpd:
Yeah, just realized -e had to be last...

galbramc:
Ahh, that I can believe. I'm guessing path is an input to -e

bqpd:
Seems like.

bqpd:
test models please

galbramc:
test gpkit please

galbramc:
One more thing. None of the repository tests are registering any coverage information. This seems to be a solution to the problem, 

http://stackoverflow.com/questions/33707441/code-coverage-does-not-cover-dynamically-imported-packages-modules/33707948

which suggest just a slight modification to use the __import__() function instead. Then you will have coverage information for all the ResearchModel tests as well.

bqpd:
I'm already using importlib and the import function, hmm.

bqpd:
test models please

bqpd:
test models please

bqpd:
test models please

galbramc:
Unfortunately we will not know if this fixed the coverage until this is merged. The two pending test did pass, not sure why jenkins didn't update the statues. I had some issues with the github plugin before, so I just downgraded it back to what seamed to be a stable version. I would say this is ready to be merged in my humble opinion.

galbramc:
test models please

galbramc:
test gpkit please

bqpd:
Okay, let's merge and see if this fixes coverage on the research repo tests!

bqpd:
@galbramc, this doesn't seem to have triggered d8_Push etc?

bqpd:
manually triggered repo tests, doesn't seem to have fixed coverage.

galbramc:
Yeah I know. There is something with authorization that even though as best as I can tell I have configured correctly, is preventing downstream builds from getting triggered.

Yeah I saw it didn't fix the coverage either... You might need to dig some more on that if you want coverage on the Research Models, which is really good to have.

bqpd:
Will do, that would be good to have.

galbramc:
For reference, this is how the coverage is generated:

COVERAGE=`which coverage`
echo "from gpkit.tests.test_repo import test_repo; test_repo(xmloutput=True)" > jenkins_test.py
python $COVERAGE run --append --source=. jenkins_test.py
python $COVERAGE html -d htmlcov --omit=gpkit/*,gpkit-models/*,venv2_gpkit/*,jenkins_test.py

So this should the directory htmlcov with coverage information for you to look at.

PR #1019, opened 2016-12-28
handle numbers, esp. zeros, in summing NomialArrays

bqpd:


bqpd:
bad test_call_time draw

PR #1020, opened 2016-12-29
catch sig_ineqs that will turn into posy_ineqs

bqpd:
Closes https://github.com/hoburg/gpkit/issues/1009

@whoburg, ready for review, upper-middle priority.


bqpd:
test models please

bqpd:
models failure on jho / Window7 cvxopt is expected until its cvxopt is updated to 1.1.9

bqpd:
test models please

bqpd:
test models please

bqpd:
test models please

bqpd:
tests are currently broken on gassolar, emailed @mjburton11

gpkit/nomials/nomial_math.py
```diff
@@ -699,8 +705,9 @@ def as_posyslt1(self, substitutions=None):
                                   "`.localsolve` instead of `.solve` to"
                                   " form your Model as a SignomialProgram")
 
-    def as_gpconstr(self, x0):
+    def as_gpconstr(self, x0, substitutions=None):
         "Returns GP approximation of an SP constraint at x0"
+        # todo deal with substitutions
```
whoburg: hmm I think we should raise here to be safe. Otherwise we could have some unexpected behavior.

bqpd: This is about determining when substitutions cause a SigIneq to become a MonoIneq, so I don't think we'd have unexpected behaviour, just sub-optimal!

whoburg: Unless I'm missing something, we currently blatantly ignore any substitutions that our users pass to `SingleSignomialEquality.as_gpconstr`. Seems unexpected to me.

bqpd: Substitutions get applied at the `as_posyslt1` level; `as_gpconstr` just uses them to check if the constraint will be changing type.

PR #1024, opened 2016-12-30
lazy keymapping

### LINKED ISSUE #993: GP and SP solutions don't do keymapping
Consider the following simple Model, which is an SP.

```python
x = Variable("x")
y = Variable("y")
with SignomialsEnabled():
    sig = (y + 6*x >= 13 + x**2)
m = Model(y, [sig])
```

The following things produce valid output:
```python
m.localsolve()("x")
m.localsolve()(x)
m.sp().localsolve()
m.sp().localsolve().table()
m.sp().localsolve()(x)
```
... but `m.sp().localsolve()("x")` results in `ValueError: no variable 'x' found in the solution`.

bqpd:
@whoburg, ready for review

Closes #993

### DIFF
--- a/gpkit/constraints/geometric_program.py
+++ b/gpkit/constraints/geometric_program.py
@@ -4,7 +4,7 @@ from time import time
 import numpy as np
 from ..nomials import NomialData
 from ..small_classes import CootMatrix, SolverLog
-from ..keydict import FastKeyDict
+from ..keydict import KeyDict
 from ..small_scripts import mag
 from ..solution_array import SolutionArray
 from .costed import CostedConstraintSet
@@ -239,8 +239,7 @@ class GeometricProgram(CostedConstraintSet, NomialData):
         nu, la = solver_out["nu"], solver_out["la"]
         # confirm lengths before calling zip
         assert len(self.varlocs) == len(primal)
-        result = {"freevariables": FastKeyDict(zip(self.varlocs,
-                                                   np.exp(primal)))}
+        result = {"freevariables": KeyDict(zip(self.varlocs, np.exp(primal)))}
 
         ## Get cost
         if "objective" in solver_out:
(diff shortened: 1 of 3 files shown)


bqpd:
test models please

bqpd:
test models please

gpkit/constraints/geometric_program.py
```diff
@@ -4,7 +4,7 @@
 import numpy as np
 from ..nomials import NomialData
 from ..small_classes import CootMatrix, SolverLog
-from ..keydict import FastKeyDict
```
whoburg: Looks like `FastKeyDict` is now unused. Should we remove it?

gpkit/keydict.py
```diff
@@ -121,9 +123,8 @@ def __setitem__(self, key, value):
         key, idx = self.parse_and_index(key)
         if key not in self.keymap:
             self.keymap[key].add(key)
-            if hasattr(key, "keys") and self.keymapping:
-                for mapkey in key.keys:
-                    self.keymap[mapkey].add(key)
+            if hasattr(key, "keys"):
+                self._unmapped_keys.add(key)
```
whoburg: confirming: `__setitem__` is the only method that can add a key? There's no `add`, `update`, etc?

whoburg: interesting, I expected the laziness to occur on `__getitem__`, such that I would only map keys for entries I get (e.g. if I call `__getitem__("W")`, then I add `"W"` to `_unmapped_keys`. Doesn't this cause the entire map to get built during SolutionArray construction (which uses `__setitem__`) when that isn't necessary?

bqpd: 1) `update` is overloaded to use `__setitem_`, dictionaries have no `add`
2) the laziness occurs on `parse_and_index`, not `__setitem__`: this line just adds to a queue of "hey these should maybe be mapped at some point"

gpkit/constraints/geometric_program.py
bqpd: I was thinking we'd keep it for now as an example of a definitely unmapped dict, but there's no real reason to, and it's fun to remove code, so I'll do that. :D

PR #1027, opened 2017-01-04
implement ECOS / exponential cone solvers

bqpd:


whoburg:
retest this please

joachimdahl:
Since you already have the setup to input models on conic form, it would not require much work to make an interface for the new exponential cone solver in MOSEK 9, which will replace the general convex solver (including mskexpopt). The new solver seems quite accurate and reliable; more reliable than mskexpopt (much fewer failures in our tests), and just as accurate on problems both solvers can solve.

bqpd:
@joachimdahl that's excellent news! I just downloaded the 9 alpha to give it a spin.

PR #1040, opened 2017-01-18
let's try not installing dependencies for pandas

bqpd:


galbramc:
strange, the build didn't trigger

test gpkit please

galbramc:
test models please

galbramc:
well it looks like conda create does not actually create a separate environment. pip install -e for both cvxopt and mosek appear to be installed int the base python, and hence creating a conflict. I guess I need to hunt more on the virtualenv numpy problem.

bqpd:
:((( dangit

galbramc:
we still probably want this fix to remove the uninstall of numpy

bqpd:
Might it be better to instead update numpy on that computer to the newest version?

galbramc:
No, using different versions of numpy on the different systems is part of the effort to test as many different versions of numpy as possible. Are you using --upgrade when you install pandas? Is so, that's the flag that needs to be removed to prevent pip from trying to upgrade the dependencies.

galbramc:
test gpkit please

galbramc:
test gpkit please

galbramc:
test models please

bqpd:
test models please

bqpd:
@whoburg, merging to fix gpkit model tests on master

PR #1064, opened 2017-02-28
simplify and fix sensitivites, add test

### LINKED ISSUE #1060: incorrect sensitivity mapping on trivial example
Urgent bug.

Here's a simple MWE:

```python
import numpy as np
import pandas as pd
from gpkit import Model, Variable, units


class Aircraft(Model):
    def setup(self):
        W_payload = Variable("W_{payload}", 175*(195 + 30), "lbf", "payload weight")
        f_oew = Variable("f_{oew}", 0.53, "-", "OEW/MTOW")
        fuel_per_nm = Variable("\\theta_{fuel}", 13.75, "lbf/nautical_mile", "fuel burn per nm")
        R = Variable("R", 3000, "nautical_miles", "range")
        mtow = Variable("MTOW", "lbf", "max take off weight")
        self.cost = 61.3e6*units.USD*(mtow/(1e5*units.lbf))**0.807
        return (mtow >= W_payload + f_oew*mtow + fuel_per_nm*R),
    
m = Aircraft()
sol = m.solve()
print sol.table()
print "sens to Wpay: %s" % sol["sensitivities"]["constants"]["W_{payload}"]
```

The result is (note sensitivity to W_payload should be non-zero):

```python
Using solver 'mosek'
Solving for 1 variables.
Solving took 0.0116 seconds.

Cost
----
 9.475e+07 [USD] 

Free Variables
--------------
MTOW : 1.715e+05  [lbf] max take off weight

Constants
---------
            R : 3000       [nmi]     range
  W_{payload} : 3.938e+04  [lbf]     payload weight
theta_{fuel} : 13.75      [lbf/nmi] fuel burn per nm
      f_{oew} : 0.53                 OEW/MTOW

Sensitivities
-------------
            R : +0.41 range
theta_{fuel} : +0.41 fuel burn per nm
      f_{oew} : +0.39 OEW/MTOW
  W_{payload} : +0    payload weight

sens to Wpay: 0.0
```

bqpd:
Closes #1060

@whoburg, ready for review.

### DIFF
--- a/gpkit/nomials/data.py
+++ b/gpkit/nomials/data.py
@@ -165,10 +165,11 @@ def simplify_exps_and_cs(exps, cs, return_map=False):
         if return_map:
             expmap[exp][i] = cs[i]
 
-    if len(matches) > 1:
-        zeroed_terms = (exp for exp, c in matches.items() if mag(c) == 0)
-        for exp in zeroed_terms:
-            del matches[exp]
+    zeroed_terms = (exp for exp, c in matches.items() if mag(c) == 0)
+    for exp in zeroed_terms:
+        del matches[exp]
+    if not matches:  # it's a zero monomial
+        matches[HashVector()] = 0
 
     exps_ = tuple(matches.keys())
     cs_ = list(matches.values())
--- a/gpkit/nomials/nomial_math.py
+++ b/gpkit/nomials/nomial_math.py
@@ -467,6 +467,7 @@ class PosynomialInequality(ScalarSingleEquationConstraint):
                 exps_.pop(i)
                 exps = tuple(exps_)
                 if pmap is not None:
+                    # move constant term's mmap to another attribute
                     self.const_mmap = pmap.pop(i)  # pylint: disable=attribute-defined-outside-init
                     self.const_coeff = coeff  # pylint: disable=attribute-defined-outside-init
                 break
@@ -501,11 +502,14 @@ class PosynomialInequality(ScalarSingleEquationConstraint):
 
         out = []
         for posy in posys:
+            # 1) substitution
             _, exps, cs, subs = substitution(posy, substitutions)
             self._last_used_substitutions = subs
+            # 2) algebraic simplification
             exps, cs, pmap = simplify_exps_and_cs(exps, cs, return_map=True)
+            # 3) constraint simpl. (subtracting the constant term from the RHS)
             exps, cs, pmap = self._simplify_posy_ineq(exps, cs, pmap)
-            if not exps and not cs:  # tautological constraint
+            if not exps and not cs:  # skip tautological constraints
                 continue
 
             #  The monomial sensitivities from the GP/SP are in terms of this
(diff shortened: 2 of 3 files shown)


bqpd:
test models please

gpkit/nomials/data.py
```diff
@@ -165,10 +165,11 @@ def simplify_exps_and_cs(exps, cs, return_map=False):
...
 
-    if len(matches) > 1:
-        zeroed_terms = (exp for exp, c in matches.items() if mag(c) == 0)
-        for exp in zeroed_terms:
-            del matches[exp]
+    zeroed_terms = (exp for exp, c in matches.items() if mag(c) == 0)
+    for exp in zeroed_terms:
+        del matches[exp]
+    if not matches:  # it's a zero monomial
+        matches[HashVector()] = 0
```
whoburg: interesting, why do we need to set `matches[HashVector()] = 0` here? I would have thought we just let `exps_ = (),` and `cs = []` (and therefore `mmap = []` if returned) in that case.

gpkit/nomials/nomial_math.py
```diff
@@ -454,34 +455,27 @@ def __init__(self, left, oper, right):
...
-        return tuple(exps_), cs/coeff
+            if not exp:
+                cs = cs.tolist()
+                coeff -= cs.pop(i)
+                cs = np.array(cs)/coeff
+                exps_ = list(exps)
+                exps_.pop(i)
+                exps = tuple(exps_)
+                if pmap is not None:
+                    self.const_mmap = pmap.pop(i)  # pylint: disable=attribute-defined-outside-init
```
whoburg: whoa, this mutates pmap? That ought to be documented.

whoburg: and is that really intended?

gpkit/nomials/nomial_math.py
```diff
@@ -509,10 +503,10 @@ def as_posyslt1(self, substitutions=None):
         for posy in posys:
             _, exps, cs, subs = substitution(posy, substitutions)
             self._last_used_substitutions = subs
-            exps, cs = self._simplify_posy_ineq(exps, cs)
+            exps, cs, pmap = simplify_exps_and_cs(exps, cs, return_map=True)
+            exps, cs, pmap = self._simplify_posy_ineq(exps, cs, pmap)
```
whoburg: There's a lot going on in the four lines above (substitution, and two simplification steps). Can you add an inline comment to clarify what the two different simplification steps are? Or perhaps make the method names more descriptive?

It's particularly interesting to me that the two simplification steps switched order in this PR -- what's up with that?

gpkit/nomials/data.py
bqpd: Not in the current scheme; the current abstraction for simplification is Bad, but this wasn't the PR to fix it...

gpkit/nomials/nomial_math.py
bqpd: Oh yes.

bqpd:
Added comments per review.

PR #1076, opened 2017-04-05
add soltime to solution

bqpd:
@whoburg, ready to review


bqpd:
test models please

bqpd:
I like that, but can't think of a usecase besides just documentation?

Issue #1079, opened 2017-04-14
getting the top-level variable from a model


whoburg:
Below is a simple model with multiple variables called `W`. In the second to last line, I'd like to do `m["W"]` (or similar), and have that correspond to the top-level W, instead of needing to set `self.W = W`. What's the best practice for doing that? I thought `Model.__getitem__` was designed to get the top-level variable, as documented in the 0.4.2 release notes.

```python
from gpkit import Model, Variable


class Aircraft(Model):
    "The vehicle model"
    def setup(self):
        W = Variable("W", "lbf", "weight")
        self.W = W
        self.propulsor = Propulsor()
        self.airframe = Airframe()
        self.components = [self.propulsor,
                           self.airframe]
        return [W >= sum(c["W"] for c in self.components)]


class Propulsor(Model):
    "the propulsion"
    def setup(self):
        W = Variable("W", 1000, "lbf", "weight")


class Airframe(Model):
    "the aircraft structure"
    def setup(self):
        W = Variable("W", 3000, "lbf", "weight")


if __name__ == "__main__":
    m = Aircraft()
m.cost = m.W
```

bqpd:
(close by #1084)

PR #1084, opened 2017-04-25
add topvar method to get the variable with the same naming

bqpd:
@whoburg, ready for review

(tested in the performance_modeling example)

### DIFF
--- a/gpkit/constraints/set.py
+++ b/gpkit/constraints/set.py
@@ -52,28 +52,38 @@ class ConstraintSet(list):
         if isinstance(key, int):
             return list.__getitem__(self, key)
         else:
-            variables = self.variables_byname(key)
-            if not variables:
-                raise KeyError(key)
-            if variables[0].key.veckey:
-                # maybe it's all one vector variable!
-                from ..nomials import NomialArray
-                vk = variables[0].key.veckey
-                arr = NomialArray(np.full(vk.shape, np.nan, dtype="object"))
-                arr.key = vk
-                for variable in variables:
-                    if variable.key.veckey == vk:
-                        arr[variable.key.idx] = variable
-                    else:
-                        arr = None
-                        break
-                if arr is not None:
-                    return arr
-            elif len(variables) == 1:
-                return variables[0]
-            raise ValueError("multiple variables are called '%s'; use"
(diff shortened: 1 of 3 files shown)


bqpd:
test models please

bqpd:
closes #1079

bqpd:
retest this please

bqpd:
passes dependency tests, unit tests are blocked on jenkins right now but it was passing them before so I'm going to merge!

gpkit/constraints/set.py
```diff
@@ -52,28 +52,40 @@ def __getitem__(self, key):
...
-                if arr is not None:
-                    return arr
-            elif len(variables) == 1:
-                return variables[0]
-            raise ValueError("multiple variables are called '%s'; use"
-                             " variables_byname('%s') to see all of them"
-                             % (key, key))
+            return self._choosevar(key, self.variables_byname(key))
+
+    def _choosevar(self, key, variables):
```
whoburg: docstring please?

whoburg:
see also #1094

gpkit/constraints/set.py
whoburg: you could push the docstring to the `doctopvar` branch (#1094)

gpkit/constraints/set.py
```diff
@@ -52,28 +52,40 @@ def __getitem__(self, key):
...
+        elif len(variables) == 1:
+            return variables[0]
+        raise ValueError("multiple variables are called '%s'; use"
+                         " variables_byname('%s') to see all of them"
+                         % (key, key))
+
+    def topvar(self, key):
+        "If a variable by a given name exists in the top model, return it"
+        # TODO: could this be done with unique_varkeys?
+        if not self.naming:  # pylint: disable=no-member
```
whoburg: where do `ConstraintSet`s get their `naming` attribute? From `Model` I assume? Shouldn't every `ConstraintSet` have `naming` if we want `topvar` to work?

bqpd: We could give every constraintset a naming, yeah! Right now Models law claim to all variables created in them that aren't in a named submodel, whether those are created in lower-level constraintsets or not, and we could extend that by giving that naming to those constraintsets.

PR #1090, opened 2017-05-02
fix #1078

### LINKED ISSUE #1078: Returning soltime from autosweep
@bqpd, I'm sure you have a good reason for doing this, but can you explain why, when I use autosolve, `bst.nsols = 9` but `bst.sols` only has 2 solution dicts.  Mostly I want to return how long it took to solve the 9 passes.  

```python
In [7]: bst = autosweep_1d(M, tol, M["h_{batt}"], [hmin, hmax], solver="mosek")

Solved after  9 passes, cost logtol +/-0.00735
Autosweeping took 5.15 seconds.

In [8]: len(bst.sols)
Out[8]: 2

In [10]: bst.nsols
Out[10]: 9
```

bqpd:
closes #1078 

@whoburg, ready for review


bqpd:
@mjburton11, this is merged now, you can get the solution array from an autosweep

gpkit/tools/autosweep.py
```diff
@@ -1,9 +1,21 @@
...
+from ..small_scripts import mag
+from ..solution_array import SolutionArray
+
+
+def flatten(listy):
+    """ Recursively flatten a list.
+
+    from stackoverflow.com/questions/12472338/flattening-a-list-recursively
+    """
+    if listy == []:
```
whoburg: wouldn't `if not listy` be more pythonic? What if listy is a tuple?

gpkit/tools/autosweep.py
```diff
@@ -126,6 +138,27 @@ def sample_at(self, values):
         "Creates a SolutionOracle at a given range of values"
         return SolutionOracle(self, values)
 
+    @property
+    def sollist(self):
+        "Returns a list of all the solutions in an autosweep"
+        if self.splits:
+            return [self.sols[0], self.splits[0].sollist[1:-1],
+                    self.splits[1].sollist[:-1], self.sols[1]]
```
whoburg: shouldn't this get flattened?

whoburg: i.e. implement using `extend`?

bqpd: I flatten it in solarray, but thought maybe it'd be good to preserve structure here?

bqpd: But, yeah, if I do that I can remove the flatten function. I'm into it.

gpkit/tools/autosweep.py
bqpd: Fixed on the `refactor` branch!

gpkit/tools/autosweep.py
bqpd: Fixed on the `refactor` branch!

PR #1098, opened 2017-05-14
Fix1097

### LINKED ISSUE #1097: changing order of constraints breaks model
Here's an incredibly simple model:

```python

x = Variable("x", 3)
y = Variable("y")
z = Variable("z")

m = Model(y, [
              x == z,
              y >= x + y**0.5,
             ])
sol = m.solve("mosek")
print sol.table()
```
It runs fine:
```
Using solver 'mosek'
Solving for 2 variables.
Solving took 0.0147 seconds.

Cost
----
 5.303

Free Variables
--------------
y : 5.303
z : 3

Constants
---------
x : 3

Sensitivities
-------------
x : +0.72
```

Now let's try the exact same model with the order of the two constraints reversed:

```python

x = Variable("x", 3)
y = Variable("y")
z = Variable("z")

m = Model(y, [
              y >= x + y**0.5,
              x == z,
             ])
sol = m.solve("mosek")
print sol.table()
```
now, mosek can't solve the model:
```python
Using solver 'mosek'
Solving for 2 variables.
The index subj[3] = 1 is to large
The index subj[4] = 1 is to large
python(3824,0x7fff7bd36310) malloc: *** error for object 0x2000000000000: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
Abort trap: 6
```

Doesn't occur for cvxopt. I suspect this is on our side but I haven't debugged yet.

whoburg:
Fixes #1097. Ready for review.

@bqpd, remind me why we implemented our own sparse matrix class?

### DIFF
--- a/gpkit/small_classes.py
+++ b/gpkit/small_classes.py
@@ -45,9 +45,9 @@ class CootMatrix(CootMatrixTuple):
             raise ValueError("Only positive indices allowed")
         if not self.shape:
             self.shape = [row + 1, col + 1]
-        elif row >= self.shape[0]:
+        if row >= self.shape[0]:
             self.shape[0] = row + 1
-        elif col >= self.shape[1]:
+        if col >= self.shape[1]:
             self.shape[1] = col + 1
         self.row.append(row)
         self.col.append(col)
--- a/gpkit/small_classes.py
+++ b/gpkit/small_classes.py
@@ -37,14 +37,15 @@ def matrix_converter(name):
 
 class CootMatrix(CootMatrixTuple):
     "A very simple sparse matrix representation."
-    shape = None
+    def __init__(self, *args, **kwargs):
+        super(CootMatrix, self).__init__(*args, **kwargs)
+        self.shape = [(max(self.row) + 1) if self.row else 0,
+                      (max(self.col) + 1) if self.col else 0]
 
     def append(self, row, col, data):
         "Appends entry to matrix."
         if row < 0 or col < 0:
             raise ValueError("Only positive indices allowed")
-        if not self.shape:
-            self.shape = [row + 1, col + 1]
         if row >= self.shape[0]:
             self.shape[0] = row + 1
         if col >= self.shape[1]:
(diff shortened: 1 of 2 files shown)


gpkit/small_classes.py
```diff
@@ -37,17 +37,18 @@ def to_(self):  # used in tocoo, tocsc, etc below
...
         "Appends entry to matrix."
         if row < 0 or col < 0:
             raise ValueError("Only positive indices allowed")
-        if not self.shape:
-            self.shape = [row + 1, col + 1]
-        elif row >= self.shape[0]:
+        if row >= self.shape[0]:
             self.shape[0] = row + 1
-        elif col >= self.shape[1]:
+        if col >= self.shape[1]:
```
whoburg: the core issue was the `elif` vs `if` here -- this line wasn't getting run when it needed to.

whoburg:
test models please

whoburg:
models only failing turbofan which is a known issue.

This is a high priority bug so I'm, going to merge this fix immediately. @bqpd please review nevertheless.

gpkit/small_classes.py
```diff
@@ -37,17 +37,18 @@ def to_(self):  # used in tocoo, tocsc, etc below
 
 class CootMatrix(CootMatrixTuple):
     "A very simple sparse matrix representation."
-    shape = None
+    def __init__(self, *args, **kwargs):
+        super(CootMatrix, self).__init__(*args, **kwargs)
+        self.shape = [(max(self.row) + 1) if self.row else 0,
+                      (max(self.col) + 1) if self.col else 0]
```
bqpd: Is there a purpose to setting shape in init?

gpkit/small_classes.py
bqpd: Oh dang, that is bad.

bqpd:
@whoburg, as this only occurs in (literally) corner cases, and we had that `# add constant terms` bit in genA, this might not have occurred very often? Only explanation I can think of.

PR #1109, opened 2017-05-21
Fast SP

bqpd:
@mayork, @1ozturkbe, this implements an experimental A-matrix-changing faster SP alg; see if it works on your code! This is based on 4cce21bc7e42f816d65f62164162cc681f066321, which you said worked? (don't update it to the current master!)

Known strangenesses:
  - sp.gps no longer gives you access to every gp
  - SP boundedness no longer checks for "sensitive to boundary", just "near to"

@edfan, here's one implementation of the hacky way to do this I proposed on another thread.


1ozturkbe:
Omg much excitement! Can't wait to put to the test!

bqpd:
retest this please (it passes tests on all solvers here)

bqpd:
@galbramc, it's not clear why this is failing on cvxopt on linux/mac? No test failures, no failures in parsed console output...

bqpd:
Ah, found it, for some reason the parsed console output on reynolds wasn't showing it, but macys does; forgot to commit the relaxation output changes.

bqpd:
test models please

bqpd:
@1ozturkbe, @mayork, this brings `SP_aircraft.__main__` at 53a872894d1e822eeb40ac27196c5e24c135646e from ~60s to 20s on my machine, identical solutions.

Note that with the current fast sp there isn't a `.gps` attribute, but instead a `.results` attribute, so `post_process` needs to be modified to the below:

```python
def post_process(sol):
    """
    Model to print relevant info for a solved model with relaxed constants

    ARGUMENTS
    --------
    sol: the solution to the solved model
    """

    for i in range(len(sol.program.gps)):
        varkeys = [k for k in sol.program.gps[i].varlocs
                   if "Relax" in k.models
                   and sol.program.results[i](k) >= 1.00001]
        if varkeys:
            print "GP iteration %s has relaxed constants" % i
            print sol.program.results[i].table(varkeys)
            if i == len(sol.program.gps) - 1:
                print "WARNING: The final GP iteration had relaxation values greater than 1"
```

mayork:
@bqpd love the speed this gives us but it has broken computing sensitivities to relaxed constants which is pretty critical to what we're doing right now

bqpd:
Ah dang taking a look.

bqpd:
They still work in my tests...I'm doing it a different way now, though. Are you seeing the new `_{before}` constants in your solution tables? Do they have sensitivities?

mayork:
ah okay...but where are those values stores? I'm doing this and it isn't working

```python
sol['sensitivities']
Out[6]: 
{'constants': {},
 'la': array([  1.00000000e+00,   8.06594821e-02,   2.36313415e-03, ...,
          2.76055482e+00,   0.00000000e+00,   2.97970521e+00]),
 'nu': array([  1.00000000e+00,   8.06594821e-02,   2.36313415e-03, ...,
          2.76055482e+00,   0.00000000e+00,   2.97970521e+00])}
```

bqpd:
There should be sensitivities in both the `.results` and the final solution. What file are you running, so I can try to replicate?

bqpd:
Okay, can reproduce. Debugging.

mayork:
I was running the function `run_optimal_D8` from `SP_aircraft.py`...but this seemed to be happening on all my solves from `SP_aircraft.py`

bqpd:
found it, fixing.

mayork:
it's working great now, thanks!!

bqpd:
It gets the sensitivities right (confirmed against master) but doesn't properly map them to the original variables...

bqpd:
Okay, now all the sensitivities are good

bqpd:
@whoburg, merging to bring @mayork and @1ozturkbe back on to master, but hoping to implement this in an even faster way before release, will be writing up some thoughts on the relevant issue.

PR #1111, opened 2017-05-22
Nomialmap and fastsp

bqpd:
@mayork, @1ozturkbe, also try out your code on this branch -- should be even faster.


bqpd:
test models please

PR #1091, opened 2017-05-04
Fix1081

bqpd:
closes #1081 

@whoburg, ready for review.

Note that this converts all substitutions to the key's units. Strictly speaking this only needed to be done for VectorVariables, but I feel safer having as few pint Quantities floating around as possible. However, this means that the table print out will convert to the key's units, not show the original substitution's units... @mjburton11, @1ozturkbe, @mayork, is that a problem?

PR #1112, opened 2017-05-24
fix hoburg/d8#65

bqpd:
Once I could reproduce, determined #1091 was in fact responsible for hoburg/d8#65 ! Merging once it passes tests to get d8 back on master, but ready for @whoburg review. :)


gpkit/keydict.py
```diff
@@ -131,6 +140,15 @@ def __getitem__(self, key):
     def __setitem__(self, key, value):
         "Overloads __setitem__ and []= to work with all keys"
         key, idx = self.parse_and_index(key)
+        if hasattr(value, "exp") and not value.exp:
+            value = value.value  # substitute constant monomials
```
whoburg: this seems surprisingly similar to the first two lines of `clean_value`? Maybe no way around that

gpkit/keydict.py
```diff
@@ -131,6 +140,15 @@ def __getitem__(self, key):
...
+        # pylint: disable=too-many-boolean-expressions
+        if (self.collapse_arrays and hasattr(key, "descr")
+                and "shape" in key.descr and not idx  # if a veckey, not
+                and not isinstance(value, (np.ndarray, Quantity))  # an array,
+                and not is_sweepvar(value)  # not a sweep, and
+                and not isinstance(value[0], np.ndarray)):  # not a solarray,
```
whoburg: wow

gpkit/keydict.py
bqpd: Yeah, there's this awkward overlap between what's required for a variable value and for an array that might contain variable values. I tried representing that more directly, but this was more readable.

gpkit/keydict.py
bqpd: yeah...something needs to give, here

PR #1117, opened 2017-06-19
V0.5.3

bqpd:
release v0.5.3


bqpd:
@whoburg, small changes for review!

gpkit/build.py
```diff
@@ -186,7 +186,10 @@ def look(self):
         lib_dir = pathjoin(tools_dir, "platform", mosek_platform)
         h_path = pathjoin(lib_dir, "h", "mosek.h")
         self.bin_dir = pathjoin(lib_dir, "bin")
-        self.lib_path = glob.glob(self.bin_dir+os.sep+libpattern)[0]
+        try:
+            self.lib_path = glob.glob(self.bin_dir+os.sep+libpattern)[0]
+        except IndexError:  # mosek folder found, but mosek not installed
```
whoburg: should we print a useful warning/error message in this case?

bqpd:
ah, sorry! I realized I hadn't made a release for the re-ordering bug you found+fixed a month ago, and wanted to do that as soon as it passed install tests

gpkit/build.py
bqpd: I think it's part of the "MOSEK not found" case that will be printed on a blank return

Issue #1115, opened 2017-06-04
unexpected behavior of to() for Variables


whoburg:
`gpkit.Variable.to()` does not behave as expected. See interactive session below:

```python
In [1]: from gpkit import Variable

In [2]: x = Variable("x", "ft")

In [3]: x.to("inch")  # should return something with units of inches, right?
Out[3]: gpkit.Variable(\fbox{0})

In [4]: x.to("inch").units is None  # actually, it returns an fbox with no units
Out[4]: True

In [5]: y = Variable("y")

In [6]: y.to("dimensionless")  # and now for something really unexpected...
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-6-f06329e5c1cf> in <module>()
----> 1 y.to("dimensionless")

/Users/whoburg/MIT/dev/gpkit/gpkit/nomials/nomial_core.pyc in to(self, arg)
     94         "Create new Signomial converted to new units"
     95          # pylint: disable=no-member
---> 96         return self.__class__(self.exps, self.cs.to(arg).tolist())
     97
     98     def convert_to(self, arg):

AttributeError: 'numpy.ndarray' object has no attribute 'to'
```

bqpd:
Problem is in `nomial.core` and its use of `self.__class__` to cast. 

fixed by adding the following to `Variable`:

```
    def to(self, arg):
        "Create new Signomial converted to new units"
         # pylint: disable=no-member
        return Monomial(self.exps, self.cs.to(arg).tolist())
```

(or the similar fix for nomialmap)

PR #1121, opened 2017-06-27
small issue fixes

bqpd:
@whoburg, ready for review

### DIFF
--- a/gpkit/constraints/set.py
+++ b/gpkit/constraints/set.py
@@ -184,22 +184,22 @@ class ConstraintSet(list):
                     yield yielded_constraint
 
     def subinplace(self, subs):
-        "Substitutes in place."
+        """Substitutes in place, updating self.substitutions accordingly.
+
+        Keys substituted with `subinplace` are no longer present, so if such a
+        key is also in self.substitutions that substitution is now orphaned. If
+        `subs[key]` describes some key in the ConstraintSet (i.e. one key has
+        been substituted for another), then a substitution is added, mapping
+        the orphaned value to this new key; otherwise, an error is raised.
+        """
         subs = {k.key: getattr(v, "key", v) for k, v in subs.items()}
         for constraint in self:
             constraint.subinplace(subs)
-        for key in subs:
-            if key not in self.substitutions:
-                continue
-            if hasattr(subs[key], "key"):
-                self.substitutions[subs[key]] = self.substitutions[key]
+        for key, value in subs.items():
+            if key in self.substitutions:
+                valkey, _ = self.substitutions.parse_and_index(value)
+                self.substitutions[valkey] = self.substitutions[key]
                 del self.substitutions[key]
-            else:
-                raise ValueError("the substitution {%s: %s} is invalidated"
-                                 " by the subinplace {%s: %s}, because"
-                                 " %s does not have a `key` attribute"
-                                 % (key, self.substitutions[key],
-                                    key, subs[key], subs[key]))
         self.unique_varkeys = frozenset(subs[vk] if vk in subs else vk
                                         for vk in self.unique_varkeys)
         self.reset_varkeys()
--- a/gpkit/constraints/set.py
+++ b/gpkit/constraints/set.py
@@ -214,7 +214,9 @@ class ConstraintSet(list):
     def as_posyslt1(self, substitutions=None):
         "Returns list of posynomials which must be kept <= 1"
         posylist, self.posymap = [], []
-        for constraint in self:
+        for i, constraint in enumerate(self):
+            if not hasattr(constraint, "as_posyslt1"):
+                raise_badelement(self, i, constraint)
             posys = constraint.as_posyslt1(substitutions)
             self.posymap.append(len(posys))
             posylist.extend(posys)
--- a/gpkit/constraints/set.py
+++ b/gpkit/constraints/set.py
@@ -44,6 +44,7 @@ class ConstraintSet(list):
                         # so we can catch them (see above) in ConstraintSets
             if hasattr(self[i], "substitutions"):
                 self.substitutions.update(self[i].substitutions)
+                self[i].substitutions = self.substitutions
         self.reset_varkeys()
         if substitutions:
             self.substitutions.update(substitutions)
--- a/gpkit/nomials/nomial_core.py
+++ b/gpkit/nomials/nomial_core.py
@@ -2,8 +2,8 @@
 from .data import NomialData
 from ..small_classes import Numbers
 from ..small_scripts import nomial_latex_helper
-from ..small_scripts import mag, unitstr
-from ..repr_conventions import _str, _repr, _repr_latex_
+from ..small_scripts import mag
+from ..repr_conventions import _str, _repr, _repr_latex_, unitstr
 
 
 class Nomial(NomialData):
@@ -15,6 +15,7 @@ class Nomial(NomialData):
     __str__ = _str
     __repr__ = _repr
     _repr_latex_ = _repr_latex_
+    unitstr = unitstr
 
     def str_without(self, excluded=None):
         "String representation excluding fields ('units', varkey attributes)"
@@ -38,7 +39,7 @@ class Nomial(NomialData):
                 cstr = [cstr] if (cstr != "1" or not varstrs) else []
                 mstrs.append("*".join(cstr + varstrs))
         if "units" not in excluded:
-            units = unitstr(self.units, " [%s]")
+            units = self.unitstr(" [%s]")
         else:
             units = ""
         return " + ".join(sorted(mstrs)) + units
@@ -61,7 +62,7 @@ class Nomial(NomialData):
         if "units" in excluded:
             return " + ".join(sorted(mstrs))
 
-        units = unitstr(self.units, r"\mathrm{~\left[ %s \right]}", "L~")
+        units = self.unitstr(r"\mathrm{~\left[ %s \right]}", "L~")
         units_tf = units.replace("frac", "tfrac").replace(r"\cdot", r"\cdot ")
         return " + ".join(sorted(mstrs)) + units_tf
 
--- a/gpkit/nomials/variables.py
+++ b/gpkit/nomials/variables.py
@@ -72,6 +72,11 @@ class Variable(Monomial):
         "a Variable's descr is derived from its VarKey."
         return self.key.descr
 
+    def to(self, arg):
+        "Create new Signomial converted to new units"
+         # pylint: disable=no-member
+        return Monomial(self).to(arg)
+
     def sub(self, *args, **kwargs):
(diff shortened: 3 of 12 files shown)


whoburg:
Will review. Weird, what's going on with `macys_Sierra,cvxopt`? It's failing on a call to mosek??

whoburg:
I'll find this a little easier to review if we split the Python 3 compatibility (which touches a huge number of files) into a separate PR. I'm going to make that change now -- 

 - Will rebase this to drop the two Python 3 related commits and force push
 - will open a new PR with the two Python 3 commits

whoburg:
just force pushed, update local copies of this branch

bqpd:
good point/plan re: splitting this!

bqpd:
test models please

bqpd:
(models failure with mosek_cli macys and reynolds which I can't reproduce)
test models please

bqpd:
the failures are semi-reproducible: turbofan_engine_flight_profile_fleet_py_mosek_cli returns `NEAR_DUAL_FEAS`

However, in the first trial a solve with MOSEK also failed...

Perhaps we should allow for NEAR_DUAL_FEAS etc to be treated like OPTIMAL?

whoburg:
I'm against passing it through as `OPTIMAL`, but I'm ok with allowing it with a printed warning.

whoburg:
argh, this somehow got the python3 fixes again, should I rebase again?

whoburg:
I just force pushed -- @bqpd please delete your local copy of this branch and re-fetch it from origin, so we can avoid those python3 commits creeping back on

bqpd:
Yup, my bad. You kept the new small fixes I'd added?

whoburg:
yep, I assume you mean 6bcd34b7904c78184b45861f874d6fd018445f86, it's in the commit hx

gpkit/constraints/set.py
```diff
@@ -44,6 +44,7 @@ def __init__(self, constraints, substitutions=None):
                         # so we can catch them (see above) in ConstraintSets
             if hasattr(self[i], "substitutions"):
                 self.substitutions.update(self[i].substitutions)
+                self[i].substitutions = self.substitutions
```
whoburg: I still think we don't have any good documentation of the behavior of substitutions for modular models with submodels. In particular , the fact that the submodel's substitutions get mutated to contain the larger model's subs is not obvious. I think this documentation should live all the way up at the read the docs level.

gpkit/constraints/set.py
```diff
@@ -183,22 +184,22 @@ def flat(self, constraintsets=True):
...
     def subinplace(self, subs):
-        "Substitutes in place."
+        """Substitutes in place, updating self.substitutions accordingly.
+
+        Keys substituted with `subinplace` are no longer present, so if such a
+        key is also in self.substitutions that substitution is now orphaned. If
+        `subs[key]` describes some key in the ConstraintSet (i.e. one key has
+        been substituted for another), then a substitution is added, mapping
+        the orphaned value to this new key; otherwise, an error is raised.
+        """
```
whoburg: great docstring!

whoburg: The last line of docstring says an error is raised, but the `ValueError` got removed according to the diff. So what raises that error? Is the docstring incorrect?

gpkit/constraints/set.py
```diff
@@ -214,7 +215,9 @@ def reset_varkeys(self):
     def as_posyslt1(self, substitutions=None):
         "Returns list of posynomials which must be kept <= 1"
         posylist, self.posymap = [], []
-        for constraint in self:
+        for i, constraint in enumerate(self):
+            if not hasattr(constraint, "as_posyslt1"):
+                raise_badelement(self, i, constraint)
             posys = constraint.as_posyslt1(substitutions)
```
whoburg: seems like an obvious place for a try/catch

```python
try:
    posys = constraint.as_posyslt1(substitutions)
except AttributeError:
    assert not hasattr(constraint, "as_posyslt1")  # might not need this, just making sure this is the cause
    raise_badelement(self, i, constraint)
```
?

whoburg: feel free to reject, I see that maybe `constraint.as_posyslt1` could exist and happen to raise its own `AttributeError`, in which case catching is not appropriate.

gpkit/nomials/nomial_core.py
```diff
@@ -61,7 +62,7 @@ def latex(self, excluded=None):
         if "units" in excluded:
             return " + ".join(sorted(mstrs))
 
-        units = unitstr(self.units, r"\mathrm{~\left[ %s \right]}", "L~")
+        units = self.unitstr(r"\mathrm{~\left[ %s \right]}", "L~")
         units_tf = units.replace("frac", "tfrac").replace(r"\cdot", r"\cdot ")
```
whoburg: seems like this replacement stuff belongs in `unitstr`? Does it really only apply to a `Nomial` and no other GPkit objects?

gpkit/nomials/variables.py
```diff
@@ -72,6 +72,11 @@ def descr(self):
         "a Variable's descr is derived from its VarKey."
         return self.key.descr
 
+    def to(self, arg):
+        "Create new Signomial converted to new units"
+         # pylint: disable=no-member
+        return Monomial(self).to(arg)
+
```
whoburg: #1115 mentions a similar fix is needed for NomialMap?

gpkit/repr_conventions.py
```diff
@@ -14,3 +15,16 @@ def _str(self):
...
+
+
+def unitstr(units, into="%s", options="~", dimless=""):
+    "Returns the string corresponding to an object's units."
+    if hasattr(units, "units") and isinstance(units.units, Quantity):
+        units = units.units
+    if not isinstance(units, Quantity):
+        return dimless
+    rawstr = (u"{:%s}" % options).format(units.units)
+    if str(units.units) == "count":
```
whoburg: seems like this special case ought to operate on `units` or `rawstr`, not on `units.units`?

gpkit/repr_conventions.py
```diff
@@ -14,3 +15,16 @@ def _str(self):
...
+        rawstr = u"count"  # TODO: remove when pint issue #356 is resolved
+    units = rawstr.replace(" ", "").replace("dimensionless", dimless)
+    return into % units or dimless
```
whoburg: not knowing the order of ops between `%` and `or`, parens would help here. Is this
`return (into % units) or dimless`
or
`return into % (units or dimless)`
?

gpkit/solution_array.py
```diff
@@ -302,7 +303,7 @@ def results_table(data, title, minval=0, printunits=True, fixedcols=True,
                                   model + r"}} \\"])
             oldmodel = model
         label = var.descr.get('label', '')
-        units = var.unitstr() if printunits else ""
+        units = var.unitstr(" [%s] ") if printunits else ""
```
whoburg: I thought we agreed to remove the square brackets awhile back? Why are they being added back in here?

whoburg: Oh I see, `VarKey.unitstr` used to have a default kwarg of `into=" [%s] "`.

gpkit/constraints/set.py
bqpd: Yeah, for now I thinking that "quacking like a duck" includes having a given method...

gpkit/solution_array.py
bqpd: yup!

gpkit/repr_conventions.py
bqpd: good point! We actually don't care which right now (nothing calls on it), but I think I like the latter better?

gpkit/repr_conventions.py
bqpd: see https://github.com/hgrecco/pint/issues/356 as mentioned in the comment

gpkit/nomials/variables.py
bqpd: the exact same fix will work, actually! I hadn't thought it through when I posted that comment

gpkit/nomials/nomial_core.py
bqpd: replacement stuff...the latex formatting? Everywhere else calls upon Nomial to establish the proper formatting, so this is the only place it's specified

gpkit/constraints/set.py
bqpd: Nope, we now (more cleanly) raise the error during `self.substitutions.parse_and_index(value)`

gpkit/constraints/set.py
bqpd: 1) self[i] might be the ConstraintSetted version of constraint

2) I just changed it so that the substitutions _would_ get mutated precisely because that was the behaviour people in learn3 expected!

bqpd:
test models please

gpkit/constraints/set.py
whoburg: 1. oooh, duh. Got it.

2. I know, it's a good improvement, and much more understandable -- I'm just saying that it's still not documented, and I think users still find this confusing and would appreciate a clear explanation of how it works. To be clear, I totally agree with doing the mutation.

gpkit/nomials/nomial_core.py
whoburg: I'm referring to the line `units_tf = units.replace("frac", "tfrac").replace(r"\cdot", r"\cdot ")`. Wondering why that is living in `Nomial`.

gpkit/nomials/variables.py
whoburg: sooo, should it be added to NomialMap before we close #1115?

gpkit/repr_conventions.py
whoburg: My point is that above you have `units = units.units` inside a conditional `if hasattr(units, "units")`. So I'm led to believe that `units` might not even have a `units` attribute. But then we do `if str(units.units) == "count":`. Why not `if str(units) == "count":`?

gpkit/constraints/set.py
bqpd: ah, yup, we should document it :) in Advanced Commands

gpkit/repr_conventions.py
whoburg: I agree the latter makes more sense, but note that the former is how Python interprets it with no parens (I just tested)

gpkit/repr_conventions.py
bqpd: Oh, it's a pint version thing: we make sure that units is a Quantity, which means units.units is a Unit, and the str representation of Unit("count") is stable across pint versions while the str representation of Quantity("count") is not...

whoburg: ooooh, so if we make it to the special case then the original input would have had `units.units.units`? I (think I) see now, sorry for the confusion.

PR #1133, opened 2017-07-12
don't re-bring costvals into sp substitutions

bqpd:
@whoburg. @mjburton11 , fixes #1131. Ready for review.

Issue #1131, opened 2017-07-11
substitutions not deleting


mjburton11:
@bqpd @whoburg ahh! need you're help on this quick! 

Running `gasmaleperf.py` on jho repo on branch `newsolves`.  It appear that deleting a substitution only temporarily deletes it re-inserts it when trying to solve.  I'm convinced this is why I got `DUAL_INFEAS` because I know it will solve with higher payloads.  Below is the ipython output (I also saw the same behavior as a script):

```python 
In [1]: %run gasmaleperf.py
Beginning signomial solve.
Allowing negative dual variable(s) as small as -1.69108000173e-12.
Allowing negative dual variable(s) as small as -1.64037945553e-13.
Solving took 4 GP solves and 24.2 seconds.
144.324433952

In [2]: M.cost = 1./M["W_{pay}"]

In [3]: del M.substitutions["W_{pay}"]

In [4]: M.substitutions["W_{pay}"]
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
...
KeyError: W_{pay}_Mission/Aircraft

In [5]: sol = M.localsolve("mosek")
Beginning signomial solve.
RuntimeWarning: final status of solver 'mosek' was 'DUAL_INFEAS_CER', not 'optimal'.

The solver's result is stored in model.program.solver_out. A result dict can be generated via program._compile_result(program.solver_out).

In [6]: M.substitutions["W_{pay}"]
Out[6]: 10
```

mjburton11:
any word on this? @whoburg need to fix for JHO performance metrics

whoburg:
@bqpd this is relatively urgent, need for some Air Force questions that we need to answer by tomorow.

bqpd:
@mjburton11, I can't run `gasmaleperf.py`, are you on a particular branch of gpkit-models

mjburton11:
No, I was running it on master

mjburton11:
what's the error?

bqpd:
`KeyError: m_{fac--fit}_Mission/Climb/FlightSegment/AircraftPerf/WingAero`

mjburton11:
That's a problem with `fit_constraintset.py` in gpfit...

mjburton11:
Is it an issue in process result?

bqpd:
yup, good call!

bqpd:
so it's solving, but you're looking for a variable that's not there

bqpd:
[...]

PR #1149, opened 2017-08-20
add test for #1131

bqpd:
as requested in #1133. @whoburg, ready for review


bqpd:
test models please

bqpd:
test models please

whoburg:
so, an alternative proposal would be to throw an error when the user sets the cost to a fixed variable. Note there's a delicate issue here in that we should still support feasibility solves (where the cost is a constant). So to be a bit more refined, we should throw the error when solving a model where (the cost is set to a fixed variable AND the fixed variable's value does not appear in the model's substitutions).

mjburton11:
@whoburg the example you posted is actually the use case I want.  For example, in JHO my initial value for `t_Loiter` is 6 days.  But I want to delete that substitution and make `t_Loiter` the objective.

whoburg:
The discussion above occurred after this got merged -- did the issue about cost being a fixed variable get captured somewhere?

bqpd:
It's come up again in a few discussions of how to handle substitutions, particularly #1237 and #1218

In general, the cost may contain substitutions that aren't contained anywhere else -- this is the best way to implement composite objectives, for example. We could ban that easily, but I think it's better to just 'pop' values after their use, as in #1237

PR #1152, opened 2017-08-25
Symmetric Variable / VarKey ==

### LINKED ISSUE #1138: Variable and Varkey __eq__ disagree
These should be symmetric:

```python
In [1]: from gpkit import Variable

In [2]: W = Variable("W", 5, "lbf", "weight of 1 bag of sugar")

In [3]: W.key == W
Out[3]: True

In [4]: W == W.key
Out[4]: False
```

bqpd:
Closes #1138 

@whoburg ready for review


bqpd:
Test models please

bqpd:
I think `__eq__` should distinguish between things which have a `.key` attribute and things which are VarKeys; a string a Variable and a VarKey are not equal, but they may both refer to the same VarKey via either their `.key` attribute or the VarKey's `.keys` attribute

PR #1168, opened 2017-09-26
don't assume there'll be a label

bqpd:


whoburg:
why wasn't this tested? It broke our tests on master.

PR #1221, opened 2017-11-09
fold generate_nula into _compile_result

### LINKED ISSUE #1142: model.program._compile_result() broken
The error message for `unknown` suggests using `model.program._compile_result()` to access the latest variable values, however this is currently not working because `solver_out` has no key `"nu"`. Workaround is to change `nu, la = solver_out["nu"], solver_out["la"]` to `la = solver_out["la"]`.

Unfortunately, that breaks the sensitivities part of `_compile_result` (not super important to me, I can just comment it out but probably should be addressed).

bqpd:
Closes #1142


bqpd:
test models please

PR #1226, opened 2017-11-09
remove string substitutions

### LINKED ISSUE #857: remove string substitution
and find a better way to rename variables.

bqpd:
Closes #857


bqpd:
test models please

bqpd:
test models please

PR #1233, opened 2017-11-11
Check Monomial Equalities

bqpd:


bqpd:
test models please

bqpd:
test models please

bqpd:
@mjburton11, this branch does some additional analysis, and it's saying jho isn't fully bounded (though still solving). Could you check to see if the bounds it's reporting as missing make sense?

mjburton11:
These are the bound errors I see:

```python
ValueError: Geometric Program is not fully bounded:
  (W/S)_Mission has no upper bound, but would gain it from any of these sets of bounds: [(S_Mission/Aircraft/Wing/Planform, 'lower'), (MTOW_Mission, 'upper')]
  l_v_Mission/Aircraft/Empennage/VerticalTail has no lower bound, but would gain it from any of these sets of bounds: [(V_v_Mission/Aircraft/Empennage/VerticalTail, 'lower'), (S_Mission/Aircraft/Empennage/VerticalTail/Planform.2, 'upper'), (b_Mission/Aircraft/Wing/Planform, 'lower'), (S_Mission/Aircraft/Wing/Planform, 'lower')]
  MTOW_Mission has no upper bound, but would gain it from any of these sets of bounds: [((W/S)_Mission, 'upper'), (S_Mission/Aircraft/Wing/Planform, 'upper')]
  V_v_Mission/Aircraft/Empennage/VerticalTail has no lower bound, but would gain it from any of these sets of bounds: [(b_Mission/Aircraft/Wing/Planform, 'upper'), (S_Mission/Aircraft/Empennage/VerticalTail/Planform.2, 'lower'), (S_Mission/Aircraft/Wing/Planform, 'upper'), (l_v_Mission/Aircraft/Empennage/VerticalTail, 'lower')]
```

In `jho.py`

- `(W/S)` is bounded by an equality constraint on line 335.  It should probably post processed because it is not being used anywhere else.  This is a legacy constraint since before `process_result` was a thing
- `l_v` is bounded by an equality constraint beginning on line 54.  I believe this is a valid lower bound.
- `MTOW` actually doesn't have an upper bound.  I usually minimize it so I don't think I've noticed it before.  In this case I am maximizing endurance.  I have done this before and my result appeared accurate but now I wonder...
- `V_v` is normally a substitution which appears to be missing.  This may be why `l_v` is unbounded as well.

Usually I use `jho.py` only to get performance numbers by calling `print_perf.py` so `l_v` and `V_v` are normally bounded by substitutions that are the actual dimensions of the plane.

bqpd:
@mjburton11 awesome! Yeah this checks bounds through equalities, instead of assuming equalities mean something is bounded. If those are all reasonable things to error on, could you fix them so that I can merge this?

mjburton11:
test models please

bqpd:
looks like V_v / l_v still need a bound in print_perf...

mjburton11:
sorry...

mjburton11:
Can you wait to merge this, I'm having some problems running `wing_test.py` on gplibrary.

bqpd:
sure thing!

bqpd:
test models please

mjburton11:
Here's a list of errors that I see from `wing_test.py` on branch `parsing`:

```python
ValueError: Geometric Program is not fully bounded:
  cmac_Wing/Planform has no upper bound, but would gain it from any of these sets of bounds: [(Re_WingAero, 'upper')] or [(croot_Wing/Planform, 'upper')]
  Re_WingAero has no upper bound, but would gain it from any of these sets of bounds: [(cmac_Wing/Planform, 'upper')]
  AR_Wing/Planform has no upper bound, but would gain it from any of these sets of bounds: [(S_Wing/Planform, 'lower')]
  croot_Wing/Planform has no upper bound, but would gain it from any of these sets of bounds: [(cmac_Wing/Planform, 'upper')] or [(b_Wing/Planform, 'lower')]
  b_Wing/Planform has no lower bound, but would gain it from any of these sets of bounds: [(S_Wing/Planform, 'lower')] or [(S_Wing/Planform, 'lower')] or [(S_Wing/Planform, 'lower')] or [(S_Wing/Planform, 'lower')] or [(croot_Wing/Planform, 'upper'), (S_Wing/Planform, 'lower')] or [(S_Wing/Planform, 'lower')]
  S_Wing/Planform has no lower bound, but would gain it from any of these sets of bounds: [(b_Wing/Planform, 'lower')] or [(b_Wing/Planform, 'lower')] or [(b_Wing/Planform, 'lower'), (AR_Wing/Planform, 'upper')] or [(b_Wing/Planform, 'lower')] or [(b_Wing/Planform, 'lower')] or [(b_Wing/Planform, 'lower')]
```

My objective is `C_d`.  I think: 
- `Re` and `cmac` are actually bounded by the objective.  But this may not always be the case so `Re` should be defined by the `==` equality. 
- `b`, `S`, `AR` are defined by `AR == b**2/S`.  This has to be an equality because large AR is good for aero but bad for structures, but I claim that all 3 are bounded by this constraint.  That may or may not be true
- `c_{root}` I think is tied to the `AR` constraint

bqpd:
checking it out!

bqpd:
Hmm, the algorithm searches through monomial equalities for which bounds are actually preserved, and so far it hasn't been wrong...I think the minimum set of bounds for this is `(croot[or cmac or Re]_Wing/Planform, 'upper')  and (S_Wing/Planform, 'lower')` or `(b_Wing/Planform, 'lower') and (AR_Wing/Planform, 'upper')`

bqpd:
is there any chance that AR has no upper bound and b no lower bound (i.e. because there's not a tip loading model or something?)

mjburton11:
No there is a tip loading model.   Let me keep looking though

mjburton11:
So is it telling me that `Re` should be `Re <=` instead of `Re == `

bqpd:
actually it looks like just lower bounding b works. solves with `W["b"] >= 0.01*units("ft")`

bqpd:
nope nope it's not saying the equalities should be relaxed to inequalities; it's pointing out that equalities can't lower bound more than one variable at a time (and similarly upper bound one variable)

bqpd:
[note re: lower-bounding b, a lower bound on S would also work]

bqpd:
potentially relevant for future debugging exercises of this kind: https://github.com/convexengineering/gpkit/issues/1235

bqpd:
test models please

mjburton11:
I don't understand why it's requiring me to put a lower bound on `b` when it the solution shows that `b` is clearly not reaching the lower bound

bqpd:
That's kind of the test of this PR :p is formally checking these things too annoying? so far such analysis has only pointed out interesting things...I think in this one the way to look at it might be that there should be a lift constraint?

mjburton11:
Ohhh that actually makes sense to me now.  And that was interesting. So let's keep it

mjburton11:
Thats why we didn't see solar or jho fail in the same way

bqpd:
yup!

bqpd:
ready to merge this? the parsing branch will start failing its [new unit tests](https://github.com/convexengineering/gplibrary/pull/138)

mjburton11:
Hmmm let's merge then we can set up the unit tests so they work with this

bqpd:
(in addition to the need to lower-bound, S, tail_emp in tails_tests is currently failing on master)

bqpd:
oops I just realized after merging that you meant merge `parsing`, not this...sorry!

PR #1270, opened 2018-03-20
sol() now always returns Quantities (even dimensionless ones)

### LINKED ISSUE #1252: Calling units of dimensionless variable should not throw error
This is relatively straightforward. Currently, if we have a dimensionless variable like aspect ratio, calling
```python
sol('AR').magnitude
```
throws an error
```python
AttributeError: 'numpy.float64' object has no attribute 'magnitude'
```
In the solution dicts, we should have uniformity so that we can call magnitude on all of them without issues.

bqpd:
Closes #1252


bqpd:
test models please

bqpd:
test this please

bqpd:
test models please

bqpd:
test models please

bqpd:
test models please

PR #1273, opened 2018-03-20
allow specification of mosek rootdir with MSKHOME environment variable

### LINKED ISSUE #1254: alternative install location for mosek
/private/var/folders/r0/5bhjqvqj3b17jr0m8n6lwpsm0000gn/T/pip-build-UydGZI/gpkit/

Not a big deal, but it might be nice to allow mosek to be installed somewhere other than $HOME, maybe through a command line arg or environment variable.  mosek v8 doesn't assume installation in $HOME like some past versions did. Just a personal preference - I like to keep these types of programs in a separate folder to not clutter my home directory - but like I said not a big deal.

Perhaps that would also allow for mosek to be installed through conda, for those that to go that route.

bqpd:
Closes #1254


bqpd:
test models please

Issue #1292, opened 2018-03-29
Vectorize doesn't work for matrix vector variable


mjburton11:
I currently can't do the following: 

```python
with Vectorize([1, 3]):
    x = Variable("x")
```

but I can do
```python
x = VectorVariable([1, 3], "x")
```

Error: 

```python
/Users/mjburton11/MIT/GPKIT/gpkit/gpkit/nomials/variables.pyc in __new__(cls, shape, *args, **descr)
    159 
    160         arraykey = VarKey(**descr)
--> 161         vl = np.empty(shape, dtype="object")
    162         it = np.nditer(vl, flags=['multi_index', 'refs_ok'])
    163         while not it.finished:

TypeError: 'list' object cannot be interpreted as an index
```

mjburton11:
Although I can do: 

```python
with Vectorize(1):
    with Vectorize(3):
        x = Variable("x")
```

bqpd:
Yup, vectorize has to be nested at present. I think this will stay part of the requirement, since one thing about the way vectorized has to work is that the innermost dimension is the leftmost, so that calls referencing vectors work (e.g., w_tip = w[-1] in the beam, or see the test below) work even inside a vectorization. In your example above this means that your last x is equivalent to `x = VectorVariable([3, 1], "x")`.

If vectorization accepted a list as input I think this would only be more confusing; @mjburton11 what do you think?

```python
class TestVectorize(unittest.TestCase):
    """TestCase for gpkit.vectorize"""

    def test_shapes(self):
        with gpkit.Vectorize(3):
            with gpkit.Vectorize(5):
                y = gpkit.Variable("y")
                x = gpkit.VectorVariable(2, "x")
            z = gpkit.VectorVariable(7, "z")

        self.assertEqual(y.shape, (5, 3))
        self.assertEqual(x.shape, (2, 5, 3))
        self.assertEqual(z.shape, (7, 3))
```

mjburton11:
Yeah that's fine with me.  I just didn't understand the nested Vectorization until I played with it a bit more.

mjburton11:
But possible a related issue can be found in `solar` commit 295051844c029d0b71bd0d33eb719c0069351cbb

```python
/Users/mjburton11/MIT/GPKIT/gpkit/gpkit/nomials/variables.pyc in newlinkedfn(c)
    193     def newlinkedfn(c):
    194         "Linked function that pulls out a particular index"
--> 195         return np.array(linkedfn(c))[i]
[...]

PR #1293, opened 2018-03-30
allow for solves for matrix vector variables

mjburton11:
Allows for matrix vector variables with linked values


bqpd:
per https://github.com/convexengineering/gpkit/issues/1292, linked works fine for matrices.

PR #1299, opened 2018-04-20
warn if no mosek versions found during build

bqpd:


bqpd:
retest this please

bqpd:
@galbramc, seems to have failed on trying to pip install coverage?

galbramc:
No idea why converge would have problems installing. I've seen pip have problems the past couple of days.

Could you include MSKHOME in the error message so it is really clear?

galbramc:
That looks good. Minor thing. Could you add a colon or something to indicate the directory name, i.e.
```
" were found in the mosek directory : " + rootdir
```
Some sort of separation from the sentence and the value would be good.

galbramc:
Also, is there anything we can do about all the crazy error message we get from I believe ctypesgen? It's really hard to tell that things are actually working correctly with thousands of error messages flying by on the terminal.

bqpd:
Unfortunately we're running ctypesgen in the shell...I don't know of a good cross-platform way to capture STDOUT/STDERR from it.

galbramc:
I think you can change the settings on this pylint warning: Too many return statements (7/6)

bqpd:
added the formatting

galbramc:
That's a bummer. Do you know why ctypesgen makes all those crazy messages? They are not really sensical...

bqpd:
:( I don't. Actually, I bet there's a way to do it in modified_ctypesgen.py...

galbramc:
Where did that come from? And how is it modified?

bqpd:
done! I think there's a few lines I changed? anyway, I added a few more.

galbramc:
Lastly, any clue why expopt.so does not work with MinGW 7? I tried it and it appear to compile, but loading the shared library seemed to fail.

galbramc:
The build is way cleaner now. Take a look at the terminals. The only confusing message still there is:

```
No filename given.
Usage:
     mskexpopt FILENAME [-primal] [-dual] [-p parameterfile]
Return code: 1052
Description: MSK_RES_ERR_FILE_OPEN [An error occurred while opening a file.]
/usr/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'install_requires'
  warnings.warn(msg)
Could not load settings file.
```
Are you using mskexpopt just to see if it is in the path? Could you use msktestlic instead? That creating an error or passing is far more informative anyways.

galbramc:
I don't understand the cvxopt errors on WIndows10. cvxopt appears to be installed before building gpkit, but the tests can't find it?!?!

galbramc:
test gpkit please

galbramc:
test gpkit please

galbramc:
test gpkit please

galbramc:
test gpkit please

bqpd:
re: minGW it seems to be a difference in some kind of struct reading where ctypesgen can't handle gcc's output format?

re: mskexpopt, I could use msktestlic! I kinda like directly calling the executable that's called for solves, though...

galbramc:
I see yeah that makes sense calling mskexpopt. Can you give it some dummy argument though so it does not generate an error message? Maybe just --version of something like that?

bqpd:
I want to check its return code, though...it has different ones for "can't find mosek binary" vs "license experied" etc

galbramc:
Oh ok. It's just unfortunate that it looks like there is an error when there really isn't the way you are probing it right now...

bqpd:
Yeah. I guess maybe the right way to do it would be to solve a dummy problem?

galbramc:
That would do it. Something really trivial. Again, it's just really confusing to see error messages flying by and trying to sort out which errors are real and which ones are expected side affects.

Issue #1289, opened 2018-03-27
pickle compatibility


bqpd:
- [x] for SolutionArrays (https://github.com/convexengineering/gpkit/issues/927)
  - [ ] for Models/Nomials (#293)

bqpd:
To add pickle compatibility for models, will need to implement https://github.com/convexengineering/gpkit/issues/1289 again

Issue #1282, opened 2018-03-21
proposed deprecations


bqpd:
- [ ] subinplace
- [x] zero_lower_unbounded
- [ ] markdown (low priority)
- [x] composite objective

bqpd:
there's a lot of code devoted to model printing that could be cleaned up as well.

PR #1315, opened 2018-05-07
pickles solutions and signomials; loses linked fns

### LINKED ISSUE #438: Solution arrays can't print 2d variables

### LINKED ISSUE #916: A better way of comparing solutions
One thing I am often interested in is how the solution of a model changes when I add constraints/change subsitutions/use different sub-models etc.

My current way of tracking this is to create a .sol that I update and commit each time I make a change  (and a git pre-commit hook ensures I don't forget to update the .sol each time I commit). `git diff` then allows me to see what has changed, and by how much.

This method is somewhat rudimentary and definitely not perfect. It would be nice to have something that does this more elegantly, and had other features, for example showing percentage changes of each variable, and even ranking the variables by percentage change. I think a git-centric solution might be the way to do it, but I am open to other suggestions.

### LINKED ISSUE #1063: Easy spreadsheet outputs
It would be really nice if GPkit was able to save the terminal outputs into some kind of spreadsheet (most simply a csv), labelled with the timestamp and the model name. This would be especially interesting for large SP models *ehem d8 ehem*, and could be developed further by putting things about runtime and number and characteristics of GP iterations in separate sheets. 
It would facilitate the manipulation/plotting of results outside of the terminal. Furthermore, I think it may be important for some people to never lose the results of a solve; allowing spreadsheet outputs would facilitate this, since terminal outputs are fleeting. 
@mjburton11 and @mayork also think this is a good idea. Storage is cheap; let's take advantage of it :P Lmk if this is realizable @bqpd.

bqpd:
Closes #916
Closes #438
Closes #1063

Partially completes #1289 and #1282

## TODO
  - see if there's a convenient way to save linkedfns...
    - dill, but it requires `__reduce__` support
      - and can't re-run a model after loading it

## Example Usage

```python
sol = m.solve()
sol.save("stored_sol.p")
sol2 = m2.solve()
print sol.diff("stored_sol.p")  # or equivalently, print sol.diff(sol2)
```
```
Solution difference
(positive means this solution is bigger)
-------------------
  k : -8%   form factor
C_D : -7%   Drag coefficient of wing
  D : -4%   total drag force
C_L : -3%   Lift coefficent of wing
  V : +2%   cruising speed
 Re : +2%   Reynold's number
W_w : +0.7% wing weight
  A : +0.5% aspect ratio
C_f : -0.3% skin friction coefficient
  S : +0.2% total wing area
  W : +0.2% total aircraft weight

Variable(s) of this solution which are not in the argument:
  pi
```


bqpd:
test models please

bqpd:
test models please

bqpd:
Unit tests should pass when we're using MOSEK 8 on all machines.

bqpd:
test models please

bqpd:
test models please

bqpd:
test models please

PR #1319, opened 2018-05-14
Making minimal Sankey diagram for solar.

### LINKED ISSUE #1314: Add to documentation that Sankey only works in iPython notebooks.
Firstly, the syntax must be changed, because the Solar code syntax has changed. 
```python
    SP = True
    Vehicle = Aircraft(Npod=3, sp=SP)
    M = Mission(Vehicle, latitude=[20])
    M.cost = M[M.aircraft.Wtotal]
    from gpkit.interactive.sankey import Sankey
    Sankey(M).diagram(M.solar.Wtotal)
```

1ozturkbe:
closes #1314

@bqpd just decided to make a PR with the required changes. One step remains: how do you save an svg from the SankeyWidget, so we can update the picture? I have been trying to figure this out and failing. Lmk.


bqpd:
Nice, that's lovely!

bqpd:
Could you update the svg in that directory as well?

bqpd:
(use the browser inspector to grab the svg element, copy-past into the text document...or I can do it)

1ozturkbe:
I'll take care of it. Thanks!

bqpd:
What actually...the SVG attached doesn't have the right unicode for the line labels...(a, b, etc)

bqpd:
You can see the preview on github with `?`s instead of the labels...

1ozturkbe:
Interesting... I checked that I could load in chrome but I didn't see the error, my b. I don't know why that is the case, but I'll check it.

Also, I've been generating the Sankey plots for solar and I get results that are not insightful (there are just too many constraints, and the Sankey diagrams look somewhat incoherent). Maybe we need to switch to a dumber model.

1ozturkbe:
@bqpd it would be great to have a way to test the code included in the GPkit documentation once in a while. None of the plots in the Sankey docs can be recreated using the code in the docs. Immensely frustrating.

bqpd:
Yeah there's an issue open for interactives testing...Sankey will present
an especial difficulty :-/ I could add it to the new-version checklist?

bqpd:
Maybe add the solar commit you're using to the docs just to ensure
reproducibility?

1ozturkbe:
Unless @mjburton11 wants to fix the code in the Sankey documentation, I may actually merge all of my thesis code into gplibrary, and make Sankey diagrams with those. The Sankey diagrams of Solar are overly complicated and don't provide much intuition.

1ozturkbe:
I'm using latest of master on solar.

bqpd:
master is a moving target though, so we should refer to an explicit commit
and update it periodically. Feel free to create the gplibrary PR!

1ozturkbe:
You're right... Honestly I think it's probably better to make a gpkitdocs branch within solar or whatever so we can refer to the branch itself.

bqpd:
Yeah that sounds like a better idea!

bqpd:
@1ozturkbe wanna update the last two svgs and I'll make the branch?

1ozturkbe:
alright... let me figure this out.

1ozturkbe:
Also, Sankey diagrams make iPython crash like no other. It's pissing me off.

bqpd:
is it the really big ones? or maybe updating ipython notebook would fix it?

1ozturkbe:
I can't update the one with aircraft span (M.aircraft.b), but I have created a branch within solar called gpkitdocs with the notebook so you can update it if you'd like. Because I gave up trying to generate that plot; even fully updated my Chrome cannot handle it. Otherwise this is ready to go.

1ozturkbe:
Damn it, this is precisely why the model numbering is annoying. If I ever do this for not my own models on my own volition, pinch me.

1ozturkbe:
Updated all. Merge when ready @bqpd.

PR #1327, opened 2018-05-23
add reset_modelnumbers function

### LINKED ISSUE #1320: Issues diffing solutions of multiple instances of a model
I guess 'diffing' can have a lot of issues. 
If you run a model and save the solution, and then build another model and save the solution of that as well, the solutions have a different set of varkeys.
The first solution will have this varkey: 
```python
u_{8}_Engine/EnginePerformance/ThrustPerformance
```
And the second solution will have another: 
```python
u_{8}_Engine.1/EnginePerformance.1/ThrustPerformance.1
```
so that the intersection set of varkeys (```sol1vars.intersection(sol2vars)```) is empty. 
This seems problematic, because users may want to diff when a constraint is changed, which requires rebuilding the model.

bqpd:
Closes #1320 

@1ozturkbe, with this PR you can now compare solutions created in the same session by resetting modelnumbers in between with `gpkit.reset_modelnumbers()`

### DIFF
--- a/gpkit/__init__.py
+++ b/gpkit/__init__.py
@@ -7,7 +7,7 @@ from .globals import settings
 from .globals import SignomialsEnabled, SIGNOMIALS_ENABLED
 from .globals import Vectorize, VECTORIZATION
 from .globals import (NamedVariables, MODELS, MODELNUMS, MODELNUM_LOOKUP,
-                      NAMEDVARS)
+                      NAMEDVARS, reset_modelnumbers)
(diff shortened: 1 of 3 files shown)


bqpd:
test models please

1ozturkbe:
neat, although it still looks like a hack to me. Can you remind me why we have different numbers in the first place?

bqpd:
let's chat about it tomorrow!

1ozturkbe:
great!

PR #1335, opened 2018-06-06
Making relaxations work with GPs and SPs

1ozturkbe:
@bqpd this should be complete. Feel free to review.

### DIFF
--- a/gpkit/constraints/relax.py
+++ b/gpkit/constraints/relax.py
@@ -3,7 +3,8 @@ from .set import ConstraintSet
 from ..nomials import Variable, VectorVariable, parse_subs, NomialArray
 from ..keydict import KeyDict
 from .. import NamedVariables, MODELNUM_LOOKUP
-
+from ..small_classes import Count
+from gpkit import SignomialsEnabled
 
 class ConstraintsRelaxedEqually(ConstraintSet):
     """Relax constraints the same amount, as in Eqn. 10 of [Boyd2007].
@@ -30,23 +31,24 @@ class ConstraintsRelaxedEqually(ConstraintSet):
         if not isinstance(constraints, ConstraintSet):
             constraints = ConstraintSet(constraints)
         substitutions = dict(constraints.substitutions)
-        relconstrs = []
+        relcons = []
         with NamedVariables("Relax"):
             self.relaxvar = Variable("C")
-        for constr in constraints.flat(constraintsets=False):
-            if constr.oper == ">=":
-                relconstrs.append(self.relaxvar*constr.left >= constr.right)
-            elif constr.oper == "<=":
-                relconstrs.append(constr.left <= self.relaxvar*constr.right)
-            elif constr.oper == "=":
-                relconstrs.append(constr.left <= self.relaxvar*constr.right)
-                relconstrs.append(self.relaxvar*constr.left >= constr.right)
-            else:
-                raise ValueError("Constraint had unknown operator %s."
-                                 " Cannot relax the constraint %s"
-                                 % constr.oper, constr)
+        with SignomialsEnabled():
+            for constr in constraints.flat(constraintsets=False):
+                if constr.oper == ">=":
+                    relcons.append(self.relaxvar*constr.left >= constr.right)
+                elif constr.oper == "<=":
+                    relcons.append(constr.left <= self.relaxvar*constr.right)
+                elif constr.oper == "=":
+                    relcons.append(constr.left <= self.relaxvar*constr.right)
+                    relcons.append(self.relaxvar*constr.left >= constr.right)
+                else:
+                    raise ValueError("Constraint had unknown operator %s."
+                                     " Cannot relax the constraint %s"
+                                     % constr.oper, constr)
 
-        ConstraintSet.__init__(self, [relconstrs,
+        ConstraintSet.__init__(self, [relcons,
                                       self.relaxvar >= 1], substitutions)
 
 class ConstraintsRelaxed(ConstraintSet):
@@ -56,7 +58,6 @@ class ConstraintsRelaxed(ConstraintSet):
     ---------
     constraints : iterable
         Constraints which will be relaxed (made easier).
-
 
     Attributes
     ----------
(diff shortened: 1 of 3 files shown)


1ozturkbe:
@bqpd did you see this?

bqpd:
yup, sorry, fell behind on email. taking a look now!

bqpd:
test models please

bqpd:
the failures are because of changed example output, as shown in https://acdl.mit.edu/csi/job/gpkit_PullRequest_Unit/2557/buildnode=macys_VM,optimizer=cvxopt/console

Unfortunately since we only test these outputs with cvxopt, you need to run `./runtests.sh` (or the equivalent) to get the right printouts, and then commit the changes to the .txt files

gpkit/constraints/relax.py
```diff
@@ -64,14 +76,31 @@ def __init__(self, constraints):
...
+                                      constr.right)
+                elif constr.oper == "<=":
+                    relconstrs.append(constr.left <=
+                                      self.relaxvars[i.next()]*constr.right)
+                elif constr.oper == "=":
+                    relconstrs.append(constr.left <=
+                                      self.relaxvars[i.next()]*constr.right)
+                    relconstrs.append(self.relaxvars[i.count]*constr.left >=
+                                      constr.right)
+                else:
```
bqpd: there's a lot of duplication of code between this and the above; consider making a new function that handles the constraint addition?

and instead of using Counter, I'd do `for i, constr in enumerate(constraints.flat(constraintsets=False)):`

gpkit/tests/t_constraints.py
```diff
@@ -8,7 +8,8 @@
 from gpkit.constraints.loose import Loose
 from gpkit.tests.helpers import run_tests
 from gpkit.exceptions import InvalidGPConstraint
-from gpkit.constraints.relax import ConstraintsRelaxed
+from gpkit.constraints.relax import ConstraintsRelaxed, \
```
bqpd: instead of using slashes just add another line of import:

```python
from gpkit.constraints.relax import ConstraintsRelaxed
from gpkit.constraints.relax import ConstraintsRelaxedEqually, ConstantsRelaxed
```

gpkit/constraints/relax.py
```diff
@@ -30,13 +31,25 @@ def __init__(self, constraints):
...
-        posynomials = constraints.as_posyslt1()
+        relcons = []
         with NamedVariables("Relax"):
             self.relaxvar = Variable("C")
-        ConstraintSet.__init__(self,
-                               [[p <= self.relaxvar for p in posynomials],
-                                self.relaxvar >= 1], substitutions)
-
+        with SignomialsEnabled():
+            for constr in constraints.flat(constraintsets=False):
```
bqpd: consider using the more readable loop variable `constraint` instead of `constr`

gpkit/constraints/relax.py
```diff
@@ -30,13 +31,25 @@ def __init__(self, constraints):
         if not isinstance(constraints, ConstraintSet):
             constraints = ConstraintSet(constraints)
         substitutions = dict(constraints.substitutions)
-        posynomials = constraints.as_posyslt1()
+        relcons = []
```
bqpd: consider giving this a more meaningful name, like `relaxed_constraints`

gpkit/constraints/relax.py
```diff
@@ -64,14 +76,31 @@ def __init__(self, constraints):
         if not isinstance(constraints, ConstraintSet):
             constraints = ConstraintSet(constraints)
         substitutions = dict(constraints.substitutions)
-        posynomials = constraints.as_posyslt1()
-        N = len(posynomials)
+        relconstrs = []
+        N = len(constraints)
```
bqpd: because N is only used once it probably isn't worth declaring as a variable...just stick it in the VectorVar call

gpkit/tests/t_constraints.py
```diff
@@ -59,6 +60,19 @@ def test_equality_relaxation(self):
...
+        y = Variable('y')
+        z = Variable('z')
+        with SignomialsEnabled():
+            m = Model(x, [x+y >= z], {z:2})
+        r1 = ConstantsRelaxed(m)
+        r2 = ConstraintsRelaxed(m)
+        r3 = ConstraintsRelaxedEqually(m)
+        self.assertEqual(len(r1.varkeys), 5)
+        self.assertEqual(len(r2.varkeys), 4)
+        self.assertEqual(len(r3.varkeys), 4)
```
bqpd: maybe maybe add another constraint to the model so that r2 and r3 have a different number of varkeys?

gpkit/constraints/relax.py
1ozturkbe: It was for pylint... I didn't want all of the lines to be broken.

1ozturkbe:
@bqpd I cannot for the life of me figure out why the other computers have a different simpleflight example test output than my computer. Can you see if you can replicate it on your machine?

bqpd:
test models please

bqpd:
@1ozturkbe I had slightly different output from my computer...let's see if that corresponds with the server

docs/source/examples/debug_output.txt
```diff
@@ -18,8 +18,8 @@ and these constants relaxed:
 > Trying with relaxed constraints:
 
 Solves with these constraints relaxed:
-   1: 3500% relaxed, from 3*x**-1 <= 1
-                       to 3*x**-1 <= 36
+   1: 3500% relaxed, from 1 [yd] <= 1
```
bqpd: it looks like the relaxation has messed up this output...the old one is the correct thing to do here, I think!

docs/source/examples/relaxation_output.txt
```diff
@@ -13,8 +13,8 @@ With constraints relaxed equally
   # minimize
         C_Relax
   # subject to
-        C_Relax >= x*x_max**-1
-        C_Relax >= x**-1*x_min
+        x <= C_Relax*x_max
```
bqpd: The old order seems more readable to me, but this is minor.

docs/source/examples/relaxation_output.txt
```diff
@@ -45,7 +45,8 @@ With constraints relaxed individually
   # minimize
         C_Relax.1_(0,)*C_Relax.1_(1,)*x**0.01
   # subject to
-        C_Relax.1 >= [gpkit.Monomial(x*x_max**-1), gpkit.Monomial(x**-1*x_min)]
+        x <= C_Relax.1_(0,)*x_max
```
bqpd: this is probably better than the old output, but worth looking at the difference

gpkit/nomials/math.py
```diff
@@ -388,6 +393,20 @@ def subinplace(self, substitutions):
         self.varkeys = KeySet(self.left.vks)
         self.varkeys.update(self.right.vks)
 
+    def relaxed(self, relaxvar):
```
bqpd: oh, interesting! why is this a method of a Posynomial, however? I think this should be a function that takes a posynomial as an argument, and it should be in `relax.py`

1ozturkbe: I have thought about this. Let's discuss.

bqpd:
test models please

gpkit/nomials/math.py
1ozturkbe: I think we agreed this is good!

1ozturkbe:
darnit, tests are failing because of variable ordering, which differs between machines. Why does that occur? Seems like a persistent problem.

bqpd:
It's persistent between your computer and jenkins, not between my computer and jenkins, so we should look into the configuration on your computer...first, are you solving with cvxopt? second, we can increase the weight we give to alphabetization in the sorting (by changing the coefficient below)

https://github.com/convexengineering/gpkit/blob/a8999737980ba45682e6e00770cf4546ca5337af/gpkit/solution_array.py#L494

1ozturkbe:
I run tests through the shell script, which definitely uses cvxopt. I think the alphabetization weight might work!

1ozturkbe:
@bqpd I'm done with trying to mess with solution_array. I don't see any other way to fix this than using OrderedDicts for the solutions, testing something other than complete outputs, or doing my outputs on a different machine. Seeing that this is a persistent issue I recommend the second. What do you think?

1ozturkbe:
@bqpd as per our discussion yesterday, it would be great if you could update the outputs and merge this. Thanks!

bqpd:
will do once I have a laptop again!

bqpd:
test models please

1ozturkbe:
yeyeyey finally.

bqpd:
note that I extended the tests to verify solve results as well...

PR #1344, opened 2018-08-01
added copy model classes

alisaab:
I added two new classes under gpkit.Model: the first copies a model with  its original substitutions only while flattening constraints,  while the second copies a models with all its substitutions while preserving hierarchy.

### DIFF
--- a/gpkit/constraints/model.py
+++ b/gpkit/constraints/model.py
@@ -1,7 +1,7 @@
 "Implements Model"
 import numpy as np
 from .costed import CostedConstraintSet
-from ..nomials import Monomial
+from ..nomials import Monomial, MonomialEquality, PosynomialInequality
 from .prog_factories import _progify_fctry, _solve_fctry
 from .gp import GeometricProgram
 from .sgp import SequentialGeometricProgram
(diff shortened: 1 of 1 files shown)


bqpd:
add to whitelist

bqpd:
test this please

bqpd:
test models please

1ozturkbe:
Looks OK, but needs to also work for signomials. Let me look into this.

alisaab:
EqualModel works for signomials. I was thinking that we might only need EqualModel, maybe with a different name. SameModel might not be of much importance. what do you think ? @1ozturkbe

1ozturkbe:
Sorry this slipped under my radar. If we are only adding EqualModel, it's probably better to not go ahead with this pull request, and users can figure it out on their own that you can have more of the same model by calling Model(m.cost, m.substitutions) etc.

1ozturkbe:
@alisaab thoughts?

alisaab:
yes sure, this can be replaced by Model(m.cost, m.flat(constraintsets=False), m.substitutions). No need for the EqualModel if this is straight forward for the users. @1ozturkbe

1ozturkbe:
Aight, closing then!

PR #1346, opened 2018-08-01
 fix and unify almost_equal and diff for 0 substitutions 

### LINKED ISSUE #1345: diff() and almost_equal() give ZeroDivisionError
When testing the usage of diff() and almost_equal() to compare different solutions, I find that these methods give division by zero errors when used to compare a solution to itself, e.g:

```
sol.almost_equal(sol)
sol.diff(sol)
```

I expect this is because the solutions are exactly equal and thus you cannot calculate a relative error.

However, I think the expected behavior should rather be statements saying that the solutions are exactly equal within some tolerance.

bqpd:
Closes #1345

### DIFF
--- a/gpkit/solution_array.py
+++ b/gpkit/solution_array.py
@@ -64,7 +64,7 @@ def insenss_table(data, _, maxval=0.1, **kwargs):
 TABLEFNS = {"sensitivities": senss_table,
             "topsensitivities": topsenss_table,
             "insensitivities": insenss_table,
-            }
+           }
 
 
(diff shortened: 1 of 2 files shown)


bqpd:
test models please

bqpd:
@1ozturkbe ready for review!

michK:
I found another case that breaks `almost_equal` but works for `diff`.

Emailed source to Ned.

Run `mission_all_electric.py` in arch_test_cases to reproduce error.

Regards

PR #1361, opened 2018-10-31
add sorted_diff option

bqpd:


bqpd:
todo: option to not do model printing

PR #1364, opened 2018-11-07
Drafting new test and fixes for tautological constraint.

1ozturkbe:
Trying to modify the tautological constraint error for signomials, but having some minor issues.


bqpd:
This looks great! Folded this into https://github.com/convexengineering/gpkit/pull/1367 with a few small changes

PR #1374, opened 2019-01-08
add basic mosek9 support

bqpd:
todo: 
  - [ ] integer programming
  - [ ] faster problem generation
    - [ ] faster regeneration on sweeps
    - [ ] warmstarts on sweeps and SPs


rileyjmurray:
@1ozturkbe this should probably be closed.

bqpd:
good call!

Issue #1386, opened 2019-02-12
Discussing post-processing of Tight for complex ArrayConstraints


1ozturkbe:
This is mostly to have a debate on the right process for using information about untight constraints. Right now, we post a warning on untight Tight constraints of ```printwarning=True```, but this is somewhat useless if the user needs to be able to use that information for some kind of post-processing (eg. me, for solid rocket motor shape design) or to get some intuition. And also, text is *meh*. Being able to make some visual interface would be nice. 

One proposal was to simply add kwargs to each Tight constraint, but that is super messy, especially if you have an ArrayConstraint. I will be debating this for the next while, so let me know if you have ideas @bqpd and others!

bqpd:
@1ozturkbe, check out the new warnings and constraints-printing in https://github.com/convexengineering/gpkit/pull/1388. (try running a model with Tight constraints then doing `print sol.table(tables=["warnings", "tightest constraints", "loose constraints"])`) and let me know what you think! The warnings table will print by default now, so it's less important to print warnings during a solve.

bqpd:
For your purposes the new format of `sol["warnings"]["Unexpectedly Loose Constraints"]`, where each entry is a tuple of `(warning_message, flagged_constraint)` should help with parsing according to your flags above.

bqpd:
I think this is closed by the above functionality; @1ozturkbe, can you confirm?

1ozturkbe:
That’s right, thanks for following up on this @bqpd.

PR #1387, opened 2019-03-04
Adding kwargs into Tight

1ozturkbe:
Hey @bqpd, this is partly about issue #1386. I decided that the use cases of Tight are so unique that users will need to figure out their own ways of dealing with it. My way for the solid rocket motor branch (soromotoropt) is to have kwargs to label Tight constraints. This PR just adds the kwargs so I don't have to work out of a branch other than master.


1ozturkbe:
If you are good with this, merge, and then close issue #1386.

1ozturkbe:
@bqpd this is a two minute review. Would really appreciate it!

bqpd:
Sure thing! Labels specifically sound like a great idea, though; would you be interested in making a version that only adds the labeling functionality?

1ozturkbe:
I debated doing that, but I think it is better to use kwargs for multiple indexing. There may be reasons why you might want multiple attributes on a Tight constraint (eg. you want to validate the constraint for certain kinds of studies).

bqpd:
sounds good to me, I'll merge back at my comp

PR #1393, opened 2019-04-16
properly parse showvars in diff and savecsv

bqpd:
Uses the `SolutionArray.parse_showvars` method to parse showvars, instead of a couple of ad-hoc implementations.


whoburg:
nice.

PR #1402, opened 2019-06-12
number paths, not models

### LINKED ISSUE #1371: only number models if the whole path is not unique

bqpd:
Closes #1371


bqpd:
test models please

gpkit/globals.py
```diff
@@ -103,16 +103,17 @@ def __exit__(self, type_, val, traceback):
 
 def reset_modelnumbers():
     "Zeroes all model number counters"
```
whoburg: Update the docstring please?

bqpd: Good call.

PR #1416, opened 2019-08-16
PenaltyCCP algorithm.

1ozturkbe:
This implements an updated signomial algorithm which currently does not work terribly well. But always nice to have the option. Let me know if you have opinions about the implementation @bqpd.

### DIFF
--- a/gpkit/constraints/model.py
+++ b/gpkit/constraints/model.py
@@ -82,6 +82,8 @@ class Model(CostedConstraintSet):
     solve = _solve_fctry(_progify_fctry(GeometricProgram, "solve"))
     localsolve = _solve_fctry(_progify_fctry(SequentialGeometricProgram,
                                              "localsolve"))
+    penalty_ccp_solve = _solve_fctry(_progify_fctry(SequentialGeometricProgram,
+                                                    "penalty_ccp_solve"))
 
     def verify_docstring(self):  # pylint:disable=too-many-locals,too-many-branches,too-many-statements
         "Verifies docstring bounds are sufficient but not excessive."
--- a/gpkit/constraints/sgp.py
+++ b/gpkit/constraints/sgp.py
@@ -5,9 +5,11 @@ from collections import OrderedDict
 import numpy as np
 from ..exceptions import InvalidGPConstraint
 from ..keydict import KeyDict
-from ..nomials import Variable
+from ..nomials import Variable, VectorVariable
 from .gp import GeometricProgram
 from ..nomials import SignomialInequality, PosynomialInequality
+from ..nomials import SingleSignomialEquality
+from .. import SignomialsEnabled
 from .costed import CostedConstraintSet
 from ..small_scripts import mag
 
@@ -75,7 +77,7 @@ class SequentialGeometricProgram(CostedConstraintSet):
 
     # pylint: disable=too-many-locals
     def localsolve(self, solver=None, verbosity=1, x0=None, reltol=1e-4,
-                   iteration_limit=50, mutategp=True, **kwargs):
+                   iteration_limit=50, mutategp=True, relax = False, **kwargs):
         """Locally solves a SequentialGeometricProgram and returns the solution.
 
         Arguments
@@ -95,6 +97,12 @@ class SequentialGeometricProgram(CostedConstraintSet):
             consecutive solve's objective values.
         iteration_limit : int
             Maximum GP iterations allowed.
+        mutategp: boolean
+            Prescribes whether to mutate the previously generated GP
+            or to create a new GP with every solve.
+        relax: boolean
+            Implements the penalty convex-concave algorithm [Lipp,Boyd 2016]
+            instead of vanilla SP heuristic.
         *args, **kwargs :
             Passed to solver function.
 
@@ -110,6 +118,29 @@ class SequentialGeometricProgram(CostedConstraintSet):
         self.gps = []  # NOTE: SIDE EFFECTS
         self.solver_outs = []
         self._results = []
+        self.relax = False
+        if relax:
+            self.relax = True
+            print("Using penalty CCP...")
+            self.signomials = []
+            self.gp_constrs = []
+            for constr in self.flat(constraintsets=False):
+                if isinstance(constr, SingleSignomialEquality) or \
+                        isinstance(constr, SignomialInequality):
+                    self.signomials.append(constr)
+                else:
+                    self.gp_constrs.append(constr)
+            slack = VectorVariable(len(self.signomials))
+            with SignomialsEnabled():
+                self.relaxed_signomials = [constr.relaxed(slack[i])
+                                  for i,constr in enumerate(self.signomials)]
+            relaxed_obj = self.cost*np.prod(slack)**20.
+            self.relaxed_model = SequentialGeometricProgram(relaxed_obj,
+                        [self.gp_constrs, self.relaxed_signomials], self.substitutions)
+            self.result = self.relaxed_model.localsolve(solver, verbosity, x0, reltol,
+                   iteration_limit, mutategp, relax=False, **kwargs)
+            self.gps = self.relaxed_model.gps
+            return self.result
         # if there's external functions we can't mutate the GP
         mutategp = mutategp and not self.externalfns
         if x0 and not mutategp:
@@ -219,7 +250,7 @@ class SequentialGeometricProgram(CostedConstraintSet):
         self._numgpconstrs = len(gp.hmaps) - len(spapproxs)
         return gp
 
-    def gp(self, x0=None, mutategp=False):
+    def gp(self, x0=None, mutategp=False, **kwargs):
         "The GP approximation of this SP at x0."
         if mutategp:
             if not self.gps:
@@ -237,7 +268,7 @@ class SequentialGeometricProgram(CostedConstraintSet):
             gp.gen()
         else:
             x0 = self._fill_x0(x0)
-            gp_constrs = self.as_gpconstr(x0)
+            gp_constrs = self.as_gpconstr(x0, kwargs)
             if self.externalfn_vars:
                 for v in self.externalfn_vars:
                     posyconstr = v.key.externalfn(v, x0)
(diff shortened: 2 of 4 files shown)


1ozturkbe:
ugh, as useful as pylint is, it sometimes rubs me the wrong way. eg. it says
```python
            relaxed_model = SequentialGeometricProgram(relaxed_obj,
                            [gp_constrs, relaxed_signomials], self.substitutions)
```
has wrong indentation because it wants
```python
            relaxed_model = SequentialGeometricProgram(relaxed_obj,
                                                                                        [gp_constrs, relaxed_signomials], self.substitutions)
```
which is also invalid due to line being too long.

whoburg:
It should also allow either

```python
relaxed_model = SequentialGeometricProgram(
      relaxed_obj, [gp_constrs, relaxed_signomials], self.substitutions)
```

or

```python
relaxed_model = SequentialGeometricProgram(
      relaxed_obj,
      [gp_constrs, relaxed_signomials],
      self.substitutions)
```

gpkit/constraints/sgp.py
```diff
@@ -110,6 +118,28 @@ def localsolve(self, solver=None, verbosity=1, x0=None, reltol=1e-4,
         self.gps = []  # NOTE: SIDE EFFECTS
         self.solver_outs = []
         self._results = []
+        if relax:
+            self.relax = True
+            print("Using penalty CCP...")
```
whoburg: check verbosity level here?

gpkit/constraints/sgp.py
```diff
@@ -110,6 +118,28 @@ def localsolve(self, solver=None, verbosity=1, x0=None, reltol=1e-4,
...
+            signomials = []
+            gp_constrs = []
+            for constr in self.flat(constraintsets=False):
+                if isinstance(constr, (SingleSignomialEquality, \
```
whoburg: Don't think you need these backslashes. Unless I'm missing something?

gpkit/constraints/sgp.py
```diff
@@ -110,6 +118,28 @@ def localsolve(self, solver=None, verbosity=1, x0=None, reltol=1e-4,
...
+                                       SignomialInequality)):
+                    signomials.append(constr)
+                else:
+                    gp_constrs.append(constr)
+            slack = VectorVariable(len(signomials))
+            with SignomialsEnabled():
+                relaxed_signomials = [constr.relaxed(slack[i]) \
```
whoburg: same comment re backslash

whoburg:
Super glad to see some new exploration of SP algorithms in GPkit!

1ozturkbe:
@bqpd this should now be working. Merge when ready.

1ozturkbe:
test models please

1ozturkbe:
argh, no idea why it is taking the windows machine 30 mins to run tests... it simply times out, and thus the testing issues. @bqpd what does one do?

gpkit/constraints/sgp.py
```diff
@@ -95,10 +100,15 @@ def localsolve(self, solver=None, verbosity=1, x0=None, reltol=1e-4,
             consecutive solve's objective values.
         iteration_limit : int
             Maximum GP iterations allowed.
+        mutategp: boolean
+            Prescribes whether to mutate the previously generated GP
+            or to create a new GP with every solve.
+        relax: boolean
+            Implements the penalty convex-concave algorithm [Lipp,Boyd 2016]
+            instead of vanilla SP heuristic.
```
whoburg: Thanks for adding documentation!

gpkit/constraints/sgp.py
```diff
@@ -110,6 +120,15 @@ def localsolve(self, solver=None, verbosity=1, x0=None, reltol=1e-4,
         self.gps = []  # NOTE: SIDE EFFECTS
         self.solver_outs = []
         self._results = []
+        if relax:
+            if verbosity > 0:
+                print("Using penalty CCP...")
+            relaxed_model = self.penalty_ccp()
```
whoburg: It seems a bit weird to me for the user-facing interface to be setting the `relax` kwarg of local solve. As we increase the number of choices for SP algorithms, we should decide whether they should all be accessed as different user-facing methods (`sp.penalty_ccp_solve(reltol=...)`), or whether they should be accessed via args and kwargs of `localsolve`. I would vote for the former. It allows cleaner algorithm-specific parameters.

Philosophically, I think it's nice for GPkit to expose users to the names (from the literature) of algorithms where possible. I think that builds trust (that we implemented things the same way as the literature) and encourages new algorithm developers to use GPkit to sandbox their algorithm (they get 'credit' because users actually interact with the name of their algorithm.

gpkit/constraints/sgp.py
```diff
@@ -246,3 +265,24 @@ def gp(self, x0=None, mutategp=False):
...
+        "Returns the penalty convex-concave form of this SP."
+        signomials = []
+        gp_constrs = []
+        for constr in self.flat(constraintsets=False):
+            if isinstance(constr, (SingleSignomialEquality,
+                                   SignomialInequality)):
+                signomials.append(constr)
+            else:
+                gp_constrs.append(constr)
+        slack = VectorVariable(len(signomials))
```
whoburg: Should we give it a name, so that it makes more sense in user-facing output and for debugging?

gpkit/tests/t_constraints.py
```diff
@@ -358,9 +358,23 @@ def test_substitution_issue905(self):
...
+        x = Variable("x")
+        y = Variable("y", 2.)
+        z = Variable("z")
+        with SignomialsEnabled():
+            m = Model(1/z, [z <= x**2 + y, x*z == 2])
+        sol = m.localsolve(verbosity=0)
+        m.program.relax = True
+        sol_pccp = m.localsolve(verbosity=0, relax=True)
+        self.assertEqual(len(m.program.gps[-1].varkeys), 3)
+        self.assertAlmostEqual(sol['cost'], sol_pccp['cost'])
```
whoburg: 🙌  hooray for unit tests

whoburg: This makes it seem like we do not have tests for vanilla `localsolve()`, but those are implemented in `t_model.py`. We should decide where these tests belong and put them all in the same place? The tests in `t_model` also give access to `MULTI_SOLVER_TESTS` to run for all solvers, which is probably appropriate for testing `penalty_ccp`.

gpkit/tests/t_constraints.py
```diff
@@ -358,9 +358,23 @@ def test_substitution_issue905(self):
         sol = bm.solve(verbosity=0)
         self.assertAlmostEqual(sol["cost"], 1.0)
 
+class TestLocalsolve(unittest.TestCase):
+    """Test different SP solution methods"""
+
+    def test_penaltyccp(self):
```
whoburg: Add underscore to match method name?

```suggestion
    def test_penalty_ccp(self):
```

gpkit/constraints/sgp.py
1ozturkbe: Makes sense. Any thoughts @bqpd?

1ozturkbe: Nvm, I agree with you and I'm just going to pull the trigger.

gpkit/constraints/model.py
```diff
@@ -82,6 +82,8 @@ def __init__(self, cost=None, constraints=None, *args, **kwargs):
     solve = _solve_fctry(_progify_fctry(GeometricProgram, "solve"))
     localsolve = _solve_fctry(_progify_fctry(SequentialGeometricProgram,
                                              "localsolve"))
+    penalty_ccp_solve = _solve_fctry(_progify_fctry(SequentialGeometricProgram,
+                                                    "penalty_ccp_solve"))
```
bqpd: (whistles appreciatively)

gpkit/constraints/sgp.py
```diff
@@ -110,6 +117,7 @@ def localsolve(self, solver=None, verbosity=1, x0=None, reltol=1e-4,
         self.gps = []  # NOTE: SIDE EFFECTS
         self.solver_outs = []
         self._results = []
+        self.solve_method = 'localsolve'
```
bqpd: is this used anywhere?

gpkit/constraints/sgp.py
```diff
@@ -171,6 +179,26 @@ def localsolve(self, solver=None, verbosity=1, x0=None, reltol=1e-4,
...
+        """ Implements the penalty convex-concave algorithm [Lipp,Boyd 2016]
+            instead of vanilla SP heuristic.
+
+            Same arguments as localsolve, but also
+                exp : float (optional)
+            Sets penalty for violated signomial constraints
+        """
+        self.solve_method = 'penalty_ccp_solve'
+        if verbosity > 0:
+            print("Using penalty CCP...")
```
bqpd: Needs clarification: perhaps "Applying penalty CCP transformation before solving"?

gpkit/constraints/sgp.py
```diff
@@ -246,3 +274,24 @@ def gp(self, x0=None, mutategp=False):
...
+        "Returns the penalty convex-concave form of this SP."
+        signomials = []
+        gp_constrs = []
+        for constr in self.flat(constraintsets=False):
+            if isinstance(constr, (SingleSignomialEquality,
+                                   SignomialInequality)):
+                signomials.append(constr)
+            else:
+                gp_constrs.append(constr)
+        slack = VectorVariable(len(signomials), 's')
```
bqpd: You should do this in a `with NamedVariables(foo):` environment to ensure the variable name is unique. Maybe `foo = CCP`?

gpkit/constraints/sgp.py
1ozturkbe: Removing it. It wasn't being used.

gpkit/constraints/model.py
1ozturkbe: Still don't fully understand what this does, all I know is that it allows me to call it on a Model as well as a SGP :p

1ozturkbe:
Great folks, merging this.

PR #1421, opened 2019-08-23
add unichr spackle for python3

### LINKED ISSUE #1410: unichr not defined
Hi,

I tried running a sankey diagram, and it failed because unichr was not defined. 

Could someone help update the sankey diagram code?

Thanks!
Dev

Here is a reference to where the issue is.

`~/anaconda3/lib/python3.6/site-packages/gpkit/interactive/sankey.py in diagram(self, variables, flowright, width, height, top, bottom, left, right, printing)
    134                 variables = [variables]
    135             for var in variables:
--> 136                 self.varlinks(self.gp[0], var.key, printing=printing)
    137             # if var_eqs were found, label them on the diagram
    138             lookup = {key: i for i, key in

~/anaconda3/lib/python3.6/site-packages/gpkit/interactive/sankey.py in varlinks(self, constrset, key, target, printing)
     97                 if constr not in self.constr_name:
     98                     # use unicode's circled letters for constraint labels
---> 99                     source = unichr(self.counter.next()+9398)
    100                     self.constr_name[constr] = source
    101                 else:

NameError: name 'unichr' is not defined
`

bqpd:
Closes #1410

### DIFF
--- a/gpkit/interactive/sankey.py
+++ b/gpkit/interactive/sankey.py
@@ -1,5 +1,6 @@
 "implements Sankey"
 from __future__ import unicode_literals, print_function
+import sys
 from collections import defaultdict
 import numpy as np
 from ipywidgets import Layout
@@ -12,6 +13,9 @@ from .. import GPCOLORS
 
 
 INSENSITIVE = 1e-2
+
+if sys.version_info >= (3, 0):
+    unichr = chr  # pylint: disable=redefined-builtin,invalid-name
 
 
 def getcolor(value):


gpkit/interactive/sankey.py
```diff
@@ -13,6 +14,9 @@
 
 INSENSITIVE = 1e-2
 
+if sys.version_info >= (3, 0):
+    unichr = chr  # pylint: disable=redefined-builtin,invalid-name
```
whoburg: wondering if it would be better to define `version_insensitive_unichr` (holy moly hopefully someone comes up with a better name), but anyway use that everywhere in this file instead of redefining the builtin?

Issue #1417, opened 2019-08-20
str_without returning Varkey


1ozturkbe:
I am running tests in robust that I am trying to update to python3, and realized that there are cases when calling Varkey.str_without() returns Varkey! This breaks robust tests on line 42 of gpkit.varkey.py:
```python
        fullstr = self.str_without(["modelnums", "vec"])
        self.eqstr = fullstr + str(self.lineage) + self.unitrepr
```
when fullstr is actually not a string. Trying to figure out what is going on rn, lmk if you have an idea @bqpd.

whoburg:
Would be great to have a minimum working example that can be turned into a test. Ideally without dependencies.

1ozturkbe:
Alright... I think I know why this happens, now that I have spent most of today dissecting it...
This occurs when I am creating a Monomial in robust.
```python
        first_monomial = Monomial(self.p.exps, self.p.cs)
```
where I am passing a dict (self.p.exps) of {Varkey: float} describing the exponents, and a float coefficient. 
Since Monomial inherits from Signomial, the error occurs here in Signomial:
```python
            elif isinstance(hmap, dict):
                exp = HashVector({VarKey(k): v
                                  for k, v in hmap.items() if v})
                hmap = NomialMap({exp: mag(cs)})
                hmap.units_of_product(cs)
```
It understands that the map of exponents is a dict, then GPkit proceeds to try to create a HashVector with VarKey(k) as keys, where k is a VarKey! I am super confused why GPkit does this. Can you explain @bqpd?

1ozturkbe:
So from the look of the implementation in GPkit tests, it seems like the dict of exps passed onto Monomial has keys that are varkey strings, not VarKey objects. Will set up a fix.

whoburg:
@1ozturkbe I suggest creating a separate issue for the Monomial `__init__` discussion... in part because I think it will help us to refine our unit tests for that interface. Would be great to have MWE's for both that issue and this issue.

1ozturkbe:
This is referenced in pull request #1419.

bqpd:
You can't create monomials like that anymore...since like a year ago? pass it `dict(zip(exps, cs))`

bqpd:
oh hmm nvm

bqpd:
So. I don't understand this method of creating a copy of `self.p`: why not just use `self.p` itself? But if you really need to, don't recreate the nomial map from scratch: pass it `Monomial(self.p.hmap)`

[...]

PR #1422, opened 2019-08-25
Signomial.chop to obtain Monomials. 

1ozturkbe:
This is a standard way to chop posynomials and signomials (with SignomialsEnabled()) into their Monomials. Imagine this is super useful for a number of codebases (eg. robust).


1ozturkbe:
And its associated test.

1ozturkbe:
Also for documentation, this is in response to #1417, it being a little painful to have to deal with hmaps to subdivide posynomials.

PR #1438, opened 2019-09-28
Allowing substituted, unitless signomials in exponents. 

1ozturkbe:
This will take a bunch of work, but I wanted this to be a clear WIP subject to discussion.

### DIFF
--- a/gpkit/small_classes.py
+++ b/gpkit/small_classes.py
@@ -192,9 +192,8 @@ class HashVector(dict):
 
     def __pow__(self, other):
         "Accepts scalars. Return Hashvector with each value put to a power."
-        if isinstance(other, Numbers):
-            return self.__class__({key: val**other
-                                   for (key, val) in self.items()})
+        return self.__class__({key: val**other
+                               for (key, val) in self.items()})
         return NotImplemented
 
     def __mul__(self, other):
@@ -202,7 +201,8 @@ class HashVector(dict):
 
         If the other object inherits from dict, multiplication is element-wise
         and their key's intersection will form the new keys."""
-        if isinstance(other, Numbers):
+        from gpkit.nomials.math import Signomial
+        if isinstance(other, Numbers + (Signomial,)):
             return self.__class__({key: val*other
                                    for (key, val) in self.items()})
         elif isinstance(other, dict):
--- a/gpkit/constraints/sp.py
+++ b/gpkit/constraints/sp.py
@@ -1,0 +1,127 @@
+"""Implement the SignomialProgram class"""
+from __future__ import unicode_literals, print_function
+from time import time
+from collections import OrderedDict
+import numpy as np
+from ..exceptions import InvalidGPConstraint
+from ..keydict import KeyDict
+from ..nomials import Variable, VectorVariable
+from .gp import GeometricProgram
+from ..nomials import SignomialInequality, PosynomialInequality
+from ..nomials import SingleSignomialEquality
+from .. import SignomialsEnabled, NamedVariables
+from .costed import CostedConstraintSet
+from ..small_scripts import mag
+
+
+# pylint: disable=too-many-instance-attributes
+class SignomialProgram(CostedConstraintSet):
+    """Prepares a collection of signomials for a SP solve.
+
+    Arguments
+    ---------
+    cost : Posynomial
+        Objective to minimize when solving
+    constraints : list of Constraint or SignomialConstraint objects
+    verbosity : int (optional)
+        Currently has no effect: SignomialPrograms don't know
+        anything new after being created, unlike GeometricPrograms.
+
+    Attributes with side effects
+    ----------------------------
+    TODO: DETERMINE WHAT TO SET AFTER MEETING WITH RILEY
+     `result` is set at the end of a solve
+
+    """
+    def __init__(self, cost, constraints, substitutions):
+        # pylint:disable=super-init-not-called
+        self.gps = []
+        self.solver_outs = []
+        self._results = []
+        self.result = None
+        self._spconstrs = []
+        self._spvars = set()
+        self._approx_lt = []
+        self._numgpconstrs = None
+        self._gp = None
+
+        if cost.any_nonpositive_cs:
+            raise TypeError("""SPs still need Posynomial objectives.
+
+    The equivalent of a Signomial objective can be constructed by constraining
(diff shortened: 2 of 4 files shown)


1ozturkbe:
Already encountering issues in Nomials.str_without, since it expects floats as exponents. Will implement fix.

1ozturkbe:
Alright, as I dig I realize there are more and more things to do. I will be tracking signomials with variable exponents with a self.varexps attribute, and making sure we track these variables shortly.

whoburg:
Big picture strategy discussion. These types of expressions are not GP compatible, but are compatible with the exponential cone (possibly requiring some substitutions to get them into an exponential cone compatible form). So this will play nicely with mosek 9. Do we all agree with that overall picture?

Do we want something along the lines of `SignomialsEnabled` to switch between GP mode and exponential cone mode? I actually think that might be too overbearing (with mosek 9 support, we're solving using exponential cone instead of GP anyway, so maybe we should let users create this type of expression by default). With that said, I still think we should also provide some transparency to the user about when their program is a pure GP vs an exponential cone program. It's one of the virtues of GPkit that we expose users to a bit of the mathematics underlying their modeling decisions.

Not trying to side track! Just thought it would be a useful high-level discussion.

1ozturkbe:
The primary intended use for this is to allow for obtaining the sensitivities to the exponents. I'm not even thinking about exponential cones atm. But good to think about for the future. I do think that we should strive for mosek 9 compatibility before we consider allowing exponents to be variables anyways.

gpkit/nomials/map.py
```diff
@@ -20,8 +20,9 @@ class NomialMap(HashVector):
     x and y are VarKey objects.
     """
     units = None
-    expmap = None  # used for monomial-mapping postsubstitution; see .mmap()
-    csmap = None   # used for monomial-mapping postsubstitution; see .mmap()
+    expmap = None   # used for monomial-mapping postsubstitution; see .mmap()
+    csmap = None    # used for monomial-mapping postsubstitution; see .mmap()
+    varexps = False # used for signomial exponent operations
```
bqpd: convention for comments in GPkit is double space between code and `#`, no alignment

bqpd: I think pylint may complain about one of those or the other?

gpkit/small_classes.py
```diff
@@ -192,17 +192,17 @@ def __neg__(self):
 
     def __pow__(self, other):
         "Accepts scalars. Return Hashvector with each value put to a power."
-        if isinstance(other, Numbers):
-            return self.__class__({key: val**other
-                                   for (key, val) in self.items()})
+        return self.__class__({key: val**other
+                               for (key, val) in self.items()})
         return NotImplemented
```
bqpd: this fallback return statement is currently unnecessary

1ozturkbe:
Wow, this has been an absolute battle against existing GPkit norms. I can hardly get anything to work, since it involves really messing with NomialData and Nomial. I have figured out that I will likely need to store the signomial exponents using an hmap embedded in an hmap.

whoburg:
> The primary intended use for this is to allow for obtaining the sensitivities to the exponents. I'm not even thinking about exponential cones atm. But good to think about for the future. I do think that we should strive for mosek 9 compatibility before we consider allowing exponents to be variables anyways.

Per the issue title, the exponents would be unitless signomials, but it sounds like the intent is for them just to be constant 'variables' such that sensitivities can be extracted. I think this should be made more clear -- is this PR for the general signomial case, or just for fixed variables?

whoburg:
> Wow, this has been an absolute battle against existing GPkit norms. I can hardly get anything to work, since it involves really messing with NomialData and Nomial. I have figured out that I will likely need to store the signomial exponents using an hmap embedded in an hmap.

I wonder if we need to think about how `NomialData` and `Nomial` get extended beyond GP to the more general exponential cone.

1ozturkbe:
More difficulties... It turns out that differentiating a variable**signomial is non-trivial. Will try and make it neat.

1ozturkbe:
@bqpd I wanted to ask you a question about differentiation in GPkit. Does it exist anywhere in the core code other than tests? (The derivative of var**sig w.r.t a variable in the signomial has a log, and was wondering how much havoc this is going to cause.) I have already scoured the code but couldn't find anything. Let me know.

bqpd:
hmap.diff is used in monomial_lower_bound creation and in the calculation
of sensitivities for Signomials turned into Posynomials by substitution

1ozturkbe:
Oh poop of course. Alright, I will try to make a non-intrusive method.

bqpd:
todo for Ned:
  - [ ] check whether some old speedups (vks / varkeys for example) can be simplified

1ozturkbe:
@bqpd I see that this only breaks test_boundschecking, but what does that f'n actually do? It just seems to have a ```pass```?

bqpd:
@1ozturkbe ah, that means it just checks that `boundschecking.py` in the examples directory runs without raising errors, but doesn't perform any additional tests

bqpd:
It looks like the error might just be in the type of error raised by cvxopt, which the test expects to be a ValueError but which is now a RuntimeWarning. I'm a little concerned that the error type has changed; this indicates that the bounds-checking in gp.py is no longer raising the ValueError that it should for an unbounded model like this one, but is instead attempting to solve it. Did you disable boundschecking errors?

1ozturkbe:
umm not that I know of! it is possible I created a bug since I have been messing with bounds (since signomial exponents, unsubstituted, cannot bound anything).

bqpd:
@1ozturkbe I'm afraid all the recent changes on master have created some conflicts here...but on the plus side, the removal of 1700 lines of code from GPkit will probably make it easier to implement :p

1ozturkbe:
It's alright. This was quite the mongo chunk of code to do at once... at least I get less code to work with, you are right!

1ozturkbe:
@bqpd I'm never going to get to finish this before I graduate. If you have a desire to continue this (for the sake of getting sensitivities to exponents), feel free to reuse the code. Otherwise, this can be closed for the near future.

bqpd:
@1ozturkbe no worries, makes sense! & let me know if I can do anything to help!

bqpd:
I'll leave it open just for visibility's sake...

PR #1439, opened 2019-10-01
fix pint's unicode printing of ohms

bqpd:


bqpd:
test this please

bqpd:
right? ugh it was preventing savetxt from working.

PR #1440, opened 2019-10-01
Update build status icons. 

1ozturkbe:


1ozturkbe:
Hmmm, also pip install seems to be failing, likely because of the changes you made in Jenkins @bqpd.

PR #1443, opened 2019-10-08
Various fixes

bqpd:


gpkit/nomials/math.py
```diff
@@ -519,6 +523,8 @@ def sens_from_dual(self, la, nu, result):  # pylint: disable=unused-argument
         self.relax_sensitivity = la
         if self.sgp_parent:
             self.sgp_parent.relax_sensitivity = la
+            if getattr(self.sgp_parent, "sgp_parent", None):
+                self.sgp_parent.sgp_parent.relax_sensitivity = la
```
1ozturkbe: Why are there two layers of sgp_parent here?

gpkit/solution_array.py
```diff
@@ -70,9 +70,14 @@ def tight_table(self, _, ntightconstrs=5, tight_senss=1e-2, **kwargs):
...
     title = "Tightest Constraints"
-    data = [((-float("%+6.2g" % c.relax_sensitivity), str(c)),
-             "%+6.2g" % c.relax_sensitivity, id(c), c)
-            for c in self.model.flat() if c.relax_sensitivity >= tight_senss]
+    try:
+        data = [((-float("%+6.2g" % c.relax_sensitivity), str(c)),
+                 "%+6.2g" % c.relax_sensitivity, id(c), c)
+                for c in self.model.flat()
+                if c.relax_sensitivity >= tight_senss]
+    except AttributeError:
```
1ozturkbe: Agree that this is a better error message!

gpkit/nomials/math.py
1ozturkbe: Oh let me guess, because you can have relaxations and SP approximations?

bqpd: yup.

bqpd:
test models please

PR #1445, opened 2019-10-17
try additional file removal strategy

bqpd:


galbramc:
test this please

galbramc:
I still see temporary files lingering. I saw a number of temporary directories created and deleted, but not all got deleted. The directories that remain have gpkit_mosek and gpkit_mosek.sol files in them.

galbramc:
Maybe you need some while loop to check if the file still lingers.

bqpd:
huh, they should all be getting deleted: either by os.remove or by
shutil.rmtree! Even weirder that some are and some are not deleted.

One (kinda silly) alternative could be to reuse the same temporary filename
for each session of gpkit, which should work but also creates an entirely
unnecessary potential race condition.

bqpd:
There is a while loop in the rmtree function! If all remaining directories
also have files leftover, that implies os.remove isn't doing anything, so
this PR isn' necessary.

galbramc:
Welcome to the world of windoze. I'm guessing Windoze is locking the file claiming it is still in use when you are trying to remove it. Humor me and put a while exists in front of each remove command.

bqpd:
done! added while loop

galbramc:
test this please

galbramc:
You missed the '.path.'. Should be:
```
os.path.isfile(filename)
```

bqpd:
derp

galbramc:
Nope... try this:

```
try:
    shutil.rmtree(path)
except OSError as e:
    # Reraise unless ENOENT: No such file or directory
    # (ok if directory has already been deleted)
    if e.errno != errno.ENOENT:
        raise
```

bqpd:
Heh, wrapped it around my existing recursion; how's that?

>

galbramc:
Geez! I don't know why some still linger... For all these incarnations there are fewer left behind, there are still ~5-10 that don't get deleted.

galbramc:
Oh wait. Wrong PR building.

galbramc:
Try removing the extra arguments to shutil.rmtree.

bqpd:
done! :| thanks so much for like...watching the temporary files directory

galbramc:
I don't understand this... There are still files lingering, but it is better than what it was without any of these changes. Maybe the while should be on the directory instead of the file inside?

bqpd:
heh. giving that a try...

PR #1446, opened 2019-11-25
sort variables for ~determinism~

bqpd:


bqpd:
test models please

1ozturkbe:
Also, yeah, why varlocs and _varlocs? I understand _varlocs is private, but what's the difference? Otherwise, I'm happy with this once tests pass.

bqpd:
This hasn't come up yet (bc constraints are deterministically ordered), and will be moot once we move to python 3.6

PR #1488, opened 2020-04-23
Fixing .debug issues with VectorVars. 

1ozturkbe:
Issues have to do with ```constants``` having different varkeys than ```substitutions```. Now, just create a new subs dict to avoid issues using del.

### DIFF
--- a/gpkit/constraints/relax.py
+++ b/gpkit/constraints/relax.py
@@ -116,8 +116,8 @@ class ConstantsRelaxed(ConstraintSet):
 
         if not isinstance(constraints, ConstraintSet):
             constraints = ConstraintSet(constraints)
-        substitutions = KeyDict(constraints.substitutions)
-        constants, _, linked = parse_subs(constraints.varkeys, substitutions)
+        old_substitutions = KeyDict(constraints.substitutions)
+        constants, _, linked = parse_subs(constraints.varkeys, old_substitutions)
         if linked:
             kdc = KeyDict(constants)
             constrained_varkeys = constraints.constrained_varkeys()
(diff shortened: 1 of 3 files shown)


bqpd:
i think the del is likely to be intentional, will take a look at this

1ozturkbe:
Not going to lie, VarKeys are really confusing.  Consider the following.

I am taking a bunch of solutions of models that I saved, and I create some bounds on a set of variables in a dict(VarKey: [min, max]) form, just like substitutions but each element has a list. Then, I take an identical GP model and try to get the variables using Model[varkey] so I can make constraints over them, and obviously it doesn't find them because the varkeys are different. I get that each VarKey is unique, but then what is the best way to be able to transfer data between identical (or similar) instances of models? 

I looked at SGP.gp since it gets passed an x0, and thought it would answer this question. But instead, I see that it calls x0 = KeyDict(x0). Frankly, KeyDict is a behemoth and I don't really get what it's doing, If I apply KeyDict to a substitution dict it seems to do nothing? I could use some clarification. (These core parts of GPkit are clearly not meant to be looked at, but to do any development over GPkit requires a little more help than is provided by the documentation).

1ozturkbe:
@bqpd I guess if you will want to avoid making copies of substitutions I can restore the del statement and go back to updating the subs? any updates?

1ozturkbe:
It all has to do with the behavior of use of del. Using del substitutions[key] where key pertains to a single element of a VectorVariable, it replaces that element with nan, but doesn't allow the retrieval of other elements.

bqpd:
closed by #1496

PR #1498, opened 2020-07-02
compressed solpickles

bqpd:


bqpd:
test models please

bqpd:
@pgkirsch newest commit halves compressed pickled sol size again while keeping saving time about the same as before and speeding up loading to about the same as an uncompressed pickle

bqpd:
if that carries over to your 110MB file you should expect ~15MBs taking ~20s to compress and ~10 to decompress

bqpd:
test models

bqpd:
test models please

Issue #1497, opened 2020-07-01
Pickled solutions are very large


pgkirsch:
`SolutionArray.save()` is a wonderful feature but it can produce some seriously big files. I recently solved a 42000 free variable model and the pickle file was 110 MB! Even the models I solve on a more frequent basis yield 10-20 MB pickle files. In case anyone is curious the simple text file representation of the solution (output of `SolutionArray.savetxt()` is ~700 kB.

I previously spoke to @bqpd about this and he mentioned that it should be possible to make these files smaller. If it is a trade-off between how much original model data/functionality is preserved and file size, it would be great if there was an argument to specify tiers of data preservation.

bqpd:
without looking into changing just what is pickled, using cpickle and bz2 takes it from 12 to 3.6 MB:

```python
class SolutionArray(...):

    def save_compressed(self, title="solution", **cpickleargs):
        "Pickle a file and then compress it into a file with extension."
        with bz2.BZ2File(title + ".pbz2", "w") as f:
            cPickle.dump(self, f, **cpickleargs)

    @staticmethod
    def decompress_file(file):
        "Load any compressed pickle file"
        return cPickle.load(bz2.BZ2File(file, "rb"))
```

bqpd:
@pgkirsch check out https://github.com/convexengineering/gpkit/pull/1498; what's that bring the 110MB file down to?

There's obviously a lot more that can be done by changing what is pickled as well...

pgkirsch:
27 MB!

pgkirsch:
Definitely a performance trade off, as I'm sure you're aware. In case you're curious: compressing takes ~20 seconds and decompressing takes ~32 seconds (coincidentally the original solution time takes 52 seconds!). With regular pickle it takes 7 seconds to save and 12 seconds to load.

What kinds of things could be selectively left out of the pickle?

bqpd:
I think there's something about the way it's including all the constraints that's taking more space than it should. I haven't found any debugging tools which show what's taking up space in a pickle file, so it'll be trial and error figuring out what that might be...

bqpd:
(note: down to 1.9MB, much faster loads in lastest commit on https://github.com/convexengineering/gpkit/pull/1498)

bqpd:
[...]

PR #1503, opened 2020-07-14
add SolSavingEnvironment, halving pickle size

bqpd:
Per #1497

@pgkirsch this halves pickle size without removing any solution information. Between this and the compression I'd expect that 110MB file to drop to less than 10MB.

### DIFF
--- a/gpkit/nomials/math.py
+++ b/gpkit/nomials/math.py
@@ -368,7 +368,6 @@ MONS = Numbers + (Monomial,)
 
 class ScalarSingleEquationConstraint(SingleEquationConstraint):
     "A SingleEquationConstraint with scalar left and right sides."
-    nomials = []
     generated_by = v_ss = parent = None
     bounded = meq_bounded = {}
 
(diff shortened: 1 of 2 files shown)


bqpd:
test models please

pgkirsch:
@bqpd just curious: what do the "construction/solve attributes" allow you to do with an already-computed solution? Re-solve without needing the original model object?

bqpd:
The following attributes are removed: `["mfm", "pmap", "bounded", "meq_bounded", "v_ss", "unsubbed", "varkeys"]`.

`mfm`: maps how monomials simplified during substitution so that sensitivities can be properly mapped back
`pmap`: maps how the constant term (if any) was taken to the `<= 1` side and divided out
`bounded`: which variables were upper or lower bounded by this constraint at its creation
`meq_bounded`: which variables were conditionally upper or lower bounded through a monomial equality
`v_ss`: the individual variable sensitivities in each constraint
`unsubbed`: the presubstitution polynomial representations of the constraint
`varkeys`: the varkeys in the constraint

All of these except for `v_ss` are recreatable by initing a new constraint with the left, oper, and right of the original.

pgkirsch:
Ah awesome, thanks for the detailed explanation!

PR #1509, opened 2020-07-21
Documentation bug fixes and Sphinx refactor

bqpd:
@mtwichan this is a version of https://github.com/convexengineering/gpkit/pull/1508 hosted on this repository so that we can all make commits! Let me know if you'd rather work on that PR, you can just cherry-pick the lint commit I made on this one.

To run pylint with our settings, you can just run `./pylint.sh` after installing pylint version 2.4.4 in your virtual environment. (`pip install pylint==2.4.4`)


mtwichan:
@bqpd awesome thanks! I'll work on this PR instead.

mtwichan:
Hey, @bqpd I ran all the example code naively using a simple bash script:

```
for file in *.py; do (
    echo 
    echo "$file"
    echo "------------------------------"
    python3 $file
)
done
```
and I fixed up all the errors I encountered.

I also added tests for all the new example code I added and filled in a couple of additional tests. I'm still learning about writing proper tests, so before I filled in the rest of them I wanted to get your input:

`t_examples.py`: https://github.com/convexengineering/gpkit/blob/303-add-literals/gpkit/tests/t_examples.py
Diff: https://github.com/convexengineering/gpkit/pull/1509/commits/50f44749c0c4f1d3ac50f05b031806469395e05d

Also, should I remove the `assert` statements from the user-facing documentation code, or is that intentional?

It looks like all my new tests passed successfully, so I'm not quite sure why the `unit test` failed ... hmm:
Screenshot from 2020-07-24 14-50-10

bqpd:
Fabulous, and now it's passing all the tests!

The user-facing asserts are intentional but that doesn't mean they're a good idea...

bqpd:
added this branch to the docs. it'll build here: https://gpkit.readthedocs.io/en/303-add-literals/

so far, looks all good

bqpd:
I think the code on https://gpkit.readthedocs.io/en/303-add-literals/gettingstarted.html should be lines excerpted from a larger file, rather than a bunch of small files which can run on their own

bqpd:
the Vectorize example should probably be "from gpkit import [whatever's used]" instead of namespacing gpkit

bqpd:
the Advanced Commands section really benefits from this change

bqpd:
the SP example should also not be namespaced

bqpd:
otherwise the docs look good to me!

mtwichan:
> I think the code on https://gpkit.readthedocs.io/en/303-add-literals/gettingstarted.html should be lines excerpted from a larger file, rather than a bunch of small files which can run on their own

I found that it's possible to grab specific lines of code in a file using [Sphinx](https://www.sphinx-doc.org/en/1.7/markup/code.html):

Ex: 
```
.. literalinclude:: example.py
   :lines: 1,3,5-10,20-
``

What do you think this file should look like?

If I combined all the files should they share a global import? If so, then I'd have concerns about each example not having `import ...` line, because it's quite nice to be able to copy, paste and go with example code. I could also smoosh all the files into one larger file as they are, where each example would have its own set of imports, but I wasn't sure if this was something you'd be okay with since it's a bit messy.

@bqpd

bqpd:
@mtwichan I think the [gettingstarted](https://gpkit.readthedocs.io/en/latest/gettingstarted.html) page should be fine with just the top-level imports: it's not really for copy-pasting, but for verbose explanation of the parts of a file.

bqpd:
@mtwichan are you planning any more big changes to this, or should I do a last review and then merge?

mtwichan:
Hey @bqpd, just pushed up my final changes. 

As a consequence of condensing the `gettingstarted.rst` documentation code into a single file, there are no tests for `gettingstarted.py` unless I change all the variable names. I was thinking about it, but the code seems pretty trivial so I don't think it's worth it.

bqpd:
Fabulous! Will review and merge.

PR #1514, opened 2020-08-06
remove logged_gets, allow substitution of a linked function at an index

bqpd:


bqpd:
test models please

bqpd:
test models please

PR #1527, opened 2020-11-17
Working MIGP

### LINKED ISSUE #414: Mixed-integer geometric programming
To support #413, how hard would it be to add a way of specifying integer variables?

For now, I don't think we want to get be in the business of solving MIGP problems, so the needed functionality is just using GPkit to get from a symbolic description to dumping out the solver format (c, A, map, and some way of encoding which variables are integer, probably borrowing the formatting conventions from AMPL or one of the MILP solvers).

bqpd:
Closes #414

 potential extensions:
  - [ ] allow for choosing between sets of substitutions
    - [ ] allow "freeing" variables to be between eps and 1/eps if they're only in some of the sets of substitutions
  - [ ] allow for choosing between sets of ConstraintSets

@rileyjmurray @erling-d-andersen, this implements a binary variable technique for discretizing logspace variabes, but it's a bit...slow. Any recommendations? Is this just because solving integer problems has a constant-speed slowdown? (if there's only one "choice" there's no slowdown, but I presume that's the presolver optimizing it away...)


1ozturkbe:
I believe you are adding a new binary var for each possible integer solution to do this, so the slowdown is is somewhat expected @bqpd. There are two possible formulations here. 

Rn you do this. 
new binary variable z[i=1:number of options]
sum(z[i]) = 1
choice = sum(choices * z)

The following is likely faster, since it introduces hierarchy into binary vars. (Just so you know, I rushed this so something might be slightly off about the formulation.)
new binary variable z[i=1:number of options]
z[j] >= z[j+1] for all j=1:number_of_options-1
choice = choices[0] + sum(choice_differences[1:j] * z)
where the choice differences are the logspace differences between choices[0] and choices[1] etc.
I hope that helps.

bqpd:
oh indeed, that's a much better way to do it!

bqpd:
unfortunately that didn't make it any faster, but it does reduce the number of binary variables by 1...

rileyjmurray:
@bqpd different mixed-integer solvers will use different branching strategies depending on the integer formulation. So even if MOSEK isn't running much faster, it's possible that other solvers will (should any be added in the future). I specifically suspect that simpler (open-source) mixed-integer solvers will do better with the new formulation.

1ozturkbe:
@bqpd three questions. 
- What is the size of your integer domain? Don't want to crush your hopes, but I don't think MIGP will be tractable in the scale that MILP is, at least with current solvers. I think most MICP algorithms are lightweight and don't do anything much smarter than B&B under the hood (eg. cutting planes etc.). And a naive approach to B&B will mostly enumerate...
- Is your problem a GP? Because if it is, since the objective space is "smooth", you'd be much better off solving the relaxed problem, then only allowing integer solutions "in the vicinity" of the relaxation and resolving. If the user wants a higher quality solution, you could run the more exhaustive MIGP approach, but I assume the first will work optimally in most cases. This is what a lot of MICP solvers do. 
- Have you tried passing an initial guess to the second formulation? I know this is not ideal, but good guesses dramatically speed up B&B, much more so than IP solves.

bqpd:
1. ha it's very small right now, just a few options! haven't even tried multiple discrete variables yet.

2. I'm letting MOSEK do their conic integer alg rather than implementing my own branch & bound, so I presume they're handling that (and given the results upthread they're doing something a little more complex than what I would have implemented on my own!)

3. I haven't! I've been thinking what an interface for that might be; my feeling rn is that because they're discrete sets rather than integers this is much less of a problem, but I could either allow for a guess or automatically choose the median?

bqpd:
test models please

bqpd:
test models please

bqpd:
test models please

bqpd:
retest this please

bqpd:
test models please

PR #1531, opened 2020-12-01
Partial vector substitution

bqpd:
allows substitutions like ```m.substitutions[v[0,2:]] = 1  # or np.ones(v[0,2:].shape)```

### DIFF
--- a/gpkit/keydict.py
+++ b/gpkit/keydict.py
@@ -4,6 +4,7 @@ from collections.abc import Hashable
 import numpy as np
 from .small_classes import Numbers, Quantity, FixedScalar
 from .small_scripts import is_sweepvar, isnan
+from .varkey import GLOBAL_KEYMAP
 
 DIMLESS_QUANTITY = Quantity(1, "dimensionless")
 INT_DTYPE = np.dtype(int)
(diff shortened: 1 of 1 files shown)


bqpd:
test models please

bqpd:
@pgkirsch what do you think of this feature's utility? It seems to me like something that should Just Work, so I'm willing to take the performance hit.

pgkirsch:
@bqpd that seems helpful! This seems like it would further simplify the proposed aero changes?

I assume that syntax carries over to the setup method form?
```python
def setup(self):
...
    substitutions = {
        v[0,2:]: 1,
    ...
    }

    return constraints, substitutions
```

I know I'm sort of pushing the envelope on some of these vector variable subs, but do you have other users with anything similar in use case?

bqpd:
yup, carries over!

I think this is one of those things where people might have tried to use it, it didn't work, they found an alternative without mentioning...

pgkirsch:
Awesome!

Have you profiled what the performance hit is for a large model? I can't imagine it's too significant, right?

bqpd:
it was ~4% of the time in that function...don't remember exactly the timing
in my test run, but no more than a couple hundredths of a second.

>

pgkirsch:
Seems very tolerable!

bqpd:
I'm just so used to making things quicker when I make changes to core classes :p

pgkirsch:
Ha sorry!

bqpd:
test models please

PR #1532, opened 2020-12-01
Reduce memory usage and makes saving constraints optional

bqpd:
- uses sets of varkeys instead of KeySets in many places
    - deletes KeySets after use for string-substitution during ConstraintSet creation
 - deletes substitution attributes after used for mmaping / dual calculation

Reduces memory consumption for large models with string substitutions during creation by ~30%

For future reference: consider not using KeySets for such one-offs at all, instead doing a one-pass over varkeys. Honestly such iteration is cheap enough that KeySets might be removed entirely, and KeyDicts used only for solutions. This wouldn't lead to large changes in speed/memory, but it would remove a lot of code.

Also makes saving constraints with a solution optional

### DIFF
--- a/gpkit/constraints/gp.py
+++ b/gpkit/constraints/gp.py
@@ -59,11 +59,11 @@ class GeometricProgram:
 
     Examples
     --------
-    >>> gp = gpkit.geometric_program.GeometricProgram(
+    >>> gp = gpkit.constraints.gp.GeometricProgram(
                         # minimize
                         x,
                         [   # subject to
-                            1/x  # <= 1, implicitly
+                            x >= 1,
                         ], {})
     >>> gp.solve()
     """
@@ -78,8 +78,8 @@ class GeometricProgram:
                     sub = sub.to(key.units or "dimensionless").magnitude
                 self.substitutions[key] = sub
             if not isinstance(sub, (Numbers, np.ndarray)):
-                raise ValueError("substitution {%s: %s} has invalid value type"
-                                 " %s." % (key, sub, type(sub)))
+                raise TypeError("substitution {%s: %s} has invalid value type"
+                                " %s." % (key, sub, type(sub)))
         cost_hmap = cost.hmap.sub(self.substitutions, cost.varkeys)
         if any(c <= 0 for c in cost_hmap.values()):
             raise InvalidPosynomial("a GP's cost must be Posynomial")
@@ -117,18 +117,21 @@ class GeometricProgram:
         return missingbounds
 
     def gen(self):
-        "Generates nomial and solve data (A, p_idxs) from posynomials"
-        # k [posys]: number of monomials (rows of A) present in each constraint
+        """Generates nomial and solve data (A, p_idxs) from posynomials.
+
+        k [posys]: number of monomials (rows of A) present in each constraint
+        m_idxs [mons]: monomial indices of each posynomial
+        p_idxs [mons]: posynomial index of each monomial
+        cs, exps [mons]: coefficient and exponents of each monomial
+        varlocs: {vk: monomial indices of each variables' location}
+        meq_idxs: {all indices of equality mons} and {the first index of each}
+        varidxs: {vk: which column corresponds to it in A}
+        A [mons, vks]: sparse array of each monomials' variables' exponents
+
+        """
         self.k = [len(hmap) for hmap in self.hmaps]
-        # m_idxs [mons]: monomial indices of each posynomial
-        self.m_idxs = []
-        # p_idxs [mons]: posynomial index of each monomial
-        self.p_idxs = []
-        # cs, exps [mons]: coefficient and exponents of each monomial
-        self.cs, self.exps = [], []
-        # varlocs: {vk: monomial indices of each variables' location}
+        self.m_idxs, self.p_idxs, self.cs, self.exps = [], [], [], []
         self.varkeys = self.varlocs = defaultdict(list)
-        # meq_idxs: {all indices of equality mons} and {just the first halves}
         self.meq_idxs = MonoEqualityIndexes()
         m_idx = 0
         row, col, data = [], [], []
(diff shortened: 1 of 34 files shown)


bqpd:
test models please

bqpd:
test models please

bqpd:
test models please

bqpd:
test models please

bqpd:
test this please

bqpd:
test models please

gpkit/constraints/set.py
```diff
@@ -64,58 +64,60 @@ class ConstraintSet(list, ReprMixin):
     "Recursive container for ConstraintSets and Inequalities"
     unique_varkeys, idxlookup = frozenset(), {}
     _name_collision_varkeys = None
+    _varkeys = None
 
-    def __init__(self, constraints, substitutions=None):  # pylint: disable=too-many-branches,too-many-statements
+    def __init__(self, constraints, substitutions=None, *, bonusvks=None):  # pylint: disable=too-many-branches,too-many-statements
```
whoburg, line 69: docstring for `bonusvks`? Or at the very least an inline comment?

gpkit/constraints/gp.py
```diff
@@ -305,33 +305,33 @@ def _compile_result(self, solver_out):
...
-        self.v_ss = cost_senss.copy()
+        gpv_ss = cost_senss.copy()
         for las, nus, c in zip(la[1:], self.nu_by_posy[1:], self.hmaps[1:]):
             while getattr(c, "parent", None) is not None:
                 c = c.parent
             v_ss, c_senss = c.sens_from_dual(las, nus, result)
             for vk, x in v_ss.items():
-                self.v_ss[vk] = x + self.v_ss.get(vk, 0)
-            while getattr(c, "generated_by", None) is not None:
+                gpv_ss[vk] = x + gpv_ss.get(vk, 0)
```
whoburg, line 314: haven't thought deeply about this, but seems like this is acting a lot like a `defaultdict`? Could `gpv_ss` be a default dict, and then this simply becomes `gpv_ss[vk] += x`? Is `defaultdict` noticeably slower?

gpkit/constraints/gp.py
```diff
@@ -305,33 +305,33 @@ def _compile_result(self, solver_out):
...
+            while getattr(c, "generated_by", None):
                 c = c.generated_by
```
whoburg, line 316: inline comment here? Something like "loop to top of hierarchy"?

gpkit/constraints/gp.py
```diff
@@ -305,33 +305,33 @@ def _compile_result(self, solver_out):
...
+            dlogcost_dlogv = gpv_ss.pop(v)
             val = np.array(result["constants"][v])
             for c, dv_dc in v.gradients.items():
                 with warnings.catch_warnings():  # skip pesky divide-by-zeros
                     warnings.simplefilter("ignore")
                     dlogv_dlogc = dv_dc * result["constants"][c]/val
-                    before = self.v_ss.get(c, 0)
-                    self.v_ss[c] = before + dlogcost_dlogv*dlogv_dlogc
+                    before = gpv_ss.get(c, 0)
+                    gpv_ss[c] = before + dlogcost_dlogv*dlogv_dlogc
```
whoburg, line 327: same comment as above, seems like `defaultdict` might simplify this?

gpkit/keydict.py
```diff
@@ -45,7 +45,7 @@ class KeyMap:
     collapse_arrays = False
     keymap = []
     log_gets = False
-    varkeys = None
+    cset = None
```
whoburg, line 48: documentation / inline comment for `cset`? 😬

gpkit/constraints/set.py
```diff
@@ -64,58 +64,60 @@ class ConstraintSet(list, ReprMixin):
...
-                    del key.descr["value"]
-                    if key.veckey and key.veckey.value is not None:
-                        del key.veckey.descr["value"]
+            self._varkeys = None
+        for key in self.vks:
+            if key not in self.substitutions:
+                if key.veckey is None or key.veckey not in self.substitutions:
+                    continue
+                if np.isnan(self.substitutions[key.veckey][key.idx]):
+                    continue
```
whoburg, line 104: these `continue` cases seem complex enough that an inline comment would be beneficial

gpkit/constraints/set.py
```diff
@@ -64,58 +64,60 @@ class ConstraintSet(list, ReprMixin):
...
+            self.bounded.add((key, "upper"))
+            self.bounded.add((key, "lower"))
+            if key.value is not None and not key.constant:
+                del key.descr["value"]
+                if key.veckey and key.veckey.value is not None:
+                    del key.veckey.descr["value"]
```
whoburg, line 110: ?? I've read this a few times and still confused what's going on here... inline comment? (sorry broken record)

gpkit/nomials/data.py
```diff
@@ -18,9 +18,6 @@ class NomialData(ReprMixin):
 
     def __init__(self, hmap):
         self.hmap = hmap
-        self.vks = set()
-        for exp in self.hmap:
-            self.vks.update(exp)
```
whoburg, line 23: 🎉

gpkit/nomials/map.py
```diff
@@ -152,18 +152,16 @@ def mmap(self, orig):
...
         selfexps = list(self.keys())
         for orig_exp, self_exp in self.expmap.items():
             if self_exp not in self:  # can occur in tautological constraints
                 continue              #     after substitution
             fraction = self.csmap.get(orig_exp, orig[orig_exp])/self[self_exp]
-            m_from_ms[self_exp][orig_exp] = fraction
             orig_idx = origexps.index(orig_exp)
             pmap[selfexps.index(self_exp)][orig_idx] = fraction
-        return pmap, m_from_ms
+        return pmap
```
whoburg, line 164: 🎉

whoburg (approved):

bqpd:
test models please

PR #1533, opened 2020-12-02
don't wrap arrays for singleton string keys

### LINKED ISSUE #1529: Should accidental array substitutions fail gracefully or does savetxt need a fix?
Recent use-case that caused a minor headache that could maybe have a GPkit-side fix:
1. Solve a model with some Vectorized variables
2. Fix the value of a key variable to the optimal solution multiplied by a factor (`m.substitutions.update({"velocity": 1.25*sol("velocity").magnitude})`. In this instance the variable in question has a Vectorization of 1 so its optimal value is a single-value array.
3. Re-solve
4. Save new solution to text file using `savetxt()`

Everything works fine until step 4, when `savetxt` returns:
```python
TypeError: ufunc 'isnan' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```
Probing the initial solution reveals:
```python
In [7]: sol("velocity")                                                                                                            
Out[7]: array([array([124.61993074])], dtype=object) <Unit('meter / second')>
```

Once I figured out the issue, the fix was easy (use `.item()`) but it feels like GPkit should either:
1. Flag the attempted substitution as invalid, or
2. Quietly "fix" the substitution, or
3. Be able to handle this kind of substitution in `savetxt`

MWE:
```python
from gpkit import Variable, Model, Vectorize                                                            
                                                                                 
class System(Model):                                                             
    def setup(self):                                                             
        with Vectorize(1):                                                       
            self.Fleet = Fleet()                                                 
        constraints = [self.Fleet]                                               
        self.cost = sum(self.Fleet.z)                                            
        return constraints                                                       
                                                                                 
class Fleet(Model):                                                              
    def setup(self):                                                                                                                   
        with Vectorize(1):                                                       
            self.Vehicle = Vehicle()                                             
                                                                                 
        self.z = z = Variable("z")                                               
        constraints = [                                                          
            z >= sum(self.Vehicle.y**2),                                         
            self.Vehicle,                                                        
        ]                                                                        
        return constraints                                                       
                                                                                 
class Vehicle(Model):                                                            
    def setup(self):                                                             
        self.y = y = Variable("y")                                               
        constraints = [y >= 1]                                                   
        return constraints                                                       
                                                                                 
m = System()                                                                     
sol = m.solve()                                                                  
sol.savetxt("test.txt")                                                          
m.substitutions.update({"y": sol("y")})                 # :(
#m.substitutions.update({"y": sol("y").item()})         # :)
sol2 = m.solve()                                                                 
sol2.savetxt("test2.txt")                                                        
~

bqpd:
Closes #1529

### DIFF
--- a/gpkit/keydict.py
+++ b/gpkit/keydict.py
@@ -196,6 +196,8 @@ class KeyDict(KeyMap, dict):
             self._unmapped_keys.add(key)
             if idx:
                 dty = {} if isinstance(value, Numbers) else {"dtype": "object"}
+                if getattr(value, "shape", None) and value.dtype != INT_DTYPE:
+                    dty["dtype"] = value.dtype
                 dict.__setitem__(self, key, np.full(key.shape, np.nan, **dty))
                 self.owned.add(key)
         if idx:


bqpd:
test models please

PR #1541, opened 2021-03-19
Clean user-submitted x0s during localsolve

### LINKED ISSUE #1520: .localsolve ignoring x0 and use_pccp
I have a fairly complex model, with a few SP constraints, that is sensitive to the initial guess, i.e., won't converge (or won't converge without slack) unless I give it a good x0. When I set x0= a pickle of a previous solution, it solves fine in 2 or 3 GPs. However, if I set x0= a dictionary with the same information as in the pickle, the solver appears to ignore x0 and has the same convergence history as if no x0 was given.

If I go back to 1f5f8bf, the behavior is as expected, with both the pickle-load and user-defined dictionary yielding the same convergence history. Beyond b6fd017, only the pickle-loaded x0 works, and .localsolve also appears to ignore use_pccp=False.

bqpd:
Closes #1520


bqpd:
test models please

bqpd:
test this please

PR #1554, opened 2021-08-04
jsonsave function for sol

Alexander-Da-Great:
This branch implements json saving for solutions. This is just the implementation for solutions - excluding the diffs discussion for now. Json's produced via this method have a depth of two. It might be useful to have a hierarchical structure in the json sol that reflects the structure of the model a solution is formed from.

### DIFF
--- a/gpkit/solution_array.py
+++ b/gpkit/solution_array.py
@@ -603,9 +603,9 @@ class SolutionArray(DictOfLists):
         for i, (k, v) in enumerate(data.items()):
             key = str(k)
             if isinstance(v, np.ndarray):
-                val = {'v': v.tolist(), 'u': str(k.descr["units"])}
+                val = {'v': v.tolist(), 'u': str(k.unitstr())}
             else:
-                val = {'v': v, 'u': str(k.descr["units"])}
+                val = {'v': v, 'u': str(k.unitstr())}
             sol_dict[key] = val
         for key in self.name_collision_varkeys():
             del key.descr["necessarylineage"]
--- a/gpkit/solution_array.py
+++ b/gpkit/solution_array.py
@@ -593,22 +593,22 @@ class SolutionArray(DictOfLists):
                 f.write(self.modelstr + "\n")
             f.write(self.table(**kwargs))
 
-    def savejson(self, filename="solution.json", *, showvars=None):
+    def savejson(self, filename="solution.json"):
         "Saves solution table as a json file"
         sol_dict = {}
-        # get list of variables 
+        # get list of variables
         data = self["variables"]
         # add appropriate data for each variable to the dictionary
         for i, (k, v) in enumerate(data.items()):
             key = str(k.name)
             # check if ndarray, jsonify using tolist()
-            if isinstance(v,np.ndarray):
+            if isinstance(v, np.ndarray):
                 val = {'v': v.tolist(), 'u': str(k.descr["units"])}
-            else: 
-                val = {'v': v, 'u': str(k.descr["units"])} 
+            else:
+                val = {'v': v, 'u': str(k.descr["units"])}
             sol_dict[key] = val
         with open(filename, "w") as f:
-            json.dump(sol_dict, f) 
+            json.dump(sol_dict, f)
 
     def savecsv(self, filename="solution.csv", *, valcols=5, showvars=None):
         "Saves primal solution as a CSV sorted by modelname, like the tables."
--- a/gpkit/solution_array.py
+++ b/gpkit/solution_array.py
@@ -63,7 +63,9 @@ def msenss_table(data, _, **kwargs):
     if "models" not in data.get("sensitivities", {}):
         return ""
     data = sorted(data["sensitivities"]["models"].items(),
-                  key=lambda i: (-round(np.mean(i[1]), 1), i[0]))
+                  key=lambda i: ((i[1] < 0.1).all(),
+                                 -np.max(i[1]) if (i[1] < 0.1).all()
+                                 else -round(np.mean(i[1]), 1), i[0]))
     lines = ["Model Sensitivities", "-------------------"]
     if kwargs["sortmodelsbysenss"]:
         lines[0] += " (sorts models in sections below)"
@@ -77,10 +79,8 @@ def msenss_table(data, _, **kwargs):
                 msenssstr = "%6s" % ("<1e%i" % np.log10(msenss))
             else:
                 msenssstr = "  =0  "
-        elif not msenss.shape:
-            msenssstr = "%+6.1f" % msenss
         else:
-            meansenss = np.mean(msenss)
+            meansenss = round(np.mean(msenss), 1)
             msenssstr = "%+6.1f" % meansenss
             deltas = msenss - meansenss
             if np.max(np.abs(deltas)) > 0.1:
@@ -88,7 +88,7 @@ def msenss_table(data, _, **kwargs):
                              for d in deltas]
                 msenssstr += " + [ %s ]" % "  ".join(deltastrs)
         if msenssstr == previousmsenssstr:
-            msenssstr = "      "
+            msenssstr = " "*len(msenssstr)
         else:
             previousmsenssstr = msenssstr
         lines.append("%s : %s" % (msenssstr, model))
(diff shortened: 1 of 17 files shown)


bqpd:
whitelist this

bqpd:
test this please

gpkit/solution_array.py
```diff
@@ -592,6 +593,23 @@ def savetxt(self, filename="solution.txt", *, printmodel=True, **kwargs):
...
             f.write(self.table(**kwargs))
 
+    def savejson(self, filename="solution.json"):
+        "Saves solution table as a json file"
+        sol_dict = {}
+        # get list of variables
+        data = self["variables"]
+        # add appropriate data for each variable to the dictionary
+        for i, (k, v) in enumerate(data.items()):
+            key = str(k.name)
```
bqpd, line 603: This should probably be `str(k)` within a `for key in self.name_collision_varkeys():  key.descr["necessarylineage"] = True/False` environment, as in `def tables(...)` below.

bqpd (changes requested):
made one note, otherwise looks great!

gpkit/solution_array.py
```diff
@@ -592,6 +593,25 @@ def savetxt(self, filename="solution.txt", *, printmodel=True, **kwargs):
                 f.write(self.modelstr + "\n")
             f.write(self.table(**kwargs))
 
+    def savejson(self, filename="solution.json"):
```
bqpd, line 596: Looks great! There should be a test that just verifies the output of this on e.g. the Peformance Modeling example...will add it when I can or you're welcome to take a stab!

Alexander-Da-Great: Test is up under the performance modeling example in t_examples.py. Let me know if this is the kind of test you would be looking for - my stab may have been more like a gouge.

bqpd, line 596: do you want to add an argument to turn only certain variables into json?

gpkit/solution_array.py
```diff
@@ -592,6 +593,25 @@ def savetxt(self, filename="solution.txt", *, printmodel=True, **kwargs):
...
+        "Saves solution table as a json file"
+        sol_dict = {}
+        # add appropriate data for each variable to the dictionary
+        for key in self.name_collision_varkeys():
+            key.descr["necessarylineage"] = True
+        data = self["variables"]
+        for i, (k, v) in enumerate(data.items()):
+            key = str(k)
+            if isinstance(v, np.ndarray):
+                val = {'v': v.tolist(), 'u': str(k.unitstr())}
```
bqpd, line 606: GPkit uses `"` double quotes, sorry for the bother!

bqpd, line 606: is there a reason that just `k.unitstr()` doesn't work here?

gpkit/solution_array.py
```diff
@@ -592,6 +593,25 @@ def savetxt(self, filename="solution.txt", *, printmodel=True, **kwargs):
...
+            else:
+                val = {'v': v, 'u': str(k.unitstr())}
+            sol_dict[key] = val
+        for key in self.name_collision_varkeys():
+            del key.descr["necessarylineage"]
+        with open(filename, "w") as f:
+            json.dump(sol_dict, f)
```
bqpd, line 613: should this also return the json as a string, in case the user wants it? (i.e. `return json.dumps(sol_dict)`)

maybe if the filename is `None` it doesn't save and only does that.

gpkit/tests/t_examples.py
```diff
@@ -192,6 +193,17 @@ def test_performance_modeling(self, example):
...
+        json_dict = {}
+        with open("solution.json", "r") as rf:
+            json_dict = json.load(rf)
+        for var in sol["variables"]:
+            self.assertTrue(np.all(json_dict[str(var.key)]['v']
+                                   == sol["variables"][var.key]))
+            self.assertEqual(json_dict[str(var.key)]['u'], var.unitstr())
+
+
+
```
bqpd, line 206: not sure how these extra newlines crept in but otherwise this test looks great!

gpkit/tests/t_examples.py
```diff
@@ -177,7 +178,7 @@ def test_model_var_access(self, example):
 
     def test_performance_modeling(self, example):
         m = Model(example.M.cost, Loose(example.M), example.M.substitutions)
-
+        NamedVariables.reset_modelnumbers()
```
bqpd, line 181: why is this necessary? (may be worth a comment indicating)

bqpd (changes requested):
It looks feature-complete! Added notes on a couple speculative features and lint.

gpkit/solution_array.py
Alexander-Da-Great: This would be a great addition. I was talking to Philippe about how it would be nice to have solution jsons organized with a hierarchical structure according to sub-models. It could be nice to have the option to print only variables in a certain sub-model rather than imposing a top down structure on the json.

gpkit/tests/t_examples.py
Alexander-Da-Great: When developing I encountered an issue where mode1.var != model.var - that has since disappeared. Not sure how this got resolved, either way I will removed the   NamedVariables line.

gpkit/solution_array.py
bqpd: `showvars = self._parse_showvars(showvars)` cleans up some of the various input types, then just intersect / check for inclusion of "k" in there. `parse_showvars` should handle vectorvariables for you, but worth a check if you end up implementing this!

bqpd: looks great to me

bqpd:
@Alexander-Da-Great, I merged the recent changes to master so that the unit tests would run.

gpkit/solution_array.py
Alexander-Da-Great: created a boolean called printjson that will return the string if flagged true

gpkit/solution_array.py
Alexander-Da-Great: Would love to hear any input on what would make sense in the "printing variables of specific sub-models" department

bqpd: with this method you can thankfully just use the models' `varkeys` attribute, as in this section from the performance modeling example:

```python
vars_of_interest = set(AC.varkeys)
# note that there's two ways to access submodels
assert (MISSION["flight segment"]["aircraft performance"]
        is MISSION.fs.aircraftp)
vars_of_interest.update(MISSION.fs.aircraftp.unique_varkeys)
vars_of_interest.add(M["D"])
print(sol.summary(vars_of_interest))
print(sol.table(tables=["loose constraints"]))
```

gpkit/solution_array.py
bqpd: hmm maybe `returnjson`?

gpkit/solution_array.py
Alexander-Da-Great: Thanks for the input. Added the showvars parameter to the savejson for applying this logic.

bqpd:
test this please

gpkit/solution_array.py
```diff
@@ -592,6 +593,31 @@ def savetxt(self, filename="solution.txt", *, printmodel=True, **kwargs):
...
+            key = str(k)
+            if isinstance(v, np.ndarray):
+                val = {"v": v.tolist(), "u": k.unitstr()}
+            else:
+                val = {"v": v, "u": k.unitstr()}
+            sol_dict[key] = val
+        for key in self.name_collision_varkeys():
+            del key.descr["necessarylineage"]
+        if printjson:
+            return str(sol_dict)
```
bqpd, line 616: shouldn't this be the JSON string (via `json.dumps`?)

bqpd: If you don't think this feature will be used, feel free to remove it!

bqpd: Other than this q, this looks perfect.

bqpd:
@Alexander-Da-Great sorry for the backseat driving with `printjson` :grimacing: I was thinking it'd be useful to have a "canonical" json representation that various streaming APIs could use, but realized that any early API will want to be idiosyncratic with its messaging (until a good standard is found).

bqpd (approved):

bqpd:
test this please

PR #1566, opened 2022-01-07
Clean breakdowns

bqpd:


bqpd:
test models please

galbramc:
Looks like the unit test take over an hour to run with this change...

bqpd:
i don't think they're hanging on the change though, and it's only on macys

galbramc:
Strange, I added a timer to the example execution. It definitely seems to be the examples that take a long time now.

bqpd:
oh dang I'll double check

galbramc:
I fixed the Windoze batch files, now we have a legit error (but strange) on Windoze as well.

